PR #4, opened 2020-01-10
Added kustomization for performance

slintes:
Initial Implementation of the `feature-deploy.sh` script along with manifests for e2e and demo envs.

Deployment on a test cluster with `FEATURES_ENVIRONMENT=demo make feature-deploy` succeeded.

Heads up, since we don't have official public images of the performance operator yet, the demo env uses images from my quay.io account, that's why I consider this as WIP. Also that image contains unmerged code from https://github.com/openshift-kni/performance-addon-operators/pull/34.

Also the `feature-deploy.sh` contains more imparative parts than we are aiming at. So consider this as a first iteration.


hack/feature-deploy.sh
```diff
@@ -2,11 +2,47 @@
...
+echo "[INFO]:labeling 1 worker node with worker-rt"
+node=$(${OC_TOOL} get nodes --selector='node-role.kubernetes.io/worker' -o name | head -1)
+${OC_TOOL} label --overwrite=true $node node-role.kubernetes.io/worker-rt=""
+
+# Override the image name when this is invoked from openshift ci
+# Not ideal, but kustomize does not support env vars directly :/
+if [ -n "${OPENSHIFT_BUILD_NAMESPACE}" ]; then
+        echo "[INFO]: Openshift CI detected, deploying using image $FULL_REGISTRY_IMAGE"
+        FULL_REGISTRY_IMAGE="registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:performance-addon-operators-registry"
+        cp feature-configs/e2e-gcp/performance-operator/operator_catalogsource.patch.yaml.in feature-configs/e2e-gcp/performance-operator/operator_catalogsource.patch.yaml
```
davidvossel: ha, wow. that's kind of annoying. There's really no reasonable way around this using variables at build time that I can see. The approach you've done here is about the best we can do if we don't know the image name up front.

They have ways of transforming images for containers in pods, deployments, etc... but not for this custom catalog source resource.

feature-configs/demo/performance-operator/wait_for_it.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
MarSik: I think the staged approach is exactly what is being used indeed.

feature-configs/base/performance-operator/operator_subscription.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: performance-addon-operators-subscription
+  namespace: openshift-performance-addon
+spec:
+  channel: alpha
+  name: performance-addon-operators
+  source: performance-addon-operators-catalogsource
```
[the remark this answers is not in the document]
MarSik: Yeah, they seem to have a priority list of objects to post first: https://github.com/kubernetes-sigs/kustomize/issues/202#issuecomment-483737500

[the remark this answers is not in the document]
yuvalk: Generally, they don't, and rely on loosely coupling and retries.. :-/

feature-configs/demo/performance-operator/wait_for_it.sh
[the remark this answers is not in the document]
MarSik: And as Yuval just commented above, it seems the whole thing is just executed in a loop until the errors disappear.

feature-configs/demo/sctp/sctp_module_mc.patch.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: worker-rt
```
MarSik: We should create a separate role for sctp as it is orthogonal to worker-rt. So we need manifests for MCP worker-sctp that reference worker-sctp node role and update the MC here to use it.

fedepaol: @MarSik is it fine if I do the sctp stuff in a separate PR?

feature-configs/base/performance-operator/operator_subscription.yaml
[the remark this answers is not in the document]
slintes: how can "they" (=`kustomize`?) do retries? `kustomize` just creates manifests locally. You need post them to the cluster yourself :thinking:

feature-configs/demo/performance-operator/wait_for_it.sh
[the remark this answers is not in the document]
slintes: > the whole thing is just executed in a loop until the errors disappear

by whom? By the deploy script?

hack/feature-deploy.sh
slintes: yes this is very annoying

> They have ways of transforming images for containers in pods, deployments

if you mean `kustomize edit`, that also just modifying manifests as we do here, so not much better.

But maybe I have a less uglier approach than this: using `${OPENSHIFT_BUILD_NAMESPACE}` in the e2e manifest patch, and piping it through `envsubst` before `oc apply`. Will give it a try.

feature-configs/base/performance-profile/performance_profile.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: performance.openshift.io/v1alpha1
+kind: PerformanceProfile
+metadata:
+  name: performance
+  namespace: openshift-performance-addon
```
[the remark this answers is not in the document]
slintes: fixed

feature-configs/base/namespace/namespace.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Namespace
```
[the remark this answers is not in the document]
slintes: done

feature-configs/demo/performance-operator/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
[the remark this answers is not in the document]
slintes: sure, if you have an idea? ;)

feature-configs/demo/sctp/sctp_module_mc.patch.yaml
slintes: removed

hack/feature-deploy.sh
slintes: wait...
We don't need to handle ${OPENSHIFT_BUILD_NAMESPACE} at all. We don't build the image here, we want to test an existing one :)
So due to lack of an existing upstream image I use my own here as well now, until we have official images available.

feature-configs/base/performance-operator/operator_subscription.yaml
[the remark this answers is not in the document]
davidvossel: right, so kustomize isn't going to do the retry loop for us. it's something we'll have to do.

for ordering, lets just see what order they get posted in. If this is an issue, maybe there's a way to use a transformer to influence ordering.

hack/feature-deploy.sh
davidvossel: we still need the ability to pass in an image via an ENV VAR during `make feature-deploy` though.

feature-configs/demo/performance-operator/wait_for_it.sh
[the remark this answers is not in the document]
davidvossel: yes

davidvossel:
/hold

waiting on some manual test results first.

slintes:
rebased

hack/feature-deploy.sh
```diff
@@ -2,11 +2,66 @@
...
 # expect oc to be in PATH by default
-OC_TOOL="${OC_TOOL:-oc}"
+export OC_TOOL="${OC_TOOL:-oc}"
+
+# expect kustomize to be in PATH by default
+KUSTOMIZE="${KUSTOMIZE:-kustomize}"
+
+# Label 1 worker node
+echo "[INFO]:labeling 1 worker node with worker-rt"
+node=$(${OC_TOOL} get nodes --selector='node-role.kubernetes.io/worker' -o name | head -1)
```
dshchedr: Small question. If my master nodes also have "worker" label - the first master will be chosen, is it ok?

[root@dell-r730-012 dev-scripts]# oc get nodes --selector='node-role.kubernetes.io/worker' -o name | head -1
node/master-0

[root@dell-r730-012 dev-scripts]# oc get node
NAME       STATUS   ROLES 
master-0   Ready    master,worker
master-1   Ready    master,worker
master-2   Ready    master,worker
worker-0   Ready    worker
worker-1   Ready    worker

hack/feature-deploy.sh
slintes: can you elaborate for what usecase we still need env vars? I'm asking because if we don't need to support them, we could remove download and usage of `kustomize` and use `oc -k ...`. But that would hinder usage of `envsubst` or similar.

hack/feature-deploy.sh
```diff
@@ -2,11 +2,67 @@
...
+    echo "[INFO] Deploying feature '$feature' for environment '$FEATURES_ENVIRONMENT'"
+    set +e
+    if ! ${KUSTOMIZE} build $feature_dir | ${OC_TOOL} apply -f -
+    then
+      echo "[WARN] Deployment of feature '$feature' failed."
+      feature_failed=1
+    fi
+    set -e
+
+  done
```
yuvalk: I think this logic is incorrect:
1. afaik apply will return 0 regardless of other errors that might occur
2. even if such errors occur, there's no need to re-apply, just wait for them to settle. 
3. which this loop doesnt do (something like kubectl rollout status, which isn't available in oc[??] not sure whats the alternative). ie we would probably be out of that loop before features declarations are really "done"

other then that, we can use a single oc apply -k, if we create a folder referencing all the wanted features.
which IMHO would be 'cleaner' than running multiple kustomize commands. we can even have an overlay dir with some supported/most relevant variations (all, networking, performance, etc)

slintes: Thx for review!

1. no, this works: as long as the CR can't be posted because the CRD isn't there yet, it returns an error
2. we don't want custom wait logic, that's why we iterate until all succeeds. Custom wait would be "check if CRD exists already"
3. yes, we might want to have a sanity check that everything works as expected. I'd like to leave that out of scope of this PR, in order to get it merged asap, so that others can add more  features on top of this
4. about `oc -k`, see https://github.com/openshift-kni/cnf-features-deploy/pull/4#discussion_r366218426

slintes:
let's see waht happens

/lgtm
/approve
/hold cancel

slintes:
Seems merge hangs because test status wasn't reported back to github. Let's try again...

/test all

hack/feature-deploy.sh
davidvossel: what other options do we have for providing an image override for an image in a private repo that we don't want ot link to in a public repo?

slintes: good usecase indeed. I'm not aware of any other option to do that.
So I'll leave the `kustomize | oc` apply flow and don't replace it with `oc -k`, so we can simply put `envsubst` in between.
fyi @MarSik

PR #11, opened 2020-01-16
Add sriov feature deployment

SchSeba:
This PR add the sriov operator deployment for the base.

For the demo I also create the test namespace and the policy.

### DIFF
--- a/feature-configs/base/sriov/04-sriov-networknodepolicy.yaml
+++ b/feature-configs/base/sriov/04-sriov-networknodepolicy.yaml
@@ -1,0 +1,16 @@
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetworkNodePolicy
+metadata:
+  name: sriov-network-node-policy
+  namespace: openshift-sriov-network-operator
+spec:
+  deviceType: netdevice
+  isRdma: false
+  nicSelector:
+    pfNames:
+      - eth0
+  nodeSelector:
+    node-role.kubernetes.io/worker: ""
+  numVfs: 1
+  priority: 10
+  resourceName: sriovnic
--- a/feature-configs/base/sriov/05-sriov-network.yaml
+++ b/feature-configs/base/sriov/05-sriov-network.yaml
@@ -1,0 +1,18 @@
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetwork
+metadata:
+  name: sriov-network
+  namespace: openshift-sriov-network-operator
+spec:
+  ipam: |
+    {
+      "type": "dhcp"
+    }
+  networkNamespace: sriov-testing
+  resourceName: sriovnic
+  vlan: 0
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: sriov-testing
--- a/feature-configs/demo/sriov/kustomization.yaml
+++ b/feature-configs/demo/sriov/kustomization.yaml
@@ -1,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../../base/sriov
+patchesStrategicMerge:
+  - sriov-subscription.yaml
+  - sriov-networknodepolicy.yaml
--- a/feature-configs/base/sriov/01-sriov-namespace.yaml
+++ b/feature-configs/base/sriov/01-sriov-namespace.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-sriov-network-operator
+  labels:
+    openshift.io/run-level: "1"
--- a/feature-configs/base/sriov/02-sriov-operatorgroup.yaml
+++ b/feature-configs/base/sriov/02-sriov-operatorgroup.yaml
@@ -1,0 +1,8 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: sriov-network-operators
+  namespace: openshift-sriov-network-operator
+spec:
+  targetNamespaces:
+  - openshift-sriov-network-operator
--- a/feature-configs/base/sriov/03-sriov-subscription.yaml
+++ b/feature-configs/base/sriov/03-sriov-subscription.yaml
@@ -1,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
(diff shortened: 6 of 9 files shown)


feature-configs/demo/sriov/11-sriov-networknodepolicy.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: sriov-network-node-policy
+  namespace: openshift-sriov-network-operator
+spec:
+  deviceType: netdevice
+  isRdma: true
+  mtu: 1500
+  nicSelector:
+    pfNames:
+    - eno1
```
fedepaol: I would use this as base and use kustomize to replace the pf name of the given environment.
This will make stuff easier once we have a new envorinment.

Also, only kustomization should stay under 'demo'.

feature-configs/demo/sriov/10-sriov-daemonset.yaml
```diff
@@ -0,0 +1,66 @@
+kind: ConfigMap
```
fedepaol: Since this is `demo` specific, is it worth checking if what we have on that environment has one of the supported nics? If that's the case. I would remove this.
I am going to check it.

feature-configs/demo/sriov/12-sriov-network.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: sriovnetwork.openshift.io/v1
```
fedepaol: I would put this to base (and eventually kustomize the ipam section). 
Do we have dhcp in the demo environment?

feature-configs/demo/sriov/12-sriov-network.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: SriovNetwork
+metadata:
+  name: sriov-network
+  namespace: openshift-sriov-network-operator
+spec:
+  ipam: |
+    {
+      "type": "dhcp"
+    }
+  networkNamespace: sriov-testing
```
fedepaol: I would kustomize this as well.

feature-configs/base/sriov/04-sriov-networknodepolicy.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: SriovNetworkNodePolicy
+metadata:
+  name: sriov-network-node-policy
+  namespace: openshift-sriov-network-operator
+spec:
+  deviceType: netdevice
+  isRdma: false
+  nicSelector:
+    pfNames:
+      - eno1
```
fedepaol: Now that the policy is under base, you can provide a kustomization example under demo to change something (numVfs / pfName for example)

feature-configs/base/sriov/06-app-namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
```
fedepaol: I would remove this whole file and have the namespace as part of the kustomizatoin of the network.

feature-configs/base/sriov/05-sriov-network.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetwork
+metadata:
+  name: sriov-network
+  namespace: openshift-sriov-network-operator
+spec:
+  ipam: |
```
fedepaol: Can we kustomize this section? So we'll provide the relevant ipam depending on the environment (in demo we might have dhcp)? Wdyt?

feature-configs/base/sriov/04-sriov-networknodepolicy.yaml
slintes: > provide a kustomization example under demo to change something

not sure if this repo is for showing kustomize features... IMHO we should just provide good defaults (in case there are such) in base, and then specific values which are really targeting our CI and demo environments #justMy2Cents

fedepaol: Thanks @slintes. I'll try to explain better: I did not mean to show kustomize features, but an example of how a real user would use kustomize for deploying the same network definition on different environments, maintaining the same structure.
That's why I thought the pfName was a good example (and it will be probably different between envs). Same for vf#, it may depend on the nic shipped on the environment.

feature-configs/demo/sriov/sriov-daemonset.yaml
```diff
@@ -0,0 +1,66 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: sriov-unsupported-nics-entrypoint
```
fedepaol: Sorry, I finally checked.

Under the demo machine we currently have, we are using one of the supported mellanox nics (deviceid of PF 1015).
See also https://docs.openshift.com/container-platform/4.2/networking/multiple-networks/configuring-sr-iov.html#supported-devices_configuring-sr-iov

In theory, we could drop this but in practice it could be useful to run against clusters with non officially supported devices.  I would probably remove this to until we have for example a CI machine that requires it.
What do you think @SchSeba ?

SchSeba: works for me

SchSeba:
@fedepaol can you take another look?

e-minguez:
I've tried to deploy this but it tries to pull this image `image-registry.openshift-image-registry.svc:5000/openshift/ose-sriov-network-operator@sha256:9029bbf423d7f5d091b9d27a67bfdb479845e412a3c15511fcfc8bb33232e0cd` which is not available in the internal registry. 
Is there something else that needs to be done prior to applying those manifests with kustomize? Thanks!

SchSeba:
Hi, @e-minguez thanks for the comment!

we are working on this issue but is unrelated to this PR the change needs to be in the CI system.

feature-configs/demo/sriov/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../../base/sriov
+patchesStrategicMerge:
+  - sriov-subscription.yaml
+  - sriov-networknodepolicy.yaml
```
fedepaol: Nit: missing new line

fedepaol:
/lgtm

Also, sriov is to be intended to run against the demo environment where all the necessary setup will be in place. CI on gcp does not have sr-iov devices.

PR #15, opened 2020-01-17
Set the $GOPATH/bin variable to path when running tests

fedepaol:
Our tests (and probably others we are going to import) depend on the fact that ginkgo is available.
We already download it, but it also has to be in the path

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -1,10 +1,20 @@
 #!/bin/bash
 
+which go
+if [ $? -ne 0 ]; then
+  echo "No go command available"
+  exit 1
+fi
+
+GOPATH="${GOPATH:-~/go}"
+export PATH=$PATH:$GOPATH/bin
+
 which ginkgo
 if [ $? -ne 0 ]; then
 	echo "Downloading ginkgo tool"
 	go install github.com/onsi/ginkgo/ginkgo
 fi
+
 
 FOCUS=$(echo "$FEATURES" | tr ' ' '|') 
 echo "Focusing on $FOCUS"


hack/run-functests.sh
```diff
@@ -1,11 +1,20 @@
 #!/bin/bash
 
+which go
+if [ $? -ne 0 ]; then
+  echo "No go command available"
+fi
+
+GOPATH="${GOPATH:-~/go}"
+PATH=$PATH:$GOPATH/bin
```
fromanirh: I'd export it to avoid nasty surprises in the future:
https://stackoverflow.com/questions/1158091/defining-a-variable-with-or-without-export

hack/run-functests.sh
```diff
@@ -1,11 +1,20 @@
...
+
 which ginkgo
 if [ $? -ne 0 ]; then
 	echo "Downloading ginkgo tool"
 	go install github.com/onsi/ginkgo/ginkgo
```
fromanirh: In another PR, we should vendor it.

hack/run-functests.sh
```diff
@@ -1,11 +1,20 @@
 #!/bin/bash
 
+which go
+if [ $? -ne 0 ]; then
+  echo "No go command available"
```
fromanirh: then bail out with error I guess

fedepaol: Aw yes! It remained in the keyboard!

hack/run-functests.sh
fedepaol: Thanks!

PR #19, opened 2020-01-19
Add a demo for ptp grandmaster and slave

SchSeba:

### DIFF
--- a/feature-configs/demo/ptp/ptpconfig-grandmaster.yaml
+++ b/feature-configs/demo/ptp/ptpconfig-grandmaster.yaml
@@ -1,0 +1,16 @@
+apiVersion: ptp.openshift.io/v1
+kind: PtpConfig
+metadata:
+  name: grandmaster
+  namespace: openshift-ptp
+spec:
+  profile:
+  - name: "grandmaster"
+    interface: "eno1"
+    ptp4lOpts: ""
+    phc2sysOpts: "-a -r -r"
+  recommend:
+  - profile: "grandmaster"
+    priority: 4
+    match:
+    - nodeLabel: "ptp/grandmaster"
--- a/feature-configs/demo/ptp/ptpconfig-slave.yaml
+++ b/feature-configs/demo/ptp/ptpconfig-slave.yaml
@@ -1,0 +1,16 @@
+apiVersion: ptp.openshift.io/v1
+kind: PtpConfig
+metadata:
+  name: slave
+  namespace: openshift-ptp
+spec:
+  profile:
+  - name: "slave"
+    interface: "eno1"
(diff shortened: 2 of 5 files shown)


feature-configs/demo/ptp/ptpconfig-grandmaster.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: ptp.openshift.io/v1
```
fedepaol: Would it make sense to provide a default for these manifests and only the kustomizations under demo?

feature-configs/demo/ptp/ptpconfig-slave.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: ptp.openshift.io/v1
```
fedepaol: Same for this file

feature-configs/demo/ptp/ptpconfig-grandmaster.yaml
SchSeba: this is not a production way you should not configure the master to be one of the nodes.
All the nodes in the cluster should only be slaves, we do it like that in our environment for testing (we don't have an external device to sync from) this is why this file is in the demo folder

SchSeba:
@fedepaol thanks for the comment I was thinking about that but I will like to leave the base folder to be the installation process only what you think?

feature-configs/demo/ptp/ptpconfig-grandmaster.yaml
fedepaol: Ah, got it. Makes sense, thanks for the explanation

feature-configs/demo/ptp/ptpconfig-slave.yaml
fedepaol: This should belong to base then?

fedepaol:
> @fedepaol thanks for the comment I was thinking about that but I will like to leave the base folder to be the installation process only what you think?

Yes, but having the slaves set is part of the installation of a cluster I think.
You should see the base as how would you setup a "cluster version" so you cna change as less as you can for the different cluster flavours.
In that sense, I would have the slaves configuration as part of base, customizing the interface name or something like that.

SchSeba:
@fedepaol can you take another look?

feature-configs/demo/ptp/ptpconfig-slave.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: PtpConfig
```
fedepaol: Now that the slave profile is in base, can we remove this (or maybe just have this as a little patch?)

PR #20, opened 2020-01-20
Skip vendor folder from getting go vet-ed

fedepaol:
We don't want go vet to run against vendor folder


fedepaol:
Closing as this is not the right approach

PR #24, opened 2020-01-21
Added more tests to sctp feature. Tests: KernelModule, Connectivity

nikita133:


functests/sctp/sctp.go
```diff
@@ -38,58 +41,136 @@ var _ = Describe("sctp", func() {
...
+				clientNode = nodes.Items[0].ObjectMeta.Labels[hostnameLabel]
+				serverNode = nodes.Items[0].ObjectMeta.Labels[hostnameLabel]
+				if len(nodes.Items) > 1 {
+					serverNode = nodes.Items[1].ObjectMeta.Labels[hostnameLabel]
+				}
+				createSctpService(client.Client)
+			})
+			Context("Client Server Connection", func() {
+				// OCP-26995
+				It("Negative. SCTP disabled. Should NOT connect a client pod to a server pod", func() {
```
fedepaol: Can you put "Negative - Sctp disabled" in the root describe `Describe("Test Connectivity Negative", func() ` and leave only `Should not ...`. It will end up in a more pleasant output

functests/sctp/sctp.go
```diff
@@ -212,3 +316,13 @@ func jobForNode(name, node, app string, cmd []string, args []string) *k8sv1.Pod
 
 	return &job
 }
+
+func beforeAll(fn func()) {
```
fedepaol: beforeAll was moved here https://github.com/openshift-kni/cnf-features-deploy/blob/edffd2d01e128a20d690f2710a2732d9bf455faf/functests/utils/execute/ginkgo.go#L7

Please use that

functests/sctp/sctp.go
```diff
@@ -45,59 +49,136 @@ var _ = Describe("sctp", func() {
...
+					checkForSctpReady(client.Client)
+				})
+				// OCP-26760
+				It("Should connect a client pod to a server pod", func() {
+					testClientServerConnection(client.Client, runningPod.Status.PodIP,
+						activeService.Spec.Ports[0].Port, clientNode, serverPod.Name)
+				})
+				// OCP-26763
+				It("Should connect a client pod to a server pod. Feature LatencySensitive Active", func() {
+					fg, err := client.Client.FeatureGates().Get("cluster", metav1.GetOptions{})
```
fedepaol: @nikita133 the problem here is that the server exits as soon as it receives a message (see https://github.com/openshift-kni/cnf-features-deploy/blob/master/tools/sctptester/main.go#L83 )
Since we are using the same server pod, the second test will not find it running and the client will fail.

fedepaol:
`             pods "scptserver" is forbidden: error looking up service account sctptest/default: serviceaccount "default" not found `

Not related to us.
/retest

fedepaol:
passed!
/lgtm
/approve

PR #26, opened 2020-01-21
Run the origin tests as part of the external tests

fedepaol:
This PR is to extract the ose tests binary from the ose tests image and then running the acceptance tests.
We can't consume the container as under CI, the tests are being executed in a pod and docker in docker require having privileges we don't have there.

So the not so pleasant way we do that is to use skopeo to fetch the image, look for the binary and then extract it.

### DIFF
--- a/external-tests/clone_repo.sh
+++ b/external-tests/clone_repo.sh
@@ -14,7 +14,7 @@ fi
 # detect latest master hash of the target repo if we're not pinned to a specific commit hash
 if [ -z "$TESTS_TARGET_HASH" ]; then
     echo "Using latest master branch commit for $TESTS_REPO"
-    TESTS_TARGET_HASH=$(git ls-remote "$TESTS_REPO" | grep "refs/heads/master$" | cut -f 1)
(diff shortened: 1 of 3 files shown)


fedepaol:
/hold
Requires https://github.com/openshift-kni/cnf-features-deploy/pull/10 to get merged first

fedepaol:
This is going to be fun

fedepaol:
` skopeo not available, exiting `
That's a start

fedepaol:
Uhm, skopeo is added in the image and locally I am able to execute it.

/retest

fedepaol:
`Error waiting for Creating Subnetwork: Quota 'SUBNETWORKS' exceeded.  Limit: 450.0 globally."`
I am afraid we are done for today

slintes:
the tests were running, awesome :+1: 
however, a lot (at least i spotted some in that huge log file) are failing with `exec: "kubectl": executable file not found in $PATH`. Maybe just create an alias to `oc`?

slintes:
seems the kubectl alias did not work :/

fedepaol:
Will check on Monday. I tried it locally in the container and it seemed to work.

fedepaol:
166 failed tests after the kubectl fix.
Before it was 832.

PR #37, opened 2020-01-24
Update README to the latest changes.

fedepaol:
As we changed the way we deploy / test, I am updating the readme to reflect that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,19 +2,49 @@
 
 This repo contains example kustomize configs used to installed openshift features required for CNF workloads and a e2e functional test suite used to verify cnf related features.
 
-# Contributing kustomize configs
+## Contributing kustomize configs
 
 All kustomize configs should be entirely declarative in nature. This means no bash plugin modules performing imparative tasks. Features should be installed simply by posting manifests to the cluster. After posting manifests, determining when the cluster has converged on those manifests successully should be observable.
 
-# Usage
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,19 +2,49 @@
...
+- You need at least one node with the `node-role.kubernetes.io/worker-cnf=""` label and a `MachineConfigPool` matching `worker-cnf` machine configurations
 
-## Deployment
+You can run `make setup-test-cluster` to have the first two (or the first in case of only one) workers labeled as `worker-cnf` and to have the `MachineConfigPool` created.
+
+### Configuring
+
+All the Makefile rules depend on two environment variable, either for deploying, waiting and choosing what tests to run.
+
+#### FEATURES
```
nikita133: You tried to make var FEATURE to be part of  Configuring section however because it written with all capital letters it seems like new section which doesn't related to Configuring section. I suggest to reduce the font size by adding one "#".

README.md
```diff
@@ -2,19 +2,49 @@
...
+
+i.e. `FEATURES=(sctp ptp sriov)`, drives what features are going to be deployed using kustomize, and what tests are going to be run.
+
+The current default values is `(sctp performace)`
+
+#### FEATURES_ENVIRONMENT
```
nikita133: See my comment above

PR #42, opened 2020-01-27
Change the sriov network config in demo

SchSeba:
This PR change the network config CR to work with mellanox

For mellanox cards the driver needs to by netdevice for both dpdk
workload or regular network card.

We also need to mark the RDMA as true for mellanox cards

### DIFF
--- a/feature-configs/demo/sriov/sriov-networknodepolicy-dpdk.yaml
+++ b/feature-configs/demo/sriov/sriov-networknodepolicy-dpdk.yaml
@@ -4,8 +4,8 @@ metadata:
   name: dpdk-network-node-policy
   namespace: openshift-sriov-network-operator
 spec:
-  deviceType: vfio-pci
-  isRdma: false
+  deviceType: netdevice
+  isRdma: true
   nicSelector:
     pfNames:
       - eno2
--- a/feature-configs/demo/sriov/sriov-networknodepolicy.yaml
+++ b/feature-configs/demo/sriov/sriov-networknodepolicy.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: openshift-sriov-network-operator
 spec:
   deviceType: netdevice
-  isRdma: false
+  isRdma: true
   nicSelector:
     pfNames:
       - eno1


yuvalk:
is there anyway to detect the card and act accordingly?
we should at least include both configs

fedepaol:
> is there anyway to detect the card and act accordingly?
> we should at least include both configs

Not via manifests. Also, that should be the point of having different layers shaped against different installations.

yuvalk:
> > is there anyway to detect the card and act accordingly?
> > we should at least include both configs
> 
> Not via manifests. Also, that should be the point of having different layers shaped against different installations.

so we need to add the two options maybe something like
demo
demo-intel
demo-mellanox

fedepaol:
> > > is there anyway to detect the card and act accordingly?
> > > we should at least include both configs
> > 
> > 
> > Not via manifests. Also, that should be the point of having different layers shaped against different installations.
> 
> so we need to add the two options maybe something like
> demo
> demo-intel
> demo-mellanox

I'd do that as soon as another environment with intel pops out. As the only demo environment has mellanox, I would not add the extra complexity. WDYT @yuvalk ?

yuvalk:
> > > > is there anyway to detect the card and act accordingly?
> > > > we should at least include both configs
> > > 
> > > 
> > > Not via manifests. Also, that should be the point of having different layers shaped against different installations.
> > 
> > 
> > so we need to add the two options maybe something like
> > demo
> > demo-intel
> > demo-mellanox
> 
> I'd do that as soon as another environment with intel pops out. As the only demo environment has mellanox, I would not add the extra complexity. WDYT @yuvalk ?

I think we should include both envs as other folks might wanna test and use it.
but after thinking we might better do
sriov (with general sriov kustomization, if there's such)
sriov-intel
sriov-mellanox

and have demo override the sriov-mellanox for now
so, if someone wanna deploy on intel, they can
and we don't complex anything for demo

fedepaol:
> > > > > is there anyway to detect the card and act accordingly?
> > > > > we should at least include both configs
> > > > 
> > > > 
> > > > Not via manifests. Also, that should be the point of having different layers shaped against different installations.
> > > 
> > > 
> > > so we need to add the two options maybe something like
> > > demo
> > > demo-intel
> > > demo-mellanox
> > 
> > 
> > I'd do that as soon as another environment with intel pops out. As the only demo environment has mellanox, I would not add the extra complexity. WDYT @yuvalk ?
> 
> I think we should include both envs as other folks might wanna test and use it.
> but after thinking we might better do
> sriov (with general sriov kustomization, if there's such)
> sriov-intel
> sriov-mellanox
> 
> and have demo override the sriov-mellanox for now
> so, if someone wanna deploy on intel, they can
> and we don't complex anything for demo

The kustomizatoin is still dependent on hw (i.e., we are using the nic name). If we want to split it, we either need a new demo-intel (duplicate of demo except the sriov parameters) OR an additional layer that uses demo as base and overrides only the sriov parameters.

I am a bit scared of the second one as it may feel simple now but if we'll have other layers the complexity might get out of hand.

/cc @SchSeba

fedepaol:
/test ci
Let's try again after the changes in the image

PR #51, opened 2020-01-28
Fix tester dockerfile as it will be executed from project root in CI.

fedepaol:
We want to COPY only the sctptester folder and not the whole prj folder.

This is needed to fix failures of rehearsals in https://github.com/openshift/release/pull/6793

### DIFF
--- a/tools/sctptester/Dockerfile
+++ b/tools/sctptester/Dockerfile
@@ -1,7 +1,7 @@
 
 FROM golang:1.13 AS builder
 WORKDIR /gotest
-COPY . .
+COPY ./tools/sctptester .
 RUN go mod init github.com/openshift-kni/cnf-features-deploy/sctptester
 RUN go get
 RUN go build -o sctptest


slintes:
This is a bit unusual, because usually you execute the docker command in the same directory.
Maybe fix CI instead? There seems to be a `context_dir` property for the path to the Dockerfile, and then `dockerfile_path` only has the name of the Dockerfile without path.
See https://github.com/openshift/ci-tools/search?q=context_dir&unscoped_q=context_dir

fedepaol:
> This is a bit unusual, because usually you execute the docker command in the same directory.
> Maybe fix CI instead? There seems to be a `context_dir` property for the path to the Dockerfile, and then `dockerfile_path` only has the name of the Dockerfile without path.
> See https://github.com/openshift/ci-tools/search?q=context_dir&unscoped_q=context_dir

Ah, nice! Thanks for the suggestion, will try the other way around.

/hold

fedepaol:
CI is not helping here but seems https://github.com/openshift/release/pull/6793 went past building the tester image:

```
 2020/01/28 11:15:22 Building sctptester
2020/01/28 11:17:01 Build sctptester succeeded after 1m39s
2020/01/28 11:17:02 Tagging sctptester into stable 
```

Closing this, thanks @slintes !

slintes:
perfect, glad it worked :)

PR #52, opened 2020-01-28
Replaced kustomize with oc apply -k

slintes:
Since `oc` comes with an older `kustomize` version, overrides need to use
`bases` instead `resources`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ export FEATURES_ENVIRONMENT?=e2e-gcp
 	golint \
 	govet \
 	ci-job \
-	kustomize \
 	feature-deploy
 
 TARGET_GOOS=linux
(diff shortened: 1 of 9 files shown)


fedepaol:
I'd wait for CI, but this

/lgtm

fedepaol:
> /retest

We did not fix this yet right? Only the perf operator repo I think

slintes:
> > /retest
> 
> We did not fix this yet right? Only the perf operator repo I think

yes, I wanted to exclude coincidence ;)
but since it failed here again with the same error, and operator repo is green now, the fix in the operator repo really worked :)
here the PR for this repo: https://github.com/openshift-kni/cnf-features-deploy/pull/54

slintes:
failure looks different than in other PRs, will try to reproduce locally

/hold

slintes:
So on my cluster deployment was fine, performance tests succeeded on 1st run, sctp tests on 2nd run (1st run failed with a API problem, default sa secret not found).

/hold cancel
/retest

slintes:
rebased...

slintes:
all green :tada:

PR #53, opened 2020-01-28
Add ptp labels on the nodes

SchSeba:
This PR adds a label of grandmaster on the first master and all the other nodes (master/workers) are labeled as slaves

### DIFF
--- a/hack/setup-test-cluster.sh
+++ b/hack/setup-test-cluster.sh
@@ -13,6 +13,16 @@ do
     ${OC_TOOL} label $node node-role.kubernetes.io/worker-cnf=""
 done
 
+echo "[INFO]: Labeling first node as the ptp grandmaster"
+node=$(${OC_TOOL} get nodes -o name | sed -n 1p)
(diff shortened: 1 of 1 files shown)


hack/setup-test-cluster.sh
```diff
@@ -13,6 +13,20 @@ do
...
+# Label First master as grandmaster
+echo "[INFO]: Labeling first master as the ptp grandmaster"
+nodes=$(${OC_TOOL} get nodes --selector='node-role.kubernetes.io/master' -o name | sed -n 1p)
+for node in $nodes
+do
+    ${OC_TOOL} label $node ptp/grandmaster=""
+done
+
+echo "[INFO]: Labeling all the other hosts as ptp slaves"
+nodes=$(${OC_TOOL} get nodes -o name | sed 1d)
```
fedepaol: If you do like this, you will skip the first result which is not necessarily the master you labelled above. Is that fine?

SchSeba: I make a change in the first loop can you take a look?

PR #59, opened 2020-01-29
Remove kubectl alias and create a symlink to oc.

fedepaol:
The tests are executed via /bin/sh and there is no point in adding the alias to bashrc.


fedepaol:
Everything in line, tide is saying "Not mergeable. Needs bugzilla/valid-bug label. "

Merging manually

PR #60, opened 2020-01-31
DPDK functests: adding a simple sanity test

dorzheh:
The test checks whether DPDK stack works properly.
The test spins up a pod running appropriate container bundled with the DPDK utils and required tools. Then, it mounts appropriate bash/expect script as configmap and executes the script. The script starts up testpmd and uses expect to interact with the process. Afterward, the test parses output and makes a decision.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 #TODO add default features here
-export FEATURES?=sctp performance dpdk
+export FEATURES?=sctp performance
 
 # The environment represents the kustomize patches to apply when deploying the features
 export FEATURES_ENVIRONMENT?=e2e-gcp
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 #TODO add default features here
-export FEATURES?=sctp performance
+export FEATURES?=sctp performance dpdk
 
 # The environment represents the kustomize patches to apply when deploying the features
 export FEATURES_ENVIRONMENT?=e2e-gcp
--- a/functests/dpdk/dpdk.go
+++ b/functests/dpdk/dpdk.go
@@ -1,0 +1,250 @@
+package dpdk
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/client"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/execute"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/namespaces"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/nodes"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/pods"
+)
+
+const (
+	dpdkHostRole         = "worker"
+	hostnameLabel         = "kubernetes.io/hostname"
+	dpdkAnnotationNetwork = "dpdk-network"
+	testDpdkNamespace     = "dpdk-testing"
+	testCmdPath           = "/opt/test.sh"
+
+)
+
+var dpdkAppImage string
+
+func init() {
+	// Set DPDK app image
+	dpdkAppImage = os.Getenv("DPDK_APP_IMAGE")
+	if dpdkAppImage == "" {
+		dpdkAppImage = "quay.io/schseba/dpdk-s2i-base:ds"
+	}
+}
+
+var _ = Describe("dpdk", func() {
+	var nList []corev1.Node
+
+	execute.BeforeAll(func() {
+		// create the namespace
+		err := namespaces.Create(testDpdkNamespace, client.Client)
+		Expect(err).ToNot(HaveOccurred())
+
+		// clean up the namespace
+		err = namespaces.Clean(testDpdkNamespace, client.Client)
+		Expect(err).ToNot(HaveOccurred())
+
+		// get nodes
+		nList, err = nodes.GetByRole(client.Client, dpdkHostRole)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(nList)).ShouldNot(Equal(0))
+	})
+
+	var _ = Context("Run a sanity test on each worker", func() {
+		var configMapName string
+		var p *corev1.Pod
+		AfterEach(func() {
+			pods.WaitForDeletion(client.Client,p, 2*time.Minute)
+			deleteTestpmdConfigMap(configMapName)
+		})
+		It("Should forward and receive packets", func() {
+			c := createTestpmdConfigMap(testDpdkNamespace)
+			configMapName = c.Name
+			for _, n := range nList {
+				p = createTestPod(n.Name, testDpdkNamespace, c.Name)
+				pods.WaitForCondition(client.Client, p,corev1.ContainersReady, corev1.ConditionTrue, 2*time.Minute)
+				By(fmt.Sprintf("Executing %s inside the pod %s", testCmdPath, p.Name))
+				out, err := exec.Command("oc", "rsh", "-n", p.Namespace, p.Name, "bash", testCmdPath).CombinedOutput()
+				Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("cannot execute %s inside the pod %s", testCmdPath, p.Name))
+				By("Parsing output from the DPDK application")
+				checkRxTx(string(out))
+			}
+		})
+	})
+})
+
+// creteTestPod creates a pod that will act as a runtime for the DPDK test application
+func createTestPod(nodeName, namespace, configMapName string) *corev1.Pod {
+	defaultMode := int32(0755)
+
+	res := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "test-dpdk",
+			Labels: map[string]string{
+				"app": "test-dpdk",
+			},
+			Annotations: map[string]string{
+				"k8s.v1.cni.cncf.io/networks": dpdkAnnotationNetwork,
+			},
+			Namespace: namespace,
+		},
+		Spec: corev1.PodSpec{
+			RestartPolicy: corev1.RestartPolicyNever,
+			Containers: []corev1.Container{
+				{
+					Name:  "test-dpdk",
+					Image: dpdkAppImage,
+					SecurityContext: &corev1.SecurityContext{
+						Capabilities: &corev1.Capabilities{
+							Add: []corev1.Capability{"IPC_LOCK"},
+						},
+					},
+					Command:         []string{"/bin/bash", "-c", "--"},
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -1,5 +1,5 @@
 #TODO add default features here
-export FEATURES?=sctp performance
+export FEATURES?=sctp performance dpdk
```
davidvossel: do we expect dpdk tests to work on the gcp public cloud infrastructure, or should it be limited to baremetal? It's unclear what the network requirements are here

dorzheh:
@David Vossel <dvossel@redhat.com> , despite the fact that this test is
able to run on an environment  equipped with  vNICs (para-virtualized or
emulated), at this stage I would suggest limiting this to baremetal only

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,250 @@
...
+			deleteTestpmdConfigMap(configMapName)
+		})
+		It("Should forward and receive packets", func() {
+			c := createTestpmdConfigMap(testDpdkNamespace)
+			configMapName = c.Name
+			for _, n := range nList {
+				p = createTestPod(n.Name, testDpdkNamespace, c.Name)
+				pods.WaitForCondition(client.Client, p,corev1.ContainersReady, corev1.ConditionTrue, 2*time.Minute)
+				By(fmt.Sprintf("Executing %s inside the pod %s", testCmdPath, p.Name))
+				out, err := exec.Command("oc", "rsh", "-n", p.Namespace, p.Name, "bash", testCmdPath).CombinedOutput()
```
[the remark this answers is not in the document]
dorzheh: @SchSeba , so as stated , if we are going to use  s2i , it doesn't make any sense to use neither exec.Command nor the k8s.io e2e framework, but attach to the log stream by using corev1's GetLogs method.

Makefile
fedepaol: Not only that, those features are those installed (& instrumented) by default on gcp CI.

For this reason, dpdk should not be there, the same way sriov or ptp are not there.

SchSeba:
@dorzheh as we spoke offline today I think this task can make use of this PR https://github.com/openshift-kni/cnf-features-deploy/pull/55

I will deploy the application using a buildconfig and a deploymentconfig, then the test can just validate the logs from the running application that should print the dpdk workload

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,134 @@
...
+	Expect(len(r)).To(Equal(6), "the slice doesn't contain 6 elements")
+	d, err := strconv.Atoi(r[1])
+	Expect(err).ToNot(HaveOccurred())
+	return d
+}
+
+// findPod finds a pod running on appropriate node
+func findPod(nodeName string) *corev1.Pod {
+	listOptions := metav1.ListOptions{
+		FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": nodeName}).String(),
```
SchSeba: Why do we need the node label?

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	listOptions.LabelSelector = labels.SelectorFromSet(labels.Set{"deploymentconfig": "s2i-dpdk-app"}).String()
+	p, err := client.Client.Pods(testDpdkNamespace).List(listOptions)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(len(p.Items)).ShouldNot(Equal(0))
+
+	var pod corev1.Pod
+	var i int
```
SchSeba: I think it should be better to just have a boolean in change it to true before the break

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+	for i := 0; i < len(str); i++ {
+		if strings.Contains(str[i], logEntry) {
+			i++
+			d := getNumberOfPackets(str, i)
+			Expect(d).Should(BeNumerically(">", 0), "number of received packets should be greater than 0")
+
+			i++
+			d = getNumberOfPackets(str, i)
+			Expect(d).Should(BeNumerically(">", 0), "number of transferred packets should be greater than 0")
+		}
```
SchSeba: Can we break the for if we enter the if section?

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+		testDpdkNamespace = "dpdk-testing"
+	}
+}
+
+var _ = Describe("dpdk", func() {
+	var _ = Context("Run a sanity test on a worker", func() {
+		It("Should forward and receive packets", func() {
+			var out string
+			p := findPod()
+			pods.WaitForCondition(client.Client, p, corev1.ContainersReady, corev1.ConditionTrue, 2*time.Minute)
```
SchSeba: the find pod have the same condition inside can we remove this one?

functests/dpdk/dpdk.go
dorzheh: @SchSeba, since the required lines will be at the end of the log, a break is not so important here. But I can add it, in any case.

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+	if testDpdkNamespace == "" {
+		testDpdkNamespace = "dpdk-testing"
+	}
+}
+
+var _ = Describe("dpdk", func() {
+	var _ = Context("Run a sanity test on a worker", func() {
+		It("Should forward and receive packets", func() {
+			var out string
+			p := findPod()
```
fedepaol: Wouldn't it make sense for the test to create the pod it wants to test?
@SchSeba WDYT?

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+
+// checkRxTx parses the output from the DPDK test application
+// and verifies that packets have passed the NIC TX and RX queues
+func checkRxTx(out string) {
+	str := strings.Split(out, "\n")
+	for i := 0; i < len(str); i++ {
+		if strings.Contains(str[i], logEntry) {
+			i++
+			d := getNumberOfPackets(str, i)
+			Expect(d).Should(BeNumerically(">", 0), "number of received packets should be greater than 0")
```
fedepaol: I think it's better to use `To` instead of `Should` here

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+
+			i++
+			d = getNumberOfPackets(str, i)
+			Expect(d).Should(BeNumerically(">", 0), "number of transferred packets should be greater than 0")
```
fedepaol: Same

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+
+// findPod finds a pod running a DPDK application
+func findPod() *corev1.Pod {
+	listOptions := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{"deploymentconfig": "s2i-dpdk-app"}).String(),
+	}
+
+	p, err := client.Client.Pods(testDpdkNamespace).List(listOptions)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(len(p.Items)).ShouldNot(Equal(0), "no pods found")
```
fedepaol: `ToNot`

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+	var pod corev1.Pod
+	podReady := false
+	for _, pod = range p.Items {
+		if pod.Status.Phase == corev1.PodRunning {
+			podReady = true
+			break
+		}
+	}
+
+	Expect(podReady).Should(BeTrue(), fmt.Sprintf("the pod %s is not ready", pod.Name))
```
fedepaol: 'To(BeTrue()..`

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+	pods.WaitForCondition(client.Client, &pod, corev1.ContainersReady, corev1.ConditionTrue, 3*time.Minute)
+	return &pod
+}
+
+// getPodLog connects to a pod and fetches log
+func getPodLog(p *corev1.Pod) string {
```
fedepaol: Can you move this to the utils/pods package and have it as `pods.getLog(p *corev1.Pod)` ?

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+			checkRxTx(out)
+		})
+	})
+})
+
+// checkRxTx parses the output from the DPDK test application
+// and verifies that packets have passed the NIC TX and RX queues
+func checkRxTx(out string) {
+	str := strings.Split(out, "\n")
+	for i := 0; i < len(str); i++ {
```
fedepaol: This can be for `idx, s := range str`, and you can access str[idx+1] and str[idx+2] where you need them below.
At this point you can then remove the `index` parameter from `getNumberOfPackets`, have it accept a string (dpdkLogLine?) and remove the `i++`
Also, str can be named `lines`.

functests/dpdk/dpdk.go
```diff
@@ -0,0 +1,117 @@
...
+			Expect(d).Should(BeNumerically(">", 0), "number of transferred packets should be greater than 0")
+			break
+		}
+	}
+}
+
+// getNumber of packets parses the string (represented as a slice)
+// and returns an element representing the number of packets
+func getNumberOfPackets(s []string, index int) int {
+	r := strings.Fields(s[index])
```
fedepaol: Can we do some extra check to verify that the line is really what we expect (like asserting the first element)

functests/dpdk/dpdk.go
SchSeba: I will like to leave the test like this because we deploy the application as part of the feautre-deploy (demo folder).

fedepaol: I'd like to see these tests as less tied as we can to this repo, so we can eventually make them part of the sriov test suite or some other OpenShift test in the future.

But if not, let's give `findPod` a better name. Like `findDPDKWorkloadPod` or something along that line.

SchSeba: sounds good to me @dorzheh can you address this request?

functests/dpdk/dpdk.go
```diff
@@ -51,40 +51,38 @@ var _ = Describe("dpdk", func() {
...
-		if strings.Contains(str[i], logEntry) {
-			i++
-			d := getNumberOfPackets(str, i)
-			Expect(d).Should(BeNumerically(">", 0), "number of received packets should be greater than 0")
-
-			i++
-			d = getNumberOfPackets(str, i)
-			Expect(d).Should(BeNumerically(">", 0), "number of transferred packets should be greater than 0")
+	lines := strings.Split(out, "\n")
+	for i, line := range lines {
```
fedepaol: Assert here that len(lines) >= 3

fedepaol:
One little nit, then it's good. Looks awesome, thanks!

functests/dpdk/dpdk.go
dorzheh: @fedepaol, why we need to add this assertion. We already have two verifications:
1. We check whether log contains an appropriate header (and if it does, we are good to go)
2. We check the first field for an appropriate pattern

fedepaol: Well, my reasoning was that if the file is malformed, the test would panic in `d := getNumberOfPackets(lines[i+1], "RX")` as the `i+1` assumes the next line to be there.

PR #61, opened 2020-02-03
Skip negative test if no non-sctp nodes are available

fedepaol:


fedepaol:
I don't get why this is failing to build the docker image while https://github.com/openshift-kni/cnf-features-deploy/pull/62 worked flawlessly

PR #63, opened 2020-02-03
Skip negative test if no non-sctp nodes are available.

fedepaol:


functests/sctp/sctp.go
```diff
@@ -65,6 +65,10 @@ var _ = Describe("sctp", func() {
 					LabelSelector: "node-role.kubernetes.io/worker,!" + strings.Replace(sctpNodeSelector, "=", "", -1),
 				})
 				Expect(err).ToNot(HaveOccurred())
+				if len(nodes.Items) == 0 {
```
slintes: you can remove line 72 with this

fedepaol:
> this is just a thought. i'm unsure if it makes sense or not.
> 
> should our expectation around having non cnf feature enabled workers present be something we feed into the functests at execution time as an argument?
> 
> reason being, it means we declare what tests should and shouldn't run rather than detect them based on the cluster's configuration.

Ah, good point. You mean, trying to avoid skipping when we shouldn't.
Something like passing the # of non cnf-workers from outiside would make the trick.

PR #69, opened 2020-02-06
Check if the default service account is created for the test namespace.

fedepaol:
This is to skip random "looking up service account sctptest/default: serviceaccount \"default\" not found" errors.


slintes:
good idea, hope it helps :)
feel free to remove the hold when CI is happy

/hold
/lgtm

fedepaol:
> good idea, hope it helps :)
> feel free to remove the hold when CI is happy
> 
> /hold
> /lgtm

If it does not, it's material for a bug :-). In either case, we'll sort it out.

Issue #75, opened 2020-02-10
Suggested change in the README file 


dorzheh:
In the file, https://github.com/openshift-kni/cnf-features-deploy/blob/master/README.md appears that  _FEATURES_ENVIRONMENT=demo make functests is_ the command for executing the relevant tests.
However, it seems the correct command is  _FEATURES=test1 test2 testN make functests_

fedepaol:
Ah thanks!
Will change that.

slintes:
we have both `FEATURES_ENVIRONMENT` and `FEATURES`, and both have defaults set on top of the `Makefile` [0].
The default for `FEATURES_ENVIRONMENT` is the CI config override directory (`e2e-gcp`), that's why we override it in the readme to the `demo` dir.
But I agree that it makes sense to mention that you can also override `FEATURES` :+1: 

[0] https://github.com/openshift-kni/cnf-features-deploy/blob/master/Makefile

fedepaol:
> we have both `FEATURES_ENVIRONMENT` and `FEATURES`, and both have defaults set on top of the `Makefile` [0].
> The default for `FEATURES_ENVIRONMENT` is the CI config override directory (`e2e-gcp`), that's why we override it in the readme to the `demo` dir.
> But I agree that it makes sense to mention that you can also override `FEATURES` +1
> 
> [0] https://github.com/openshift-kni/cnf-features-deploy/blob/master/Makefile

Yes, but the point is, in [testing](https://github.com/openshift-kni/cnf-features-deploy/blob/master/README.md#testing) we suggest to set `FEATURES_ENVIRONMENT` which does not make sense there.

slintes:
fixed in #76 

/close

PR #76, opened 2020-02-11
Fix testing section stating that we need FEATURES 

fedepaol:
Solves https://github.com/openshift-kni/cnf-features-deploy/issues/75

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,4 +47,4 @@ In case a feature needs to be instrumented in a custom way, we expect a `test.sh
 
 A common use case is to rely on the feature's tier 1 tests taken from an external repo and [performance](/external-tests/performance/test.sh) provides a good example for that.
 
-`FEATURES_ENVIRONMENT=demo make functests` is the command for executing the relevant tests.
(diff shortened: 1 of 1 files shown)


slintes:
I would keep FEATURES_ENVIRONMENT. Testing != CI, and the latter is default for FEATURES_ENVIRONMENT. You might want to run tests on another environment...

fedepaol:
> I would keep FEATURES_ENVIRONMENT. Testing != CI, and the latter is default for FEATURES_ENVIRONMENT. You might want to run tests on another environment...

Yes, but if you call `make functests` the `FEATURES_ENVIRONMENT` variable is of no use at all.

slintes:
> > I would keep FEATURES_ENVIRONMENT. Testing != CI, and the latter is default for FEATURES_ENVIRONMENT. You might want to run tests on another environment...
> 
> Yes, but if you call `make functests` the `FEATURES_ENVIRONMENT` variable is of no use at all.

that is a very fair point indeed :smile: 

/lgtm

PR #81, opened 2020-02-17
Proper number of nodes message

e-minguez:
Minor fix to properly output the number of nodes labeled as worker-cnf

### DIFF
--- a/hack/setup-test-cluster.sh
+++ b/hack/setup-test-cluster.sh
@@ -5,10 +5,10 @@ set -e
 # expect oc to be in PATH by default
 export OC_TOOL="${OC_TOOL:-oc}"
 
-# Label 2 worker nodes as worker-cnf
-echo "[INFO]: Labeling 2 worker nodes with worker-cnf"
+# Label worker nodes as worker-cnf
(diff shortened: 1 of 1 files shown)


hack/setup-test-cluster.sh
```diff
@@ -5,8 +5,10 @@ set -e
 # expect oc to be in PATH by default
 export OC_TOOL="${OC_TOOL:-oc}"
 
-# Label 2 worker nodes as worker-cnf
-echo "[INFO]: Labeling 2 worker nodes with worker-cnf"
+# Label worker nodes as worker-cnf
+numnodes=$(${OC_TOOL} get nodes --selector='node-role.kubernetes.io/worker' \
+  --selector='!node-role.kubernetes.io/master' -o name | wc -l)
+echo "[INFO]: Labeling ${numnodes} worker nodes with worker-cnf"
```
fedepaol: We are not labelling all the workers as workers-cnf (only two, actually).
I think it would be better to do this after the labelling, counting the actual labelled workers (so it will keep working even if we change the labelling logic)

e-minguez: Thanks. I think it is better now :)

fedepaol:
Failures not related to this change.
/approve

PR #82, opened 2020-02-18
Remove the creation of ~/bin and .bashrc as pods run /bin/sh

fedepaol:
Trying to fix CI failures.

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -19,12 +19,9 @@ RUN dnf -y install \
     python2 \
     && dnf clean all
 
-RUN mkdir ~/bin && \
-    # install Go using gimme
-    curl -sL -o /usr/local/bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme && \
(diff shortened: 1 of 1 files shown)


fedepaol:
I thikn

> Do they invoke like "/bin/sh -c" ... ?
> That explains why .bashrc is ignored.
> anyway it looks good to me, bar the fact I don't know what "GIMME_ENV" was used for

I think it was meant to display gimme setup.

fromanirh:
> > Do they invoke like "/bin/sh -c" ... ?
> > That explains why .bashrc is ignored.
> > anyway it looks good to me, bar the fact I don't know what "GIMME_ENV" was used for
> 
> Edit: I misunderstood the question. GIMME_ENV was setting up PATH to /root/.gimme/ installed binaries and it was being ignored anyway. That's why we had to explicitly add `ENV PATH=$PATH:/root/.gimme/versions/go"$GOVERSION".linux.amd64/bin:$GOBIN` so my assumption is that it's safe to remove it. Ci is broken anyway so it's worth trying.

makes sense.

PR #88, opened 2020-02-20
fix: use resources for dpdk demo environment

tripledes:
* Using resources allows to create overlays and still permits to deploy
it directly.

Signed-off-by: Sergi Jimenez <sjr@redhat.com>

### DIFF
--- a/feature-configs/demo/dpdk/kustomization.yaml
+++ b/feature-configs/demo/dpdk/kustomization.yaml
@@ -5,7 +5,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-bases:
+resources:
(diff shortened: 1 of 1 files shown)


SchSeba:
> weird, i just realized there's no base configs for dpdk.
> @SchSeba was this an oversight or on purpose?

@davidvossel this was on purpose because there is no base for dpdk. this is just a demo workload running on the cluster.
I also create a PR to update the comment there we also need to performance operator, not just the sriov.
PR: https://github.com/openshift-kni/cnf-features-deploy/pull/93

tripledes:
Failed tests don't seem to be related to this PR, do they ?

SchSeba:
@tripledes they are not related

fedepaol:
The error is weird though. let's check if it pops out again.

PR #94, opened 2020-02-24
Fix dpdk workload when using version 18.11

SchSeba:
This PR add the --iova-mode=va flag.
This flag is must when running dpdk version 18.11 on a netdevice driver like mellanox.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1785933


fedepaol:
Question: will it still work in case of mellanox?

SchSeba:
@fedepaol yes this works with intel.

BTW in newer dpdk version 19.11 there is not need for this flag. the default it `va` and not `pa`

PR #96, opened 2020-02-24
Use different junit filename than external tests.

slintes:
Currently the `unit_report.xml` junit test result file of local tests will be overwritten by the performance external tests.

With this PR
- the filename of the local tests is modified so it won't be overwritten
- the junit file of external tests will be renamed after running in case they use `unit_report.xml`, in order to prevent multiple external tests overwriting their junit files.

@fedepaol


slintes:
/hold

let's see if this works

slintes:
looks good now, we have `unit_report_local.xml` and `unit_report_external_performance.xml`

/hold cancel

PR #106, opened 2020-02-29
Fixed MCP by removing null value

slintes:
Fails now because MCO added openAPIV3Schema to their CRDs

Signed-off-by: Marc Sluiter <msluiter@redhat.com>


fedepaol:
CI seemed to be stuck

PR #108, opened 2020-03-07
Added new numa.topologyPolicy to performance profile for CI

slintes:
See https://github.com/openshift-kni/performance-addon-operators/pull/117
Fixes the last failing functest in periodic jobs

No backport needed for now

### DIFF
--- a/feature-configs/base/performance/performance_profile.yaml
+++ b/feature-configs/base/performance/performance_profile.yaml
@@ -10,3 +10,5 @@ spec:
     defaultHugepagesSize: "1G"
   realTimeKernel:
     enabled: true
+  numa:
+    topologyPolicy: "best-effort"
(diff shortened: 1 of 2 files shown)


feature-configs/base/performance/performance_profile.yaml
```diff
@@ -10,3 +10,5 @@ spec:
     defaultHugepagesSize: "1G"
   realTimeKernel:
     enabled: true
+  numa:
+    topologyPolicy: "best-effort"
```
fedepaol: nit: new line

MarSik:
@slintes A question, isn't best-effort the default?
Also about the backport, this seems to be affecting downstream testing that pulls the 4.4 branch as git submodule. So we might need it.

slintes:
> @slintes A question, isn't best-effort the default?

yes, I added it to base for showing that we have it.

> Also about the backport, this seems to be affecting downstream testing that pulls the 4.4 branch as git submodule. So we might need it.

The feature wasn't backported, so it should not affect 4.4 at all.

feature-configs/base/performance/performance_profile.yaml
slintes: meh, thx, fixed

slintes:
I see green lights :)
Can I have a lgtm please @MarSik @fedepaol ?

PR #109, opened 2020-03-09
Temporarily disable RT kernel

slintes:
Reason: https://bugzilla.redhat.com/show_bug.cgi?id=1811630


slintes:
`should be enabled with the single-numa-node policy [It]` - expected, see #108 
`should have RT kernel enabled [It]`  - expected, caused by this change

`Non-guaranteed POD can work on any CPU [It]` - unexpected
`Guaranteed POD should work on isolated cpu [It]` - unexpected

/retest

fedepaol:
/lgtm

Feel free to remove the hold if you get to this before me

slintes:
only the 2 expected failures now

/hold cancel

PR #111, opened 2020-03-09
Revert "Temporarily disable RT kernel"

slintes:
Reverts openshift-kni/cnf-features-deploy#109

RT kernel RPMs are back, see https://github.com/openshift-kni/performance-addon-operators/pull/136


slintes:
only 1 expected test failure `should be enabled with the single-numa-node policy` which is fixed in #108 

/hold cancel

PR #112, opened 2020-03-10
Update dpdk s2i build

SchSeba:
This commit updates the dpdk s2i building process.

Now we are also building the testpmd application and not using the one that is shipped by the dkdk package.


fedepaol:
Just one comment. My understanding here is that we are building a custom testpmd instead of using the one available with the image.
If that is the case, can you just rename the executable in a different way? like customtestpmd / sampleapp or something?

SchSeba:
@fedepaol done can you please take another look :)

PR #129, opened 2020-03-25
Add skip flag to ptp tests if less than 2 slaves, Nodes selected

nikita133:


slintes:
/ok-to-test
/lgtm

just to be sure, I leave the approval to someone from networking team :)

PR #136, opened 2020-04-03
Fix deployment and tests

slintes:
We are running into 2 issues after OCP has been rebased on k8s 1.18:

1. `ocp apply -f` fails to deploy ANY resource if ONE of them has an unknown kind. So apply all one by one, else our loop-until-succeed deployment doesn't work (profile can't be deployed because CRD doesn't exist in 1st iteration(s). This is a known and fixed bug in kubectl, so a temporary workaround only until the fix landed in OCP. See [0].

2. `oc rsh` aka `kubectl exec` write a deprecation note now to stderr:
`kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl kubectl exec [POD] -- [COMMAND] instead.`
So let's add add "--". But it didn't work with "oc rsh", so also switch to "oc exec -i"

[0] https://github.com/kubernetes/kubernetes/pull/89539

### DIFF
--- a/functests/utils/nodes/nodes.go
+++ b/functests/utils/nodes/nodes.go
@@ -64,11 +64,13 @@ func ExecCommandOnMachineConfigDaemon(cs *testclient.ClientSet, node *corev1.Nod
 	}
 
 	initialArgs := []string{
-		"rsh",
+		"exec",
+		"-i",
 		"-n", testutils.NamespaceMachineConfigOperator,
 		"-c", testutils.ContainerMachineConfigDaemon,
 		"--timeout", "30",
 		mcd.Name,
+		"--",
 	}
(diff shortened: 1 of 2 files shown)


fedepaol:
Thanks!

/lgtm

slintes:
/test ci/prow/e2e-gcp

similar test fixes are merged now in the performance operator repo

slintes:
There is one known remaining test failure in the performance tests in the e2e-gcp lane, caused by an external issue.
Will discuss with the team next week if we disable it temporarly.
Otherwise this looks good, that line at least.
I suggest to investigate errors in the oven lane in a follow if needed.

/override ci/prow/e2e-gcp

PR #137, opened 2020-04-06
Add tests failure reporter.

fedepaol:
This is a generic reporter triggered whenever a test fails. 
It's heavily inspired by the kubevirt reporter https://github.com/kubevirt/kubevirt/tree/master/tests/reporter .
It dumps information about nodes, pods and configured custom crds.


hack/run-functests.sh
```diff
@@ -29,7 +29,7 @@ fi
 echo "Running local tests"
 FOCUS=$(echo "$FEATURES" | tr ' ' '|') 
 echo "Focusing on $FOCUS"
-if ! GOFLAGS=-mod=vendor ginkgo --focus=$FOCUS functests -- -junit /tmp/artifacts/unit_report_local.xml; then
+if ! GOFLAGS=-mod=vendor ginkgo --focus=$FOCUS functests -- -junit /tmp/artifacts/unit_report_local.xml -report ./tmp/artifacts/report.log; then
```
slintes, line 32: may add `local` to the report filename, similar to the unit filename? We might get more of them if we use a custom reporter in the external repos as well.

functests/utils/k8sreporter/reporter.go
```diff
@@ -0,0 +1,205 @@
...
+	r.Cleanup()
+}
+
+func (r *KubernetesReporter) SpecWillRun(specSummary *types.SpecSummary) {
+}
+
+func (r *KubernetesReporter) SpecDidComplete(specSummary *types.SpecSummary) {
+	r.Lock()
+	defer r.Unlock()
+	fmt.Println("FEDE did complete", specSummary.ComponentTexts)
```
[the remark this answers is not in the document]
fedepaol: oops

hack/run-functests.sh
fedepaol: Will change, but my idea is to incorporate externals as vendored dependencies (see https://github.com/openshift-kni/cnf-features-deploy/pull/138 ) so on the mid-term I expect to have only one test suite.

fedepaol:
@slintes CI is not helping this days, however this https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift-kni_cnf-features-deploy/137/pull-ci-openshift-kni-cnf-features-deploy-master-e2e-gcp-ovn/316/artifacts/e2e-gcp-ovn/report_local.log is a report example.

functests/utils/k8sreporter/example/main.go
```diff
@@ -0,0 +1,55 @@
+package main
```
SchSeba, line 1: done we need this file?

fedepaol: I would say yes, for multiple reasons.
- My idea is that at some point we can even ship this as a library that consumed by others (I put it here mostly for delivery speed reasons)
- It provides a nice example of how to use / configure the reporter
- It's an easy way to check / debug the reporter without having to run the whole suite

PR #138, opened 2020-04-06
Provide a single entry point for the test suites

fedepaol:
This PR is the first step in the directoin of containerizing cnf-tests.

It provides:
- a way to build the test executable
- Dockerfiles for both u/s + openshift builds
- It consumes sriov tests from the sriov-operator repository as opposed to cloning the repo


fedepaol:
/hold
Need to validate sriov tests against a proper cluster.

hack/build-test-bin.sh
```diff
@@ -0,0 +1,18 @@
...
+
+GOPATH="${GOPATH:-~/go}"
+export PATH=$PATH:$GOPATH/bin
+
+if ! which gingko; then
+	echo "Downloading ginkgo tool"
+	go install github.com/onsi/ginkgo/ginkgo
+fi
+
+ginkgo build -mod=vendor ./functests
```
SchSeba, line 18: nit: add a new line please.

SchSeba: another comment can we please move the output file to a bin directory and add it to the `.gitignore` file

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.13 AS builder
+WORKDIR /go/src/github.com/openshift-kni/cnf-features-deploy
```
SchSeba, line 2: to speed up the building process can you take a look on this dockerfile 

https://github.com/openshift/linuxptp-daemon/blob/master/Dockerfile#L8

the first part will copy the go.mod file and download all the packages this will cache that layer for the next time we try to build the image

Makefile
```diff
@@ -62,3 +65,7 @@ feature-wait:
 test-bin:
 	@echo "Making test binary"
 	hack/build-test-bin.sh
+
+image-local:
+	@echo "Making image local"
+	$(IMAGE_BUILD_CMD) build --no-cache -f Dockerfile -t cnf-tests-local .
```
SchSeba: nit: new line here please

Makefile
```diff
@@ -66,6 +66,6 @@ test-bin:
 	@echo "Making test binary"
 	hack/build-test-bin.sh
 
-image-local:
-	@echo "Making image local"
-	$(IMAGE_BUILD_CMD) build --no-cache -f Dockerfile -t cnf-tests-local . 
\ No newline at end of file
+cnf-tests-local:
+	@echo "Making cnf-tests local"
+	$(IMAGE_BUILD_CMD) build --no-cache -f cnf-tests/Dockerfile -t cnf-tests-local .
```
SchSeba, line 71: should we have a variable for the Dockerfile name?

Dockerfile
slintes: there should be no need for downloading anything when vendoring? :thinking:

Dockerfile.openshift
```diff
@@ -0,0 +1,23 @@
+# This dockerfile is specific to building the network metrics daemon for OpenShift
```
slintes, line 1: is that comment a copy paste artifact? :)

fedepaol: Ops (2).

Makefile
fedepaol: Not sure about that. This is intended to build it locally, and I expect it to use this dockerfile. The other one is going to be used by CI to build the "openshift" version

Dockerfile
fedepaol: Agree

hack/build-test-bin.sh
fedepaol: done

fedepaol:
Validated, sriov tests are getting executed

cnf-tests/bin/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
slintes, line 4: I think adding `bin` to the root .gitignore is easier?

fedepaol: I think it won't work because git won't let me commit empty folders. 
See https://stackoverflow.com/questions/115983/how-can-i-add-an-empty-directory-to-a-git-repository

slintes: do you need to commit it? :thinking:

fedepaol: I think so, otherwise the bin folder won't persist if you checkout on a new machine.
Not sure I understand what your doubt is...

slintes: I just don't understand why you want to persist the folder.
I would just `mkdir -p` it when needed and ignore it completely.
But maybe it's just a matter of taste. Will lgtm and hold, feel free to remove the hold when you prefer to keep it as it is :slightly_smiling_face: 

/lgtm
/hold

fedepaol: I'm fine with either of them.

I changed it as per @SchSeba comment: https://github.com/openshift-kni/cnf-features-deploy/pull/138#discussion_r404706095

fedepaol: Uhm, now I realize I mis-interpreted it. Changing.

cnf-tests/Dockerfile.openshift
```diff
@@ -0,0 +1,23 @@
...
+ENV PKG_NAME=github.com/openshift-kni/cnf-features-deploy
+ENV PKG_PATH=/go/src/$PKG_NAME
+RUN mkdir -p $PKG_PATH
+
+COPY . $PKG_PATH/
+WORKDIR $PKG_PATH
+
+RUN make build-bin
+
+WORKDIR /
```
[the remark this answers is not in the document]
fedepaol: I guess it's to restore the workdir as it was before line 10.
I "took inspiration" from another dockerfile with similar purpouse

Makefile
```diff
@@ -10,7 +11,9 @@ export FEATURES_ENVIRONMENT?=e2e-gcp
 	golint \
 	govet \
 	ci-job \
-	feature-deploy
+	feature-deploy \
+	image-local \
```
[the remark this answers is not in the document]
fedepaol: Good catch, it's because I renamed it afterwards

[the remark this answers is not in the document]
fedepaol: Done

Makefile
```diff
@@ -10,7 +11,9 @@ export FEATURES_ENVIRONMENT?=e2e-gcp
...
+	test-bin \
```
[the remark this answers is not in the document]
fedepaol: Done

PR #139, opened 2020-04-06
Updating qe-tools

dshchedr:


go.sum
```diff
@@ -1191,7 +1191,7 @@ k8s.io/utils v0.0.0-20190920012459-5008bf6f8cd6/go.mod h1:sZAwmy6armz5eXlNoLmJcl
 k8s.io/utils v0.0.0-20191114184206-e782cd3c129f/go.mod h1:sZAwmy6armz5eXlNoLmJcl4F1QuKu7sr+mFQ0byX7Ew=
 k8s.io/utils v0.0.0-20200109141947-94aeca20bf09 h1:sz6xjn8QP74104YNmJpzLbJ+a3ZtHt0tkD0g8vpdWNw=
 k8s.io/utils v0.0.0-20200109141947-94aeca20bf09/go.mod h1:sZAwmy6armz5eXlNoLmJcl4F1QuKu7sr+mFQ0byX7Ew=
-kubevirt.io/qe-tools v0.1.4 h1:q4xvpLIyO6Ljl7u4Bu27EM2guNAuPolOqwCYJ2QeIik=
+kubevirt.io/qe-tools v0.1.4 h1:t1O6OtLV1auU3J60XXDnzX2JxNQ5kWJTQAHaBUGUbTY=
```
slintes, line 1194: New hash for same version? That's really bad. Was there a new release with the same version? That should never happen. That makes this file useless, and I would expect that it breaks builds when you not use vendoring (go proxy saves hashes as well IIRC)

dshchedr: exactly, there were some updates in qe-tools, but version still the same
Not sure why, but Nelly decided to keep previous version

PR #141, opened 2020-04-15
Consume performance tests as external dependency

fedepaol:
Instead of cloning the performance operator repo, we consume the performance tests as an external dependency in order to have only one suite.


slintes:
besides the need for a rebase... there an IMHO huge disadvantage of importing tests like this: you don't get the latest version automatically :thinking: That being said, can you please update the performance test dependency? It should have some improvements for less flaky tests now.

fedepaol:
> besides the need for a rebase... there an IMHO huge disadvantage of importing tests like this: you don't get the latest version automatically thinking That being said, can you please update the performance test dependency? It should have some improvements for less flaky tests now.

Yes, CI wise is a disadvantage. We must care of updating this manually as we update the dependencies. 
On the other hand, it will result in tests being more reproducible and the cnf-tests image will benefit from this.
Will update the dependency

fedepaol:
Done, had to bump the version of the client-go and another dependency because of https://github.com/kubernetes/client-go/issues/741

functests/test_suite_test.go
```diff
@@ -75,12 +79,26 @@ var _ = BeforeSuite(func() {
...
 
 var _ = AfterSuite(func() {
 	err := testclient.Client.Namespaces().Delete(testutils.NamespaceTesting, &metav1.DeleteOptions{})
 	Expect(err).ToNot(HaveOccurred())
 	err = namespaces.WaitForDeletion(testclient.Client, testutils.NamespaceTesting, 5*time.Minute)
+	Expect(err).ToNot(HaveOccurred())
+
+	err = testclient.Client.Namespaces().Delete(perfUtils.NamespaceTesting, &metav1.DeleteOptions{})
+	Expect(err).ToNot(HaveOccurred())
+	err = namespaces.WaitForDeletion(testclient.Client, testutils.NamespaceTesting, 5*time.Minute)
```
slintes, line 100: `perfUtils.NamespaceTesting`

fedepaol: good catch!

PR #143, opened 2020-04-20
Fix the openshift docker file build command.

fedepaol:
Also, vendor the gingko executable so it can be used without downloading deps.

### DIFF
--- a/cnf-tests/Dockerfile.openshift
+++ b/cnf-tests/Dockerfile.openshift
@@ -9,9 +9,7 @@ RUN mkdir -p $PKG_PATH
 COPY . $PKG_PATH/
 WORKDIR $PKG_PATH
 
-RUN make build-bin
-
-WORKDIR /
+RUN make test-bin
 
 FROM openshift/origin-base
 COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/bin/cnftests /usr/bin/cnftests
--- a/hack/build-test-bin.sh
+++ b/hack/build-test-bin.sh
@@ -8,6 +8,8 @@ if ! which go; then
 fi
 
 GOPATH="${GOPATH:-~/go}"
+GOFLAGS="${GOFLAGS:-"-mod=vendor"}"
+
 export PATH=$PATH:$GOPATH/bin
 
 if ! which gingko; then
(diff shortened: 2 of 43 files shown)


hack/build-test-bin.sh
```diff
@@ -8,13 +8,15 @@ if ! which go; then
 fi
 
 GOPATH="${GOPATH:-~/go}"
+GOFLAGS="${GOFLAGS:-"-mod=vendor"}"
+
 export PATH=$PATH:$GOPATH/bin
 
 if ! which gingko; then
```
SchSeba, line 15: now that we use the tools package can you change this to build it from the vendor folder?

cnf-tests/Dockerfile.openshift
```diff
@@ -9,7 +9,7 @@ RUN mkdir -p $PKG_PATH
 COPY . $PKG_PATH/
 WORKDIR $PKG_PATH
 
-RUN make build-bin
+RUN make test-bin
 
 WORKDIR /
```
SchSeba, line 14: why we have this line here?

hack/build-test-bin.sh
fedepaol: it works by having GOFLAGS=-mod=vendor.
The issue is that in the openshift base image GOFLAGS is already set like that, hence line 12.
With that variable, all build happen using the fendor folder.

cnf-tests/Dockerfile.openshift
fedepaol: You must really hate this. You made the same question before :-)
https://github.com/openshift-kni/cnf-features-deploy/pull/138#discussion_r409425961

I took inspiration from other dockerfiles, I guess it's to restore WORKDIR to what it was before.

SchSeba: yeah but then we jump to another container this is only the builder container

hack/build-test-bin.sh
SchSeba: great thanks!

cnf-tests/Dockerfile.openshift
fedepaol: removed it

fedepaol:
Validated the dockerfile locally, CI is broken.
Overriding it.

PR #149, opened 2020-04-22
Export goflags so ginkgo gets built using vendored dependencies.

fedepaol:


fedepaol:
This is needed for d/s build.

fedepaol:
Pressed the red button as CI is known to fail and overriding would have taken 2 hours.

PR #157, opened 2020-04-27
Make sctpNodeSelectorLabel independent from configuration

marzianor9:


functests/sctp/sctp.go
```diff
@@ -483,3 +488,59 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
 	_, err := client.Client.NetworkPolicies(namespace).Create(policy)
 	return err
 }
+
+func findSCTPNodeSelector() {
+	mcList, err := client.Client.MachineConfigs().List(metav1.ListOptions{})
+	Expect(err).ToNot(HaveOccurred())
+
+	for _, mc := range mcList.Items {
+		if doesThisEnablesSctp(mc) == true {
```
fedepaol, line 497: No need to check  `== true`, just `if ... {`

functests/sctp/sctp.go
```diff
@@ -483,3 +488,59 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
...
+			findMCPByLabel(mc.ObjectMeta.Labels["machineconfiguration.openshift.io/role"])
+			break
+		}
+	}
+}
+
+func doesThisEnablesSctp(mc mcfgv1.MachineConfig) bool {
```
fedepaol, line 504: just `enablesSCTP`

functests/sctp/sctp.go
```diff
@@ -483,3 +488,59 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
...
+func findMCPByLabel(mcLabel string) {
+	mcpList, err := client.Client.MachineConfigPools().List(metav1.ListOptions{})
+	Expect(err).ToNot(HaveOccurred())
+
+	for _, mcp := range mcpList.Items {
+		for _, lsr := range mcp.Spec.MachineConfigSelector.MatchExpressions {
+			for _, value := range lsr.Values {
+				if value == mcLabel {
+					for key, label := range mcp.Spec.NodeSelector.MatchLabels {
+						sctpNodeSelector = key + "=" + label
```
fedepaol, line 528: Can you just return the mcp here (and an error if not found) and use it outside?
In this way, this function would not have side effects. I'd like to see the nodeList set with the result of `findSctpNodeSelector`

functests/sctp/sctp.go
```diff
@@ -483,3 +488,59 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
...
+		}
+	}
+}
+
+func doesThisEnablesSctp(mc mcfgv1.MachineConfig) bool {
+	firstPathFound := false
+	secondPathFound := false
+
+	for _, file := range mc.Spec.Config.Storage.Files {
+		if file.Path == sctpBlacklistPath {
```
fedepaol, line 509: Just a nit, but no need for the else

functests/sctp/sctp.go
```diff
@@ -483,3 +488,59 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
 	_, err := client.Client.NetworkPolicies(namespace).Create(policy)
 	return err
 }
+
+func findSCTPNodeSelector() {
```
fedepaol, line 492: Can we have this return the selector or an error if not found, and use the result in `beforeAll`?

functests/sctp/sctp.go
```diff
@@ -73,6 +79,11 @@ var _ = Describe("sctp", func() {
 
 		err = namespaces.Clean(testNamespace, client.Client)
 		Expect(err).ToNot(HaveOccurred())
+
+		newSCTPNodeSelector, err := findSCTPNodeSelector()
```
fedepaol, line 83: Let's be more consistent here.
If the env variable is set, it must have priority because the user willingly set it. Otherwise we try to do the discovery.

functests/sctp/sctp.go
```diff
@@ -483,3 +494,67 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
...
 	return err
 }
+
+func findSCTPNodeSelector() (string, error) {
+	mcList, err := client.Client.MachineConfigs().List(metav1.ListOptions{})
+	Expect(err).ToNot(HaveOccurred())
+
+	for _, mc := range mcList.Items {
+		if enablesSCTP(mc) {
+			newSCTPNodeSelector, err := findSCTPNodeSelectorByMCLabel(mc.ObjectMeta.Labels["machineconfiguration.openshift.io/role"])
```
fedepaol, line 504: Not sure if the role is mandatory, but if a mc does not have a label this is going to panic.
Can you check the existence and continue if it's not there?

functests/sctp/sctp.go
```diff
@@ -483,3 +494,67 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
...
+
+	for _, mcp := range mcpList.Items {
+		for _, lsr := range mcp.Spec.MachineConfigSelector.MatchExpressions {
+			for _, value := range lsr.Values {
+				if value == mcLabel {
+					for key, label := range mcp.Spec.NodeSelector.MatchLabels {
+						newSCTPNodeSelector := key + "=" + label
+
+						// check if there are at least two nodes or skip
+						sctpNodeList, err := client.Client.Nodes().List(metav1.ListOptions{
```
fedepaol, line 543: No need to do this (and the check of the len) here anymore.
This function declares it's purpouse in the name but here we are doing more than that. Also, one node is enough.
Just return the selector and leave to the caller the responsability to use it.

marzianor9: should i check the number of nodes on the caller side?
and is 1 node enough?

functests/sctp/sctp.go
```diff
@@ -43,12 +49,15 @@ type nodesInfo struct {
 
 func init() {
 	roleWorkerCNF := os.Getenv("ROLE_WORKER_CNF")
-	if roleWorkerCNF == "" {
-		roleWorkerCNF = "worker-cnf"
+	if roleWorkerCNF != "" {
+		sctpNodeSelector = fmt.Sprintf("node-role.kubernetes.io/%s=", roleWorkerCNF)
```
fedepaol, line 53: Bear in mind that here you are in `init()` which is the module initialization.
I would not rely on the fact that this part gets triggered before the client one. If the order is not correct, you have the k8s client not inited yet.
Moreover, you are doing the same in `beforeAll`

functests/sctp/sctp.go
```diff
@@ -70,6 +79,16 @@ var _ = Describe("sctp", func() {
 
 		err = namespaces.Clean(testNamespace, client.Client)
 		Expect(err).ToNot(HaveOccurred())
+
+		if sctpNodeSelector == "" {
```
fedepaol, line 83: Check the `roleWorkerCNF` variable here, if set use it for the selector.
Otherwise, do the discovery and fail if it returns an error with `Expect`

functests/sctp/sctp.go
```diff
@@ -480,3 +499,61 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
...
+	Expect(err).ToNot(HaveOccurred())
+
+	for _, mc := range mcList.Items {
+		if enablesSCTP(mc) {
+			mcLabel, found := mc.ObjectMeta.Labels["machineconfiguration.openshift.io/role"]
+			if !found {
+				continue
+			}
+
+			newSCTPNodeSelector, err := findSCTPNodeSelectorByMCLabel(mcLabel)
```
fedepaol, line 514: This means that you exit from the discovery if you find a mc that does not have the selector (while there could be another one down the list that fits our needs).

functests/sctp/sctp.go
```diff
@@ -480,3 +499,61 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
...
+					for key, label := range mcp.Spec.NodeSelector.MatchLabels {
+						newSCTPNodeSelector := key + "=" + label
+						return newSCTPNodeSelector, nil
+					}
+				}
+			}
+		}
+	}
+
+	return "", errors.New("cannot find SCTPNodeSelector")
```
fedepaol, line 558: nit: "Cannot find.."

functests/sctp/sctp.go
```diff
@@ -70,6 +74,13 @@ var _ = Describe("sctp", func() {
 
 		err = namespaces.Clean(testNamespace, client.Client)
 		Expect(err).ToNot(HaveOccurred())
+
+		if sctpNodeSelector == "" {
+			newSCTPNodeSelector, err := findSCTPNodeSelector()
```
fedepaol, line 79: can you just assign sctpNodeSelector here?

functests/sctp/sctp.go
```diff
@@ -480,3 +491,61 @@ func setupEgress(namespace, fromPod, toPod string, port int32) error {
...
+	Expect(err).ToNot(HaveOccurred())
+
+	for _, mc := range mcList.Items {
+		if enablesSCTP(mc) {
+			mcLabel, found := mc.ObjectMeta.Labels["machineconfiguration.openshift.io/role"]
+			if !found {
+				continue
+			}
+
+			newSCTPNodeSelector, err := findSCTPNodeSelectorByMCLabel(mcLabel)
```
fedepaol, line 506: Either res, or sctpNodeSelector. No need for "new"

PR #158, opened 2020-04-27
Use yum instead of dnf in the sctp tester dockerfile.

fedepaol:


fedepaol:
Tide says retesting, it's a dockerfile change that is not going to influence tests.
Pushing the red button.

PR #160, opened 2020-04-27
Use official images mirrored on quay.io/openshift-kni

fedepaol:


fedepaol:
/hold
Need to wait until the images are available on quay

PR #167, opened 2020-04-30
Move the deploy overlay to new "base"

fedepaol:
This means that base now only deploy the single feature, without providing a configuration.
The u/s CI use it directly (no gcp overlay anymore).

The demo overlay now reflects this new change, only deploying operators but NOT configuring them.

There is a new overlay, withconfig which provides the configuration that were provided by the old base.
The demo-configs overlay is functionally identical to the old demo. It may be used as a
deployment showcase, but not for running tests.

The other integration overlays now are based on these new "functionally equivalent" overlays.

The setup script do not label ptp nodes anymore, as it is supposed to be done in the test themselves.


fedepaol:
/cc @e-minguez 
as this may have side effects on integration.

feature-configs/base/performance/operator_subscription.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: performance-addon-operator-subscription
   namespace: openshift-performance-addon
 spec:
-  channel: alpha
+  channel: 0.0.1
```
yuvalk, line 7: I think this is unrelated to this PR

fedepaol: Yup, it was to make CI happy upfront. Will wait for the channel fix pr to land first.

yuvalk:
though I havent tested all overlay options

fedepaol:
/hold
Discussion still in progress on naming

slintes:
from the performance operator perspective lgtm, dir structure also looks good now, nice :+1: 
keeping the hold to give others the chance to review as well

/lgtm

yuvalk:
unrelatedly, I think we should move all the integration-kni, etc. somewhere else
not sure we even want them upstream, but even if we do, I dont think they belong under feature-configs
we can open a new directory for clusters or something
@slintes

PR #169, opened 2020-04-30
Provide support for running the dockerized version of the tests.

fedepaol:
By setting CNF_TESTS_IMAGE, make functests will execute them in the dockerized form.

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -10,7 +10,10 @@ if [ $? -ne 0 ]; then
 fi
 
 GOPATH="${GOPATH:-~/go}"
+
+export CONTAINER_MGMT_CLI="${CONTAINER_MGMT_CLI:-docker}"
 export PATH=$PATH:$GOPATH/bin
+export TESTS_REPORTS_PATH="${TESTS_REPORTS_PATH:-/tmp/artifacts/}"
 export failed=false
 export failures=()
 
@@ -22,10 +25,20 @@ fi
 echo "Running local tests"
 FOCUS=$(echo "$FEATURES" | tr ' ' '|')
 echo "Focusing on $FOCUS"
-
 export SUITES_PATH=cnf-tests/bin
 
-if ! cnf-tests/test-run.sh -ginkgo.focus=$FOCUS -junit /tmp/artifacts/ -report /tmp/artifacts/; then
+mkdir -p "$TESTS_REPORTS_PATH"
+
+if [ "$CNF_TESTS_IMAGE" != "" ]; then
+  cp -f "$KUBECONFIG" _cache/kubeconfig
+  echo "Running dockerized version via $CNF_TESTS_IMAGE"
+  EXEC_TESTS="$CONTAINER_MGMT_CLI run -v $(pwd)/_cache/:/kubeconfig:Z -v $TESTS_REPORTS_PATH:/reports:Z -e KUBECONFIG=/kubeconfig/kubeconfig $CNF_TESTS_IMAGE /usr/bin/test-run.sh \
+       -ginkgo.focus $FOCUS -junit /reports/ -report /reports/"
+else
+  EXEC_TESTS="cnf-tests/test-run.sh -ginkgo.focus=$FOCUS -junit $TESTS_REPORTS_PATH -report $TESTS_REPORTS_PATH"
+fi
+
+if ! $EXEC_TESTS; then
   failed=true
   failures+=( "Tier 2 tests for $FEATURES" )
 fi


hack/run-functests.sh
```diff
@@ -10,6 +10,8 @@ if [ $? -ne 0 ]; then
 fi
 
 GOPATH="${GOPATH:-~/go}"
+
+export IMAGE_RUN_CMD="${IMAGE_RUN_CMD:-docker}"
```
yuvalk, line 14: CONTAINER_MGMT_CLI ?
image_run_cmd is only understandable when you see the value :-)

hack/run-functests.sh
```diff
@@ -25,9 +27,20 @@ echo "Focusing on $FOCUS"
...
+  if ! $IMAGE_RUN_CMD run -v $(pwd)/_cache/:/kubeconfig -v /tmp/artifacts/:/reports -e KUBECONFIG=/kubeconfig/kubeconfig $CNF_TESTS_IMAGE /usr/bin/test-run.sh \
+       -ginkgo.focus $FOCUS -junit /reports/ -report /reports/; then
+    failed=true
+    failures+=( "Tier 2 tests for $FEATURES" )
+  fi
+else
+  if ! cnf-tests/test-run.sh -ginkgo.focus=$FOCUS -junit /tmp/artifacts/ -report /tmp/artifacts/; then
+    failed=true
+    failures+=( "Tier 2 tests for $FEATURES" )
+  fi
```
yuvalk, line 43: maybe better unify these so:
if [ $CNF_TESTS_IMAGE ]; then
    EXEC_TESTS=$IMAGE_RUN_CM run -v ....
else
    EXEC_TESTS=cnf-tests/test-run.sh ...
fi

if ! $EXEC_TESTS; then
    failed=true
    failures+=( "Tier 2 tests for $FEATURES" )
fi

which are the same in both cases..

hack/run-functests.sh
```diff
@@ -25,9 +27,20 @@ echo "Focusing on $FOCUS"
 
 export SUITES_PATH=cnf-tests/bin
 
-if ! cnf-tests/test-run.sh -ginkgo.focus=$FOCUS -junit /tmp/artifacts/ -report /tmp/artifacts/; then
-  failed=true
-  failures+=( "Tier 2 tests for $FEATURES" )
+if [ $CNF_TESTS_IMAGE != "" ]; then
+  cp -f $KUBECONFIG _cache/kubeconfig
```
yuvalk, line 31: what if that directory does not exists?

yuvalk, line 31: do we need to add it (_cache) to .gitignore?

hack/run-functests.sh
```diff
@@ -25,9 +27,20 @@ echo "Focusing on $FOCUS"
...
+  echo "Running dockerized version via $CNF_TESTS_IMAGE"
+
+  if ! $IMAGE_RUN_CMD run -v $(pwd)/_cache/:/kubeconfig -v /tmp/artifacts/:/reports -e KUBECONFIG=/kubeconfig/kubeconfig $CNF_TESTS_IMAGE /usr/bin/test-run.sh \
```
yuvalk, line 34: here too, need to take care of cases /tmp/artifacts does not exists

hack/run-functests.sh
```diff
@@ -25,9 +27,20 @@ echo "Focusing on $FOCUS"
 
 export SUITES_PATH=cnf-tests/bin
 
-if ! cnf-tests/test-run.sh -ginkgo.focus=$FOCUS -junit /tmp/artifacts/ -report /tmp/artifacts/; then
-  failed=true
-  failures+=( "Tier 2 tests for $FEATURES" )
+if [ $CNF_TESTS_IMAGE != "" ]; then
```
yuvalk, line 30: missing "" around $CNF_TESTS_IMAGE
but actually a more resilient way to test if variable exist is:
```
if [ -n ${CNF_TESTS_IMAGE+x} ]; then
```

fedepaol: What is the +x supposed to do?

hack/run-functests.sh
fedepaol: It's already added.

fedepaol: It's created in the makefile

hack/run-functests.sh
fedepaol: Will do.

hack/run-functests.sh
fedepaol: It's in the same line, so I don't think is too obscure. Will change though.

hack/run-functests.sh
fedepaol: Will make sure.

hack/run-functests.sh
yuvalk: it might be, further down the line..

hack/run-functests.sh
yuvalk: ah, ok, though it wont be created if hack/run_functests.sh is executed separately..

hack/run-functests.sh
yuvalk: if CNF_TESTS_IMAGE exists, the "+" will replace it with "x" so no matter what the value is, it wont interfere with parsing the if statement.

this way also distinguish between unset variable and variable that equals ""

i guess a
-n "${CNF_TESTS_IMAGE}" 
is also fine in this case (as an empty value is meaningless
but actually gives a way of settinga default
ie - unset - run locally, 
set to "" mean use default 
anything else - use that..

hack/run-functests.sh
yuvalk, line 34: need to add :Z to kubeconfig
ie
```
-v $(pwd)/_cache/:/kubeconfig:Z
```
otherwise you get permission denied from within the container

hack/run-functests.sh
yuvalk: it works only with -z though
so either keep as is :-)
or do
! -z

hack/run-functests.sh
yuvalk: for reports too

hack/run-functests.sh
fedepaol: I'll keep as it is wrapping with ""

PR #170, opened 2020-04-30
Fix the performance operator subscription channel by moving it to 0.0.1

fedepaol:


slintes:
obsolete after https://github.com/openshift-kni/cnf-features-deploy/pull/171

/close

PR #174, opened 2020-05-03
Remove temp file from last PR

SchSeba:
This Commit remove a file that was merged


yuvalk:
> This Commit remove a file that was merged
> 
> Signed-off-by: Sebastian Sch [sebassch@gmail.com](mailto:sebassch@gmail.com)

better explain in the commit the why (ie remove a file that was added by mistake or something of the sort)
git log should be read like a novel 
with the why we did things (vs that what - which can [hopefully] be read from code..)

PR #175, opened 2020-05-04
Collect pod logs for all the operators involved and for all test namespaces.

fedepaol:
This includes catching test + operators pods, and adding their relevant crds to the list to be dumped.

### DIFF
--- a/functests/utils/reporter.go
+++ b/functests/utils/reporter.go
@@ -1,0 +1,69 @@
+package utils
+
+import (
+	"os"
+
+	perfUtils "github.com/openshift-kni/performance-addon-operators/functests/utils"
+	ptpUtils "github.com/openshift/ptp-operator/test/utils"
+	sriovNamespaces "github.com/openshift/sriov-network-operator/test/util/namespaces"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/openshift-kni/cnf-features-deploy/functests/dpdk"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/k8sreporter"
+
+	performancev1alpha1 "github.com/openshift-kni/performance-addon-operators/pkg/apis/performance/v1alpha1"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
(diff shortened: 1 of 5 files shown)


functests/utils/reporter.go
```diff
@@ -0,0 +1,67 @@
...
+		NamespaceTesting:                   true,
+		perfUtils.NamespaceTesting:         true,
+		dpdk.TestDpdkNamespace:             true,
+		sriovNamespaces.Test:               true,
+		ptpUtils.NamespaceTesting:          true,
+	}
+
+	crds := []k8sreporter.CRData{
+		{Cr: &mcfgv1.MachineConfigPoolList{}},
+		{Cr: &ptpv1.PtpConfigList{}},
```
SchSeba, line 44: In ptp you have two more CR

```
nodeptpdevices.ptp.openshift.io                             2020-04-29T08:15:53Z
ptpconfigs.ptp.openshift.io                                 2020-04-29T08:15:53Z
ptpoperatorconfigs.ptp.openshift.io                         2020-04-29T08:15:53Z
```

go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	github.com/openshift/machine-config-operator v4.2.0-alpha.0.0.20190917115525-033375cbe820+incompatible
 	github.com/openshift/ptp-operator v0.0.0-20200421195159-92f3d20c48f9
 	github.com/openshift/sriov-network-operator v0.0.0-20200428080346-d046d9acc3dd
+	github.com/openshift/sriov-tests v0.0.0-20200430133953-189e0fb45a7b
```
SchSeba, line 30: This is not good we should not use the sriov-tests any more.

sriov-tests is only for the 4.4 branch

functests/utils/reporter.go
```diff
@@ -0,0 +1,67 @@
+package utils
+
+import (
+	"os"
+
+	perfUtils "github.com/openshift-kni/performance-addon-operators/functests/utils"
+	ptpUtils "github.com/openshift/ptp-operator/test/utils"
+	sriovNamespaces "github.com/openshift/sriov-tests/pkg/util/namespaces"
```
SchSeba, line 8: This is not good we should not use the sriov-tests any more.

sriov-tests is only for the 4.4 branch

PR #176, opened 2020-05-05
Fix image registry env variable doc, add a section about running with podman

fedepaol:


fromanirh:
thanks for caring about podman, appreciated.

fedepaol:
> thanks for caring about podman, appreciated.

It's mostly because I got bitten by it :-)

PR #193, opened 2020-05-08
[release-4.5] Update perf operator deployment to correct 4.5 image

slintes:


slintes:
overriding the origin job, not sure if tide will find out alone that it isn't mandatory anymore :man_shrugging: 

/override ci/prow/e2e-gcp-origin

PR #194, opened 2020-05-08
Update perf operator deployment to correct 4.6 image

slintes:
Note: 4.6.0 channel does not exist yet

### DIFF
--- a/feature-configs/deploy/performance/operator_subscription.yaml
+++ b/feature-configs/deploy/performance/operator_subscription.yaml
@@ -4,7 +4,7 @@ metadata:
   name: performance-addon-operator
   namespace: openshift-performance-addon
 spec:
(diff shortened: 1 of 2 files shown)


feature-configs/deploy/performance/operator_subscription.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: performance-addon-operator
   namespace: openshift-performance-addon
 spec:
-  channel: 0.0.1
+  channel: 4.5.0
```
fedepaol, line 7: Is the channel name correct?

fedepaol: Uhm, guess I should have read the note.

PR #200, opened 2020-05-12
Vendor ptp, sriov and performance from 4.5

fedepaol:
This vendors the latest change on performance, sriov and ptp to use the cnf tests image for tests and to be less disruptive with the policies.


fedepaol:
/hold
Another change is required in sriov.

PR #201, opened 2020-05-12
Update the dpdk test to run without testpmd build

SchSeba:
This commit run the deployment base pod with the already exist testpmd inside the dpdk package.

### DIFF
--- a/functests/dpdk/dpdk.go
+++ b/functests/dpdk/dpdk.go
@@ -90,6 +90,7 @@ var _ = Describe("dpdk", func() {
 
 		findOrOverridePerformanceProfile()
 		findOrOverrideSriovNetwork()
+
 		dpdkWorkloadPod = createPod()
 	})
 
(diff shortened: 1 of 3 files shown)


SchSeba:
/hold 

We need to wait for this PR to be merge and then vendor the new sriov-network-operator commit
https://github.com/openshift/sriov-network-operator/pull/209/files

functests/dpdk/dpdk.go
```diff
@@ -355,6 +357,19 @@ func findOrOverrideSriovNetwork() {
...
+
+	// When the dpdk-testing namespace is created it takes time for the network attachment definition to be created
+	// there by the sriov network operator
+	Eventually(func() bool {
+		netattachdef := &sriovk8sv1.NetworkAttachmentDefinition{}
+		err := client.Client.Get(context.TODO(), goclient.ObjectKey{Name: "dpdk-network", Namespace: namespaces.DpdkTest}, netattachdef)
+		if err != nil {
+			return false
+		}
+
```
fedepaol, line 369: nit: can you remove the space?

fedepaol: Also, you can directly return err and assert it not to be nil (I won't hold on any of these comments)

SchSeba:
/hold

Need to fix an issue after update to the new sriov vendor

functests/dpdk/dpdk.go
```diff
@@ -78,7 +78,7 @@ func init() {
 	})
 }
 
-var _ = Describe("dpdk", func() {
+var _ = FDescribe("dpdk", func() {
```
fedepaol, line 81: Left the focus

PR #206, opened 2020-05-14
Pin the latest version of the operator, pin to particular versions of sriov and ptp tests

fedepaol:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/mailru/easyjson v0.7.1 // indirect
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
-	github.com/openshift-kni/performance-addon-operators v0.0.0-20200512233718-b069f90f7330
+	github.com/openshift-kni/performance-addon-operators v0.0.0-20200518161626-01414df7afae
 	github.com/openshift/api v3.9.1-0.20191213091414-3fbf6bcf78e8+incompatible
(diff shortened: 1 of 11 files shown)


go.mod
```diff
@@ -83,3 +83,10 @@ replace (
 	github.com/openshift/machine-config-operator => github.com/openshift/machine-config-operator v0.0.0-20200123151440-ca3e3e1921f3 // release-4.4
 	golang.org/x/tools => golang.org/x/tools v0.0.0-20191206213732-070c9d21b343
 )
+
+// Test deps
+replace (
+	github.com/openshift-kni/performance-addon-operators => github.com/openshift-kni/performance-addon-operators v0.0.0-20200514122016-65bb21a3eb94 // master
```
yanirq, line 89: @fedepaol  why pin against master and not against release-4.5 ?  (https://github.com/openshift-kni/performance-addon-operators/commits/release-4.5)

fedepaol: It's because ptp / sriov are not cutting the 4.6 branch. This master should point to 4.6. I left it explicit so it's easier to notice.

yanirq: @fedepaol  just FYI, currently I see 3 failures around performance addons that related to initramfs image but it was removed completely in https://github.com/openshift-kni/performance-addon-operators/pull/190/  (which is a newer commit in performance addons)

fedepaol: Ok, will update the pr

PR #212, opened 2020-05-15
Make it possible to override the registry / the images used for pods when running make functests.

fedepaol:


hack/run-functests.sh
```diff
@@ -32,7 +36,7 @@ mkdir -p "$TESTS_REPORTS_PATH"
 if [ "$CNF_TESTS_IMAGE" != "" ]; then
```
yuvalk, line 36: CNF_TESTS_IMAGE can be derived from TEST_POD_IMAGES_REGISTRY/TEST_POD_CNF_TEST_IMAGE
same goes for DPDK_TESTS_IMAGE

fedepaol: I don't get the `DPDK_TESTS_IMAGE` part

fedepaol: Also, CNF_TESTS_IMAGE is used as driver to choose the containerized version, whereas overriding the other two is optional. I can remove that variable, but I would need to reintroduce a flag to drive the tests to run in containerized mode, not sure the change would add value.

fedepaol: @yuvalk please take a look.
I had to introduce another variable anyway (TESTS_IN_CONTAINER) but we don't repeat the cnf_tests_image anymore and it's probably more clear than relying on the presence of the image variable to trigger the logic.

hack/run-functests.sh
```diff
@@ -17,6 +17,14 @@ export TESTS_REPORTS_PATH="${TESTS_REPORTS_PATH:-/tmp/artifacts/}"
 export failed=false
 export failures=()
 
+#env variables needed for the containerized version
+export TEST_POD_IMAGES_REGISTRY="${TEST_POD_IMAGES_REGISTRY:-quay.io/openshift-kni/}"
+export TEST_POD_CNF_TEST_IMAGE="${TEST_POD_CNF_TEST_IMAGE:-cnf-tests:4.5}"
+export TEST_POD_DPDK_TEST_IMAGE="${TEST_POD_DPDK_TEST_IMAGE:-dpdk:4.5}"
+
+export CNF_TESTS_IMAGE=$TEST_POD_IMAGES_REGISTRY$TEST_POD_CNF_TEST_IMAGE
```
yuvalk, line 25: this is disabling any (easy) way of running the tests locally
I guess we'd better introduce yet another env like RUN_LOCALLY
and replace the if in L#32

yuvalk: oh sorry, I see you already did that with $TESTS_IN_CONTAINER, ok

fedepaol: I already introduced a new variable which is `TESTS_IN_CONTAINER`
If you look at line 40 below, it does what you are saying but with a different default

PR #213, opened 2020-05-17
Create the dpdk namespace in the deploy overlay

SchSeba:


SchSeba:
@yuvalk PTAL

feature-configs/deploy/dpdk/build-config.yaml
```diff
@@ -1,4 +1,9 @@
 ---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: dpdk
+---
```
yuvalk, line 6: why not have it in a file of it's own?
it's harder to search for it that way
also impossible to remove just part, without editing the file

SchSeba: done

PR #214, opened 2020-05-17
fix feature-wait.sh to use is_ready from deploy

yuvalk:
there is no longer 'base' directory

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+pushd .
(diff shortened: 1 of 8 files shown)


hack/common.sh
```diff
@@ -0,0 +1,13 @@
...
+
+pushd .
+cd "$(dirname "$0")/.."
+
+function finish {
+    popd
+}
+trap finish EXIT
+
+
```
fedepaol, line 13: small nit: extra line

PR #217, opened 2020-05-19
Fix ptp and move the prometheus test

SchSeba:


fedepaol:
@SchSeba while we are touching this part, can you fix the cleanup logic to clean only the config created by this test (same as we did in https://github.com/openshift/ptp-operator/pull/56 ) ?

PR #221, opened 2020-05-20
fix: two steps on cnf-test image disconnected

tripledes:
* Remove blank on variable assignment
* Remove extra double quotes on $TOKEN as its value already contains them and avoids errors from parsing dockerauth.json


fedepaol:
Thanks for fixing the doc!

/ok-to-test

PR #222, opened 2020-05-21
Compress the report output when running make functests

fedepaol:
This should save some space on CI

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -39,6 +39,15 @@ else
   EXEC_TESTS="cnf-tests/test-run.sh -ginkgo.focus=$FOCUS -junit $TESTS_REPORTS_PATH -report $TESTS_REPORTS_PATH"
 fi
 
+reports="cnftests_failure_report.log setup_failure_report.log validation_failure_report.log"
+
+for report in $reports; do
+  if [[ -f "$TESTS_REPORTS_PATH/$report" ]]; then    
+    gzip "$TESTS_REPORTS_PATH/$report"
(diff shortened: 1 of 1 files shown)


hack/run-functests.sh
```diff
@@ -39,6 +39,15 @@ else
   EXEC_TESTS="cnf-tests/test-run.sh -ginkgo.focus=$FOCUS -junit $TESTS_REPORTS_PATH -report $TESTS_REPORTS_PATH"
 fi
 
+reports="cnftests_failure_report.log setup_failure_report.log validation_failure_report.log"
+
+for report in $reports; do
+  if [[ -f "$TESTS_REPORTS_PATH/$report" ]]; then    
+    gzip "$TESTS_REPORTS_PATH/$report"
```
SchSeba, line 46: do we have gzip inside the cnf container?

hack/run-functests.sh
```diff
@@ -39,6 +39,15 @@ else
...
+    mv "$TESTS_REPORTS_PATH/$report".gz "$TESTS_REPORTS_PATH/$report.""$(date +"%Y-%m-%d_%T")".gz
```
SchSeba, line 47: do we also want to remove the files?

hack/run-functests.sh
fedepaol: That is happening outside the container (the user's terminal OR the pod that is running CI tests).
On CI side, will find out :P

hack/run-functests.sh
fedepaol: gzip already removes the original file

hack/run-functests.sh
fedepaol: Just ssh-ed into the test runner pod, it has it onboard

PR #231, opened 2020-05-25
will not search for go in run-functests

yuvalk:
tests can run from container image, in which case, no need for go.
build-test-bin alread take care of making sure go is available for builds.

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -2,14 +2,6 @@
 
 . $(dirname "$0")/common.sh
 
-which go
-if [ $? -ne 0 ]; then
-  echo "No go command available"
-  exit 1
-fi
-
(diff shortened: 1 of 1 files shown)


yuvalk:
/assign @fedepaol 

I forgot to push that in the previous PR :-(

fedepaol:
This should go under `TESTS_IN_CONTAINER`.
If you run the tests with the default (building the bins) it's still required.

yuvalk:
> This should go under `TESTS_IN_CONTAINER`.
> If you run the tests with the default (building the bins) it's still required.

it is already included in hack/build-test-bin.sh
and it is not needed, and even interrupt when running tests from image (bc it pre-requisit golang, unneededly)

PR #233, opened 2020-05-25
Disable ptp test as measuring the latency with master + slave on nodes is not meaningful.

fedepaol:


yuvalk:
can you try and say something more meaningful in the commit msg? (pun intended) 

````
Disable PTP Latency Tests

Can't test for >100us without external HW
```
or something

fedepaol:
/override ci/prow/e2e-gcp 
Last fail is a flake, changed only the commit message.
Those tests are not impacted by the change either.

PR #241, opened 2020-06-03
Clean up sctp leftovers.

fedepaol:
Change the logic on how items to delete are identified. We add a common prefix and puntually delete only sctp related items.

### DIFF
--- a/functests/utils/namespaces/namespaces.go
+++ b/functests/utils/namespaces/namespaces.go
@@ -2,6 +2,7 @@ package namespaces
 
 import (
 	"os"
+	"strings"
 	"time"
 
 	k8sv1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


functests/utils/namespaces/namespaces.go
```diff
@@ -50,53 +51,59 @@ func Create(namespace string, cs *testclient.ClientSet) error {
...
 
-	err = cs.NetworkPolicies(namespace).DeleteCollection(&metav1.DeleteOptions{
-		GracePeriodSeconds: pointer.Int64Ptr(0),
-	}, metav1.ListOptions{})
+	policies, err := cs.NetworkPolicies(namespace).List(metav1.ListOptions{})
 	if err != nil {
 		return err
 	}
+	for _, p := range policies.Items {
+		if strings.HasPrefix(p.Name, "testsctp-") {
```
mmirecki, line 65: "testsctp-" - should we not use the 'prefix' parameter here?

PR #247, opened 2020-06-04
Added a small readme for the demo app

e-minguez:


tools/s2i-dpdk/test/test-app/README.md
```diff
@@ -0,0 +1,123 @@
...
+    TX queue: 0
+      TX desc=512 - TX free threshold=32
+      TX threshold registers: pthresh=32 hthresh=0  wthresh=0
+      TX offloads=0x0 - TX RS bit threshold=32
+testpmd> stop
+Telling cores to stop...
+Waiting for lcores to finish...
+
+  ---------------------- Forward statistics for port 0  ----------------------
+  RX-packets: 0              RX-dropped: 0             RX-total: 0
```
SchSeba, line 100: Can you just add a comment that this number should be > 0 ?

e-minguez: Do you mean all those 0s (including the RX/TX queue)? Do you happen to have a proper output? Thanks!

SchSeba: I don't have one right new sorry

SchSeba: This PR changed a bit the application output if you can run it again and change the output in the README it will be great! 

https://github.com/openshift-kni/cnf-features-deploy/pull/246

PR #257, opened 2020-06-10
[release-4.4] support building the test-pmd application

SchSeba:
DPDK version 1.18

/hold
I want to create a CI job to mirror this image

### DIFF
--- a/feature-configs/demo/dpdk/dpdk-network.yaml
+++ b/feature-configs/demo/dpdk/dpdk-network.yaml
@@ -8,11 +8,6 @@ spec:
     {
       "type": "dhcp"
     }
-  networkNamespace: dpdk-testing
+  networkNamespace: dpdk
   resourceName: dpdknic
   vlan: 0
----
-apiVersion: v1
-kind: Namespace
-metadata:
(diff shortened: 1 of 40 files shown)


SchSeba:
The release PR https://github.com/openshift/release/pull/9624

feature-configs/demo/dpdk/dpdk-network.yaml
```diff
@@ -8,11 +8,11 @@ spec:
...
-  networkNamespace: dpdk-testing
+  networkNamespace: dpdk
   resourceName: dpdknic
   vlan: 0
 ---
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: dpdk-testing
+  name: dpdk
```
yuvalk, line 18: you are creating the namespace object in two different places (it also exist in build-config.yaml)
better to separate to it's own file, just as we've done in release-4.5 and master

SchSeba: In the build config is an image stream I didn't find a namespace creation there.

But you are right I move it to a separate yaml file

PR #258, opened 2020-06-11
Dpdk tests: impact only nodes selected by the optional node selector.

fedepaol:
By setting the NODES_SELECTOR variable, we limit the cluster nodes that are impacted by the test.
This includes sriov configurations, choosing the nodes to run the pods on.
If not set, the tests continue to behave as before.

### DIFF
--- a/functests/dpdk/dpdk.go
+++ b/functests/dpdk/dpdk.go
@@ -36,6 +36,7 @@ import (
 	"github.com/openshift-kni/cnf-features-deploy/functests/utils/images"
 	"github.com/openshift-kni/cnf-features-deploy/functests/utils/machineconfigpool"
 	"github.com/openshift-kni/cnf-features-deploy/functests/utils/namespaces"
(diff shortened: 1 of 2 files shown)


fedepaol:
/hold
Still need validation

functests/dpdk/dpdk.go
```diff
@@ -337,12 +340,16 @@ func findOrOverrideSriovNetwork() {
 		Expect(err).ToNot(HaveOccurred())
 
 		Expect(sriovInfos).ToNot(BeNil())
-		Expect(len(sriovInfos.Nodes)).To(BeNumerically(">", 0))
 
-		sriovDevice, err := sriovInfos.FindOneSriovDevice(sriovInfos.Nodes[0])
+		nn, err := nodes.MatchingOptionalSelectorByName(sriovInfos.Nodes)
+		Expect(sriovInfos).ToNot(HaveOccurred())
```
SchSeba, line 345: this should be `err`

PR #259, opened 2020-06-11
PTP Tests: limit the nodes impacted by the tests using the optional selector

fedepaol:
By setting NODES_SELECTOR, it's possible to limit the nodes impacted by the ptp tests.
In that case, the configuration is applied to a subset of nodes only.

### DIFF
--- a/functests/ptp/ptp.go
+++ b/functests/ptp/ptp.go
@@ -36,7 +36,8 @@ var _ = Describe("ptp", func() {
 	execute.BeforeAll(func() {
 		if !discovery.Enabled() {
 			Clean()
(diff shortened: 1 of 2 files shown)


functests/ptp/ptp.go
```diff
@@ -36,7 +36,13 @@ var _ = Describe("ptp", func() {
 	execute.BeforeAll(func() {
 		if !discovery.Enabled() {
 			Clean()
-			ptpNodes, err := nodes.GetNodeTopology(client.Client)
+
+			listOptions := metav1.ListOptions{}
```
mmirecki, line 40: Do we need this?

PR #264, opened 2020-06-18
Pin to ptp / sriov / performance operator to 4.6 (master) branch

fedepaol:
Now that sriov / ptp are branched out, and that this branch is for 4.6, we pin the test dependencies to 4.6 too.


cynepco3hahue:
```
      no matches for kind "SriovNetworkNodePolicy" in version "sriovnetwork.openshift.io/v1"
  occurred
```

Looks like some change under the SR-IOV operator.

PR #266, opened 2020-06-19
Delegate the cleaning logic to each suite's clean.All method

fedepaol:
On the long term, each test suite should be responsible for cleaning its own assets.
Here we:
- add a clean.All() method for local tests
- consume and use each suite's clean.All()
- disable the cleaning of the perf profile in CI


SchSeba:
/lgtm

/hold

please remove the hold after validation

fedepaol:
/hold cancel
Validated, no leftovers around

PR #269, opened 2020-06-25
Add dpdk s2i build script for dpdk deployment

sabinaaledort:
If the registry is not ready when downloading the dpdk image the
s2i build running only once and the whole CNF deployment will wait for a timeout.

Retry and build dpdk image until compelte, with a timeout.

This script is not in the Makefile beacuse it doesnt follow the blueprint (kustomize) approach.

Signed-off-by: Sabina Aledort <saledort@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ All kustomize configs should be entirely declarative in nature. This means no ba
 
 - You need a running OCP 4.4 cluster and a valid KUBECONFIG.
 - You need at least one node with the `node-role.kubernetes.io/worker-cnf=""` label and a `MachineConfigPool` matching `worker-cnf` machine configurations
+- You need to install `jq` (a command line tool for parsing JSON) on the local machine.
 
 You can run `make setup-test-cluster` to have the first two (or the first in case of only one) workers labeled as `worker-cnf` and to have the `MachineConfigPool` created.
 
--- a/feature-configs/deploy/dpdk/post_deploy.sh
+++ b/feature-configs/deploy/dpdk/post_deploy.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+start_build=false
+last_build=$(oc get build -n dpdk -o json | jq '.items[-1].metadata.name' | tr -d '"')
+if [ $last_build == "null" ]; then
+    start_build=true
+else
+    build_status=$(oc get build -n dpdk $last_build -o json | jq '.status.phase' | tr -d '"')
+    if [ $build_status == "Complete" ]; then
+        exit 0
+    elif [ $build_status == "Running" ]; then
+        exit 1
+    elif [ $build_status == "Failed" ]; then
+        start_build=true
+    fi
+fi
+
+if $start_build; then
+    oc start-build -n dpdk s2i-dpdk
+    exit 1
+fi
+
--- a/hack/feature-deploy.sh
+++ b/hack/feature-deploy.sh
@@ -47,7 +47,21 @@ do
     if [[ $? != 0 ]]; then
       echo "[WARN] Deployment of feature '$feature' failed."
       feature_failed=1
+    else
+      deploy_complete=feature-configs/${FEATURES_ENVIRONMENT}/${feature}/post_deploy.sh
+      if [[ ! -f $deploy_complete ]]; then
+        deploy_complete=feature-configs/deploy/${feature}/post_deploy.sh
+        if [[ ! -f $deploy_complete ]]; then
+          continue
+        fi
+      fi
+
+      if ! $deploy_complete; then
+        echo "[WARN] Deployment of feature '$feature' failed."
+        feature_failed=1
+      fi
     fi
+
     set -e
 
   done


hack/build-s2i-dpdk.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
```
yuvalk, line 2: source .common.sh

hack/build-s2i-dpdk.sh
```diff
@@ -0,0 +1,34 @@
...
+ELAPSED=0
+TIMEOUT=180
+ATTEMPTS=0
+MAX_ATTEMPTS=10
+build_complete=false
+
```
yuvalk, line 8: use FEATURES_ENVIRONMENT var to actually deploy dpdk first (I think it's better than documenting it is needed to be deployed manually)

hack/build-s2i-dpdk.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+ELAPSED=0
+TIMEOUT=180
+ATTEMPTS=0
+MAX_ATTEMPTS=10
```
yuvalk, line 6: you can use a single ATTEMPTS var if you start from 10 and decrease on each try until zero.
I think it's more readable.
ie in the until do
[ $ATTEPTS -eq "0" ]
and later below do
{{ ATTEMPTS-- }}

actually same can go for timeout

hack/build-s2i-dpdk.sh
```diff
@@ -0,0 +1,34 @@
...
+        elif [ $build_status == "Failed" ]; then
+            start_build=true
+        fi
+    fi
+    if $start_build; then
+        oc start-build -n dpdk s2i-dpdk
+        (( ATTEMPTS++ ))
+    fi
+    sleep 10
+    (( ELAPSED++ ))
```
yuvalk, line 28: sleep 10
then add just 1 to ELAPSED?
so timeout 180 actually means 1800 seconds sound like wayy too much

I think 30 minutes overall should be more than enough for test_pmd on any HW

sabinaaledort: 1800 seconds = 30 minutes

hack/build-s2i-dpdk.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
+set +e
+
+if [ "$FEATURE_DIR" == "" ]; then
```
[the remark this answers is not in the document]
sabinaaledort: yes, this is the path to the directory where the dpdk configuration is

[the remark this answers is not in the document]
SchSeba: why we need this?

we can use the `FEATURES_ENVIRONMENT` + "dpdk"
but you also need to check the list of features under `FEATURES` to see if we want to deploy the dpdk or not

sabinaaledort:
@SchSeba do you have an idea how to add it as part of feature-deploy.sh and keep the blueprint pattern? this script is not declarative.

SchSeba:
@sabinaaledort I know but maybe you can add here a check if the feature is dpdk and run this script after the `oc apply -k`

https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/feature-deploy.sh#L51

@yuvalk @fedepaol WDYT?

fedepaol:
Sorry I am late to the party.
I would try to keep the same modular approach we have, isolating the deploy (which can be done via kustomize), the build (which probably needs an additional entry point) and the wait (which we already support via `is_ready.sh`)
So, my proposal is to split what is being done here in 3:

The "kustomize part" relies on the existing `feature_deploy.sh`
We add an additional (optional) post deploy phase where you start the build (with an optional entry point file, more or less the same way we do for `is_ready.sh` in `feature_wait.sh`.

Does it make sense? Is it over-engineered? I am not completely fond of the idea of having an "alien" script that do not fit with the rest.

SchSeba:
>The "kustomize part" relies on the existing feature_deploy.sh
We add an additional (optional) post deploy phase where you start the build (with an optional entry point file, more or less the same way we do for is_ready.sh in feature_wait.sh.

I like the idea of adding a post deploy phase and every feature can decide if he want it or not

yuvalk:
> Sorry I am late to the party.
> I would try to keep the same modular approach we have, isolating the deploy (which can be done via kustomize), the build (which probably needs an additional entry point) and the wait (which we already support via `is_ready.sh`)
> So, my proposal is to split what is being done here in 3:
> 
> The "kustomize part" relies on the existing `feature_deploy.sh`
> We add an additional (optional) post deploy phase where you start the build (with an optional entry point file, more or less the same way we do for `is_ready.sh` in `feature_wait.sh`.
> 
> Does it make sense? Is it over-engineered? I am not completely fond of the idea of having an "alien" script that do not fit with the rest.

No I don't think this is over engineering, I agree it's the right approach :-) 
I even told @sabinaaledort beforehand we might have to do that eventually

so we'll implement a is_deploy.sh mechanism (similarly for as is_ready is for wait)
one question remain open though is how will we re-trigger the deployment.
for s2i (which btw, dont you think it should be a feature on it's own, not part of dpdk? ie - we'll have both dpdk and s2i-dpdk features, but that unrelated to this pr), reapplying the kustomize wont retrigger the build (only that oc build start command will)
sticking it into is_deploy.sh is both ugly and does not follow the BPP (Blueprint pattern ;-) ).
so maybe also a post_deploy.sh mechanism too?

which BTW (sorry for yet another detour from this PR): I think should be fixed there, oc s2i buld process should be declerative, build operator should keep reconciling until getting the correct result)
so with that in mind, this post_deploy for dpdk is hopefully only 'tempurary'

fedepaol:
> > Sorry I am late to the party.
> > I would try to keep the same modular approach we have, isolating the deploy (which can be done via kustomize), the build (which probably needs an additional entry point) and the wait (which we already support via `is_ready.sh`)
> > So, my proposal is to split what is being done here in 3:
> > The "kustomize part" relies on the existing `feature_deploy.sh`
> > We add an additional (optional) post deploy phase where you start the build (with an optional entry point file, more or less the same way we do for `is_ready.sh` in `feature_wait.sh`.
> > Does it make sense? Is it over-engineered? I am not completely fond of the idea of having an "alien" script that do not fit with the rest.
> 
> No I don't think this is over engineering, I agree it's the right approach :-)
> I even told @sabinaaledort beforehand we might have to do that eventually
> 
> so we'll implement a is_deploy.sh mechanism (similarly for as is_ready is for wait)

`post_deploy.sh` or `custom_deploy.sh` sounds better 

> one question remain open though is how will we re-trigger the deployment.
> for s2i (which btw, dont you think it should be a feature on it's own, not part of dpdk? ie - we'll have both dpdk and s2i-dpdk features, but that unrelated to this pr), reapplying the kustomize wont retrigger the build (only that oc build start command will)
> sticking it into is_deploy.sh is both ugly and does not follow the BPP (Blueprint pattern ;-) ).
> so maybe also a post_deploy.sh mechanism too?

Well, this depends on how you implement "post_deploy.sh" which is meant to be "post kustomize deploy". Reapplying the kustomize won't trigger the build, but the entry point here is make `features-deploy` . You already have the idempotency logic in the script I think.


> 
> which BTW (sorry for yet another detour from this PR): I think should be fixed there, oc s2i buld process should be declerative, build operator should keep reconciling until getting the correct result)
> so with that in mind, this post_deploy for dpdk is hopefully only 'tempurary'

README.md
```diff
@@ -12,6 +12,7 @@ All kustomize configs should be entirely declarative in nature. This means no ba
 
 - You need a running OCP 4.4 cluster and a valid KUBECONFIG.
 - You need at least one node with the `node-role.kubernetes.io/worker-cnf=""` label and a `MachineConfigPool` matching `worker-cnf` machine configurations
+- You need to install `jq` (a command line tool for parsing JSON) in the cluster.
```
SchSeba, line 15: The command doesn't need to be inside in the cluster. it needs to be on local machine where you run `make deploy`

SchSeba:
@fedepaol @yuvalk from my side this is ready

PR #270, opened 2020-06-29
Fix error copying s2i/bin folder

alosadagrande:
When building the dpdk image found that s2i/bin folder could not be found. Just adapt the Dockerfile to the correct path. Looks like the current pwd matches the root of the cloned repository not the contextDir.

This fix solves the issue when building the dpdk base from OpenShift using the [build-config.yaml](https://github.com/openshift-kni/cnf-features-deploy/blob/master/tools/s2i-dpdk/base-image/build-config.yaml) object 


alosadagrande:
Closed. I'll open a new PR that replaces centos7 base image with by centos8

PR #272, opened 2020-07-03
Replaced centos7 base image byt centos8 in s2i-dpdk build

alosadagrande:
Replaced openshift/centos7 with centos/centos8 which requires installing PowerTool repo where libpcap-devel is placed in new CentOS release.

Also, modified the S2i scripts COPY PATH since I was facing "not found" issues during the building 
process.

CC/ @SchSeba since he is the maintainer of the built/Dockerfile.

### DIFF
--- a/tools/s2i-dpdk/Dockerfile
+++ b/tools/s2i-dpdk/Dockerfile
@@ -1,5 +1,5 @@
 # dpdk-centos7
-FROM registry.centos.org/centos/centos:8
+FROM docker.io/centos:centos8
 #FROM centos:7
 
 LABEL maintainer="Sebastian Scheinkman <sebassch@gmail.com>"
--- a/tools/s2i-dpdk/Dockerfile
+++ b/tools/s2i-dpdk/Dockerfile
@@ -1,5 +1,5 @@
 # dpdk-centos7
-FROM openshift/base-centos7
+FROM registry.centos.org/centos/centos:8
 #FROM centos:7
 
 LABEL maintainer="Sebastian Scheinkman <sebassch@gmail.com>"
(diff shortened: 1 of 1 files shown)


tools/s2i-dpdk/Dockerfile
```diff
@@ -72,7 +74,7 @@ RUN chmod -R 777 /opt/app-root
 
 # TODO: Copy the S2I scripts to /usr/libexec/s2i, since openshift/base-centos7 image
 # sets io.openshift.s2i.scripts-url label that way, or update that label
-COPY ./s2i/bin/ /usr/libexec/s2i
+COPY ./tools/s2i-dpdk/s2i/bin/ /usr/libexec/s2i
```
SchSeba, line 77: This is not needed we are running the build from `/tools/s2i-dpdk/` directory

tools/s2i-dpdk/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # dpdk-centos7
-FROM openshift/base-centos7
+FROM docker.io/centos:centos8
```
SchSeba, line 2: I think this will not work because we are using the openshift-CI to build the image maybe we need to change the image here https://github.com/openshift/release/blob/93e244edbfa8c8938cda6b20779e46dbd3f25ce5/ci-operator/config/openshift-kni/cnf-features-deploy/openshift-kni-cnf-features-deploy-master.yaml#L2

SchSeba:
@alosadagrande ping?

Can you fix my comment? I will like to merge this PR

SchSeba:
closing this one
/close

PR #273, opened 2020-07-03
Add discovery of performance profile in dpdk tests

mmirecki:
This PR adds discovery of performance profile in dpdk tests.
If discovery mode is enabled, the tests will attempt to reuse an existing performance profile.
Following PR's will add discovery of sriov policy.

### DIFF
--- a/functests/dpdk/dpdk.go
+++ b/functests/dpdk/dpdk.go
@@ -32,6 +32,7 @@ import (
 	sriovnetwork "github.com/openshift/sriov-network-operator/test/util/network"
 
 	"github.com/openshift-kni/cnf-features-deploy/functests/utils/client"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/discovery"
 	"github.com/openshift-kni/cnf-features-deploy/functests/utils/execute"
 	"github.com/openshift-kni/cnf-features-deploy/functests/utils/images"
 	"github.com/openshift-kni/cnf-features-deploy/functests/utils/machineconfigpool"
@@ -48,8 +49,9 @@ const (
 )
 
 var (
-	machineConfigPoolName  string
-	performanceProfileName string
+	machineConfigPoolName          string
+	performanceProfileName         string
+	enforcedPerformanceProfileName string
 
 	sriovclient *sriovtestclient.ClientSet
 
@@ -68,6 +70,8 @@ func init() {
 	performanceProfileName = os.Getenv("PERF_TEST_PROFILE")
 	if performanceProfileName == "" {
 		performanceProfileName = "performance"
+	} else {
+		enforcedPerformanceProfileName = performanceProfileName
 	}
 
 	OriginalSriovPolicies = make([]*sriovv1.SriovNetworkNodePolicy, 0)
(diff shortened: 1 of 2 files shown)


functests/dpdk/dpdk.go
```diff
@@ -299,10 +319,54 @@ func tryToFindDPDKPod() (*corev1.Pod, bool) {
...
+		}
+		return true, ""
+	}
 
+	performanceProfileList := &perfv1alpha1.PerformanceProfileList{}
+	err := client.Client.List(context.TODO(), performanceProfileList)
+	if err != nil {
+		return false, fmt.Sprintf(reason, err)
+	}
+	for _, performanceProfile := range performanceProfileList.Items {
```
fedepaol, line 342: You may need to return this. The ultimate goal is making sure that we are running a pod where it can run.
To drive it, we have three things to influence it:
- the resources we ask (that may fit the perf profile)
- the network resource
- the node selector
I would mix all of them while doing the discovery, to make sure the dpdk pod is scheduled on the right node.

If we handle them separately, there can be a valid performance profile impacting a set of nodes which is different from the set of nodes containing a valid sriov device.

functests/dpdk/dpdk.go
```diff
@@ -299,10 +319,54 @@ func tryToFindDPDKPod() (*corev1.Pod, bool) {
 	return nil, false
 }
 
-func findOrOverridePerformanceProfile() {
-	performanceProfile, valid, err := ValidatePerformanceProfile()
-	Expect(err).ToNot(HaveOccurred())
+func discoverPerformanceProfile() (bool, string) {
+	reason := "Can not run tests in discovery mode. Failed to find a valid perfomance profile. %s"
```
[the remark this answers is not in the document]
mmirecki: Done

functests/dpdk/dpdk.go
```diff
@@ -575,7 +629,7 @@ func CreateSriovNetwork(sriovDevice *sriovv1.InterfaceExt, sriovNetworkName stri
 
 }
 
-func CreateDPDKWorkload() (*corev1.Pod, error) {
+func CreateDPDKWorkload(profile *perfv1alpha1.PerformanceProfile) (*corev1.Pod, error) {
```
fedepaol, line 632: I would pass directly the node selector to the pod, instead of the perf profile.
This should also take into accont an eventual NODE_SELECTOR specified from outside.

mmirecki: > I would pass directly the node selector to the pod, instead of the perf profile.
Done
> This should also take into accont an eventual NODE_SELECTOR specified from outside.
Not part of this PR. Please create a story, we should add this accross all our tests.

functests/dpdk/dpdk.go
```diff
@@ -666,9 +737,14 @@ sleep INF
...
-	selector, ok := nodes.PodLabelSelector()
-	if ok {
-		dpdkPod.Spec.NodeSelector = selector
+	if len(nodeSelector) > 0 {
+		dpdkPod.Spec.NodeSelector = nodeSelector
+	}
+
+	if nodeSelector != nil && len(nodeSelector) > 0 {
+		for k, v := range nodeSelector {
+			dpdkPod.Spec.NodeSelector[k] = v
```
fedepaol, line 746: I think this is gonna panic if NodeSelector is not initialized.

fedepaol:
Looks good, one comment about the map.

PR #275, opened 2020-07-07
Add typical baremetal config

sabinaaledort:
Signed-off-by: Sabina Aledort <saledort@redhat.com>

### DIFF
--- a/feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
+++ b/feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
@@ -1,0 +1,13 @@
+apiVersion: performance.openshift.io/v1alpha1
+kind: PerformanceProfile
+metadata:
+  name: performance
+spec:
+  cpu:
+    isolated: 8-15
+    reserved: 0-1
+  hugepages:
+    pages:
+    - size: "1G"
+      count: 8
+      node: 0
--- a/feature-configs/typical-baremetal/dpdk/kustomization.yaml
+++ b/feature-configs/typical-baremetal/dpdk/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+bases:
+- ../../deploy/dpdk
+kind: Kustomization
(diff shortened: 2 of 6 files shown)


yuvalk:
@sabinaaledort can we add on record why it include 'empty' kustomizations for dpdk, ptp, sctp and sriov?
ie why doesnt it include just the performance profile?

feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: performance.openshift.io/v1alpha1
+kind: PerformanceProfile
+metadata:
+  name: performance
+spec:
+  cpu:
+    isolated: 8-15
+    reserved: 0-7
```
fromanirh, line 8: this looks suspicious/wrong. Could you elaborate why you want this setting?

yuvalk: this is also what we're currently using d/s, so if it's wrong we better fix it.

@fromanirh what would you suggest we typically use?

fromanirh: Is not "wrong" in the sense that it won't work. It will work, but sounds very wasteful, this means only half of the cores will be used for workloads, and the other half for the system (kernel, kubelet, systemd...)
I'm not aware of any good guideline yet. Considering that system task is pretty much overhead, my rule of thumb would be start with no more than 1/4 cores reserved for system, likely even less (1/8?)
Perhaps @MarSik or @cynepco3hahue have more accurate advice (or authoritative docs to follow) here.

yuvalk: I think it should not be counted in %% of CPUs, because for instance, if we have 80 cores, and not so much overhead, maybe even 2 reserved is enough.
this (typicla-baremetal) will be used mainly for testing, so maybe a better approach would be:
```
reserved: 0,1
isolated: 8-15
```
which give a good mix of isolated/non-isolated, mixing numa nodes and reserving more than one core..

MarSik: Yep, that looks much more reasonable.

PR #276, opened 2020-07-10
Discovery mode for sriov network policies

mmirecki:


functests/dpdk/dpdk.go
```diff
@@ -997,14 +1019,51 @@ func waitForSRIOVStable() {
 	}, waitingTime, 1*time.Second).Should(BeTrue())
 }
 
-func filterPerformanceProfilesWithAvailableNodes(performanceProfiles []*perfv1alpha1.PerformanceProfile, nodeSelector map[string]string) []*perfv1alpha1.PerformanceProfile {
-	availableProfiles := make([]*perfv1alpha1.PerformanceProfile, 0)
+// findPerformanceProfileAndPolicyWithAvailableNodes finds a profile/sriovPolicy match for which a node with
+// allocatable resources is available. It will return a profile/sriovPolicy for a policy with resource name
+// "dpdknic", or a pair with the most available resource on node
+func findPerformanceProfileAndPolicyWithAvailableNodes(performanceProfiles []*perfv1alpha1.PerformanceProfile, nodeSelector map[string]string,
```
fedepaol, line 1025: can we return a custom struct?

functests/utils/nodes/nodes.go
```diff
@@ -278,12 +278,15 @@ func MatchingOptionalSelectorPTP(toFilter []ptpv1.NodePtpDevice) ([]ptpv1.NodePt
 	return res, nil
 }
 
-// ValidateNodeIsAvailableForSelector check if any nodes are available for given nodeSelector
-func ValidateNodeIsAvailableForSelector(nodeSelector map[string]string) bool {
+// NodesAvailableForSelector returns nodes available for given nodeSelector
```
fedepaol, line 281: The package name is already "nodes", you can just call this "AvailableForSelector"

functests/dpdk/dpdk.go
```diff
@@ -997,14 +1019,51 @@ func waitForSRIOVStable() {
...
+) (preferredProfile *perfv1alpha1.PerformanceProfile, preferredResource string, preferredDevice *sriovv1.InterfaceExt, err error) {
+	currentResourceCount := 0
+	sriovInfos, err := sriovcluster.DiscoverSriov(sriovclient, SRIOV_OPERATOR_NAMESPACE)
+	Expect(err).ToNot(HaveOccurred())
```
fedepaol, line 1029: Either expecting inside the function or returning the error is fine, but I'd try to be consistent in following one single approach.

functests/dpdk/dpdk.go
```diff
@@ -997,14 +1019,51 @@ func waitForSRIOVStable() {
...
+
+	sriovPolicies := &sriovv1.SriovNetworkNodePolicyList{}
+	err = client.Client.List(context.TODO(), sriovPolicies, &goclient.ListOptions{Namespace: SRIOV_OPERATOR_NAMESPACE})
+	Expect(err).ToNot(HaveOccurred())
+
 	for _, profile := range performanceProfiles {
 		profileNodeSelector := nodes.NodesSelectorUnion(nodeSelector, profile.Spec.NodeSelector)
```
fedepaol, line 1036: Didn't notice before, but `NodesSelectorUnion` should be intersection?
I would also remove the initial `Nodes`, see my other comment below.

functests/dpdk/dpdk.go
```diff
@@ -997,14 +1019,51 @@ func waitForSRIOVStable() {
...
+				quantity := node.Status.Allocatable[corev1.ResourceName("openshift.io/"+sriovPolicy.Spec.ResourceName)]
+				resourceCount64, _ := (&quantity).AsInt64()
+				resourceCount := int(resourceCount64)
+				sriovDevice, err := sriovInfos.FindOneSriovDevice(node.ObjectMeta.Name)
+				if err != nil {
+					continue
+				}
+				// Return profile and policy with the prefered resource name if available
+				if sriovPolicy.Spec.ResourceName == resourceName {
+					return profile, sriovPolicy.Spec.ResourceName, sriovDevice, nil
```
fedepaol, line 1057: Can you be consistent with named vs non named returns?
Probably it comes more natural if we return a struct.

functests/dpdk/dpdk.go
```diff
@@ -107,21 +109,29 @@ var _ = Describe("dpdk", func() {
...
+			if err != nil {
+				discoverySuccessful, discoveryFailedReason = false, "Can not run tests in discovery mode. Failed to discover required resources."
 				return
 			}
-			// TODO: use the first matching profile for now. When sriov discovery is added, a profile with
-			// an available sriov node should be chosen
-			nodeSelector = performanceProfiles[0].Spec.NodeSelector
+			nodeSelector = nodes.NodesSelectorUnion(nodeSelector, profile.Spec.NodeSelector)
+
+			validSriovNetwork, err := ValidateSriovNetwork(resourceName)
```
fedepaol, line 121: Why not creating it always as we do in sriov? The get looks for a specific name that won't be there in discovery mode.

functests/dpdk/dpdk.go
mmirecki: Done

functests/utils/nodes/nodes.go
mmirecki: Done

functests/dpdk/dpdk.go
mmirecki: Done

functests/dpdk/dpdk.go
mmirecki: The selector should be an union. We want to have the nodes that are limited by both the provided selector, and the selector of the profile.
The nodes we get as a result will be an intersection.

functests/dpdk/dpdk.go
mmirecki: Done

functests/dpdk/dpdk.go
mmirecki: This is the original code (moved from [here](https://github.com/openshift-kni/cnf-features-deploy/pull/276/files#diff-3330f47928cf477b3ddae2071f543f66L404) ).

fedepaol: I see, nut now that we are more explicit about discovery / not discovery, when running in non discovery mode the network won't be there. I'd make that part simpler and just create the network

functests/dpdk/dpdk.go
fedepaol: Got it, thanks.
Can you rename the func to `SelectiorUnion` as the nodes is already implicit in the package name?

fedepaol:
/lgtm
/approve

Feel free to remove the hold once you validated it

PR #277, opened 2020-07-10
Add an integration test checking SCTP traffic across an SRIOV device.

fedepaol:
We want to ensure that SCTP works also across secondary interfaces.


functests/sctp/sctp_sriov.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/namespaces"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/sriov"
+)
+
+const (
+	sriovOperatorNamespace = "openshift-sriov-network-operator"
+	testNetwork            = "test-sctp-sriov-network"
+)
+
+var _ = Describe("[sriov] SCTP integration", func() {
```
SchSeba, line 34: how we are going to be able to run this only if sriov and sctp exist in the cluster?

functests/sctp/sctp_sriov.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/client"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/discovery"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/execute"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/namespaces"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/sriov"
+)
+
+const (
+	sriovOperatorNamespace = "openshift-sriov-network-operator"
```
SchSeba, line 30: lets allow to configure this variable just the same as we can in the sriov repo

functests/sctp/sctp_sriov.go
```diff
@@ -0,0 +1,202 @@
...
+	var testNode string
+
+	execute.BeforeAll(func() {
+		err := namespaces.Create(TestNamespace, client.Client)
+		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func() error {
+			_, err := client.Client.ServiceAccounts(TestNamespace).Get(context.Background(), "default", metav1.GetOptions{})
+			return err
+		}, 5*time.Minute, 5*time.Second).Should(Not(HaveOccurred()))
```
SchSeba, line 46: I think 5 min here is too long can we change this to 1 min?

functests/sctp/sctp_sriov.go
```diff
@@ -0,0 +1,202 @@
...
+				Expect(err).ToNot(HaveOccurred())
+				resNum, _ := testedNode.Status.Allocatable["openshift.io/sctptestres"]
+				capacity, _ := resNum.AsInt64()
+				return capacity
+			}, 3*time.Minute, time.Second).Should(Equal(int64(5)))
+
+		} else {
+			err := sriovnamespaces.CleanNetworks(sriovOperatorNamespace, sriovclient)
+			Expect(err).ToNot(HaveOccurred())
+			err = sriovnamespaces.CleanPods(sriovOperatorNamespace, sriovclient)
```
SchSeba, line 74: why we remove all the pods in the sriov namespace?

functests/sctp/sctp_sriov.go
```diff
@@ -0,0 +1,202 @@
...
+			func(node string, sriovDeviceList []*sriovv1.InterfaceExt) (*sriovv1.InterfaceExt, bool) {
+				if len(sriovDeviceList) == 0 {
+					return nil, false
+				}
+				return sriovDeviceList[0], true
+			},
+		)
+		Expect(err).ToNot(HaveOccurred())
+
+		discoveryFailed = node == "" || resourceName == "" || numVfs < 2
```
SchSeba, line 94: can we move the `if` to here to check if the discovery failed?

functests/sctp/sctp_sriov.go
```diff
@@ -0,0 +1,202 @@
...
+
+				testClientServerConnection(client.Client, TestNamespace, ips[0],
+					30101, testNode, serverPod.Name, true, testNetwork)
+			})
+		})
+	})
+})
+
+func discoverSRIOVNodes(client *sriovtestclient.ClientSet, sctpSelector string) *sriovcluster.EnabledNodes {
+
```
SchSeba, line 131: nit:remove this empty line

functests/sctp/sctp_sriov.go
fedepaol: So, I was a bit biased here. My idea is that sriov is the hardest to get, so that's why I am putting the explicit selector here.
If we are not focusing on sriov, this test will be skipped.
Finally, one would want to test sriov BUT NOT SCTP and in that case my idea was to document this particolar case.

functests/sctp/sctp_sriov.go
fedepaol: The dpdk one here is following this same approach, I would leave it as it is.
If we ever need to have it here too (read, if we want to test this on kind) we'll introduce it.

functests/sctp/sctp_sriov.go
fedepaol: This is the same as sctp, and the happy path is that it should be immediate, but I can change.

functests/sctp/sctp_sriov.go
fedepaol: I did not mean to, wrong namespace I guess.

functests/sctp/sctp_sriov.go
fedepaol: Nope, because it's `BeforeAll`. Whenever we add a new test, the flag check must happen in BeforeEach, otherwise only the first one will be skipped.

PR #280, opened 2020-07-20
[release 4.5] Bug 1858849: Use right driver for dpdk 4.5

SchSeba:
This commit fix the sriov policy creation on the dpdk test.
We need to use the vendor ID an not the Device ID.


fedepaol:
Even if it's not enforced by prow, I think we need a bz for this @SchSeba

PR #284, opened 2020-07-26
vendor latest sriov-network-operator release 4.6

SchSeba:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -81,5 +81,5 @@ replace (
 replace (
 	github.com/openshift-kni/performance-addon-operators => github.com/openshift-kni/performance-addon-operators v0.0.0-20200708085001-4c605c2d58b5 // master
 	github.com/openshift/ptp-operator => github.com/openshift/ptp-operator v0.0.0-20200626140756-02c7317c2dbd // release-4.6
-	github.com/openshift/sriov-network-operator => github.com/openshift/sriov-network-operator v0.0.0-20200720133025-d0596046f45c // release-4.6
+	github.com/openshift/sriov-network-operator => github.com/openshift/sriov-network-operator v0.0.0-20200722065212-ebf7a24cc54f // release-4.6
 )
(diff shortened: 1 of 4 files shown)


yuvalk:
is there's any way to properly validate these numbers?

SchSeba:
@yuvalk I am not sure I understand the question.

I just change the `go.mod` file to release-4.6 then run `make dep-ensure`

go.mod
```diff
@@ -81,5 +81,5 @@ replace (
 replace (
 	github.com/openshift-kni/performance-addon-operators => github.com/openshift-kni/performance-addon-operators v0.0.0-20200708085001-4c605c2d58b5 // master
 	github.com/openshift/ptp-operator => github.com/openshift/ptp-operator v0.0.0-20200626140756-02c7317c2dbd // release-4.6
-	github.com/openshift/sriov-network-operator => github.com/openshift/sriov-network-operator v0.0.0-20200720133025-d0596046f45c // release-4.6
+	github.com/openshift/sriov-network-operator => github.com/openshift/sriov-network-operator v0.0.0-20200722065212-ebf7a24cc54f
```
yuvalk, line 84: no, see here, you actually removed the 4.6 comment
and changed the version
but is there any proper way to tell this is actually 4.6 sriov?

SchSeba: You can take a look on the hash.
If you go into the https://github.com/openshift/sriov-network-operator/tree/release-4.6 
This is the hash from the last commit on the branch `ebf7a24cc54f429692d5a20e52d9593eb9459ad7` same as the go.mod entry `v0.0.0-20200722065212-ebf7a24cc54f`

yuvalk: and can we keep the comment that this actually means 4.6? maybe even add a link to that commit ?

fedepaol: There's' no proper way as go mod expect the dependencies to respect semver tagging (which we as RH don't).
The commit hash is part of the "pseudoversion", that label that follows the dependency. I don't think a link is needed, as that is enough (and unique) to refer a specific version of the dependency.

Would make sense to restore the comment though.

yuvalk: I meant a link in the comment.

fedepaol: For the comment, I do agree. The link is redundant to me because we already have the commit hash there.

yuvalk: yeah fine, I just didn't realize that this is a commit hash on the other repo.

SchSeba: done :)

SchSeba:
@yuvalk @fedepaol PTAL

PR #288, opened 2020-08-03
Add sleep time before checking last dpdk s2i build

sabinaaledort:
When deploying dpdk it takes few seconds before the build is created.
Therefore, we should wait before checking if a build exist.


feature-configs/deploy/dpdk/post_deploy.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 start_build=false
+sleep 10
```
yuvalk, line 4: space line?

PR #290, opened 2020-08-05
Use default network for dpdk test

SchSeba:
This PR change to dpdk test to create a policy when the test are not in discovery mode to find and use the default router and use that interface.

I make this change because on the default network there is always
traffic for the testpmd application to process, this makes the test stable.

### DIFF
--- a/functests/dpdk/dpdk.go
+++ b/functests/dpdk/dpdk.go
@@ -3,6 +3,7 @@ package dpdk
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/fields"
 	"os"
 	"strconv"
 	"strings"
@@ -414,7 +415,7 @@ func findOrOverrideSriovPolicyAndNetwork() {
 
 		Expect(len(nn)).To(BeNumerically(">", 0))
 
-		sriovDevice, err := sriovInfos.FindOneSriovDevice(nn[0])
+		sriovDevice, err := findDpdkSriovDevice(sriovInfos, nn[0])
 		Expect(err).ToNot(HaveOccurred())
 
 		CreateSriovPolicy(sriovDevice, nn[0], 5, resourceName)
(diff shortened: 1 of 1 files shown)


functests/dpdk/dpdk.go
```diff
@@ -581,6 +581,68 @@ func ValidateSriovPolicy() (bool, error) {
...
+		if pod.Spec.NodeName == nodeName {
+			return pod, nil
+		}
+	}
+
+	return corev1.Pod{}, fmt.Errorf("failed to find pod on node %s", nodeName)
+}
+
+// FindDpdkSriovDevice will search for the default gateway interface to be used as the PF
+func FindDpdkSriovDevice(enabledNodes *sriovcluster.EnabledNodes, nodeName string) (*sriovv1.InterfaceExt, error) {
```
fedepaol, line 595: Does this need to be exported?

functests/dpdk/dpdk.go
```diff
@@ -581,6 +581,68 @@ func ValidateSriovPolicy() (bool, error) {
...
+	s, ok := enabledNodes.States[nodeName]
```
fedepaol, line 596: nodeStatus. You are using s way below, where it's harder to get its context

functests/dpdk/dpdk.go
```diff
@@ -581,6 +581,68 @@ func ValidateSriovPolicy() (bool, error) {
...
+	if !ok {
+		return nil, fmt.Errorf("Node %s not found", nodeName)
+	}
+
+	podList := &corev1.PodList{}
+	err := client.Client.List(context.Background(), podList, &goclient.ListOptions{Namespace: SRIOV_OPERATOR_NAMESPACE,
+		LabelSelector: labels.SelectorFromSet(labels.Set{"app": "sriov-network-config-daemon"})})
```
fedepaol, line 603: Can't you just use FieldsSelector here and avoid the findPodBy..?

functests/dpdk/dpdk.go
```diff
@@ -414,7 +414,7 @@ func findOrOverrideSriovPolicyAndNetwork() {
 
 		Expect(len(nn)).To(BeNumerically(">", 0))
 
-		sriovDevice, err := sriovInfos.FindOneSriovDevice(nn[0])
+		sriovDevice, err := FindDpdkSriovDevice(sriovInfos, nn[0])
```
fedepaol, line 417: Do we need the same change in discovery mode? See https://github.com/openshift-kni/cnf-features-deploy/blob/4c169cd5bdce2a098c0923934917395ca4fe2985/functests/utils/discovery/discovery.go#L69

functests/dpdk/dpdk.go
SchSeba: no need

functests/dpdk/dpdk.go
SchSeba: +1

functests/dpdk/dpdk.go
SchSeba: +1

functests/dpdk/dpdk.go
SchSeba: No I want to leave the user the ability to test on a different interfaces that are not the default gateway interface.

In general in discovery mode the user should know why he creates the policy on a specific interface.

fedepaol: Yep, but even in discovery mode there will be interfaces belonging to the pf where the default gw is, and others belonging to other pfs. Would it make sense to favour the first one?

SchSeba: Let me fix the discovery in a separate PR, I need to cherry-pick this to release 4.5

fedepaol: Agree

PR #293, opened 2020-08-09
add xt_u32 as a feature

yuvalk:
xt_u32 is an iptables filtering kernel module.
The module was removed from RHEL8 and brought back to life becasue there are some important CNFs that relay upon it. It will be removed again for RHEL9,
but till then we wanna have a documented way of enabling it and attesting it works.

This is basically a copy of sctp feature (minus the unblacklisting which isnt required for xt_u32).

### DIFF
--- a/feature-configs/demo/xt_u32/kustomization.yaml
+++ b/feature-configs/demo/xt_u32/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - ../../deploy/xt_u32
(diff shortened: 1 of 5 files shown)


yuvalk:
/assign @fedepaol 

please let me know if this is the way to go.
we'll keep this pr on hold till we open a branch for 4.7 I guess...

fedepaol:
@yuvalk I wonder if it makes sense to add a test that verifies that the module is loaded.

yuvalk:
> @yuvalk I wonder if it makes sense to add a test that verifies that the module is loaded.

yes, tests will be added by QE.
we want a full e2e one, that also attest that it is possible to modify iptables with xt_u32 from a container (with NET_ADMIN)

PR #294, opened 2020-08-10
Delete dpdk failed build

sabinaaledort:
If a dpdk build status is failed, delete it before creating a new build.


yuvalk:
please add explanation in commit body as of WHY is this needed

PR #296, opened 2020-08-13
performance profile: consume api v1

fromanirh:
For 4.6, performance profile api is becoming stable, introducing v1.
Update everything to consume the stable API, which is also more
future proof.


fromanirh:
cc @sabinaaledort @cynepco3hahue @marcel-apf

PR #299, opened 2020-08-18
[release 4.5] Use default network for dpdk test

SchSeba:
cherry-pick from https://github.com/openshift-kni/cnf-features-deploy/pull/290

This PR change to dpdk test to create a policy when the test are not in discovery mode to find and use the default router and use that interface.

I make this change because on the default network there is always
traffic for the testpmd application to process, this makes the test stable.


SchSeba:
/hold

need to validate

SchSeba:
I validate it on my cluster

/hold cancel

SchSeba:
/test e2e-gcp
--

PR #307, opened 2020-08-28
Activate codeready builder for rhel-8.

fedepaol:
Dpdk dockerfile FROM is replaced with the base specified in the
release repo. After changing the base to rhel-8, the developer subscription
needs to be explicitly enabled. Without doing that, the yum install command
won't be able to find libpcap-devel (see also https://bugzilla.redhat.com/show_bug.cgi?id=1707919)


fedepaol:
Closed in favour of https://github.com/openshift/release/pull/11394

PR #312, opened 2020-09-08
Remove the default path for junit

SchSeba:
This commit remove the default path for the junit report.

Before this change we always create a folder with a file name.

### DIFF
--- a/functests/test_suite_test.go
+++ b/functests/test_suite_test.go
@@ -50,8 +50,8 @@ var (
 )
 
 func init() {
-	junitPath = flag.String("junit", "junit.xml", "the path for the junit format report")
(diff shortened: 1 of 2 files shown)


functests/test_suite_test.go
```diff
@@ -50,8 +50,8 @@ var (
 )
 
 func init() {
-	junitPath = flag.String("junit", "junit.xml", "the path for the junit format report")
-	reportPath = flag.String("report", "", "the path of the report file containing details for failed tests")
+	junitPath = flag.String("junit", "", "the path for the junit format report")
```
fedepaol, line 53: Do we need this for all the 3 suites?

SchSeba: The other two already have it empty :)

PR #315, opened 2020-09-16
Update PAO subscription channel and bump dependencies

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

PR #316, opened 2020-09-16
Reduce performance profile hugepages from 8 to 4

sabinaaledort:
When having a cluster with virtual machines as workers, there is not enough memory for all the pods
and the disruption budget will not allow us to drain the baremetal machine.


cynepco3hahue:
To pass the CI we will need this PR https://github.com/openshift-kni/cnf-features-deploy/pull/315 first.

PR #320, opened 2020-09-23
Clean the policy and wait after the sctp-sriov test.

fedepaol:
It's not getting cleaned if sctp-sriov runs before the sriov tests, and may affect its
execution.


SchSeba (approved):
/lgtm

SchSeba:
lets add this in a AfterEach so we are sure we clean this

SchSeba (approved):
/lgtm
/hold cancel

PR #321, opened 2020-09-24
Dump single pod instead of pod list.

fedepaol:
Due to a bug, now all the pods are dumped many times, one for each pod available. We need to dump the list only once.


SchSeba (approved):
/lgtm

PR #322, opened 2020-09-24
Change the way we look for the uplink when using ovs.

fedepaol:
There are some nodes where bridge link do not return anything.
We use ip link then, cross-checking the uplink to be a sriov device.


SchSeba (approved):
/lgtm

PR #332, opened 2020-09-30
Skip setup if in discovery mode.

fedepaol:
If discovery mode is enabled, it does not make sense to run any kind of configuration.


fedepaol:
/hold
Waiting for validation

PR #334, opened 2020-09-30
Skip the negative test if no nodes without sctp are available.

fedepaol:
In regular mode, we ask the behaviour to be defined from outside, but discovery mode is supposed
to inspect the cluster & skip instead of failing.


SchSeba (approved):
/lgtm

PR #336, opened 2020-10-01
Don't call performance clean if discovery mode is enabled.

fedepaol:
If performance clean function deletes the profile, even if it is the discovered one.
When running in discovery mode we want to make sure that the configuration of the cluster is not impacted.

### DIFF
--- a/functests/test_suite_test.go
+++ b/functests/test_suite_test.go
@@ -29,6 +29,7 @@ import (
 	testutils "github.com/openshift-kni/cnf-features-deploy/functests/utils"
 	"github.com/openshift-kni/cnf-features-deploy/functests/utils/clean"
 	testclient "github.com/openshift-kni/cnf-features-deploy/functests/utils/client"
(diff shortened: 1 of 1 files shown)


fromanirh (approved):
since discovery mode doesn't want to create new stuff in the cluster, it surely makes sense to not delete stuff either :)

fromanirh (commented):
/lgtm
/hold
looks ok, but could you please edit the commit message to clarify why using tge env var `CLEAN_PERFORMANCE_PROFILE` is not enough/ideal here?

fedepaol:
> since discovery mode doesn't want to create new stuff in the cluster, it surely makes sense to not delete stuff either :)

Some configurations are created (and deleted), but in case of performance tests the only knob is the profile and right now it would be deleted, hence the if.

fromanirh (commented):
/lgtm
/hold cancel
thanks!

PR #338, opened 2020-10-05
Consume the latest sriov / ptp tests.

fedepaol:
SRIOV Flag tests: ensure that the test pod and the hostnetwork pod run on the same node
PTP: Skip priority profile test in discovery mode, don't use the hardcoded master profile name


SchSeba (approved):
/lgtm

PR #342, opened 2020-10-19
Bug 1889275: When looking for the mc that enables sctp, skip if the raw config is empty

fedepaol:
There are some configurations where we only apply kernel arguments to machine configs.
In this case, the config section is empty and this makes the checks performed by sctp to fail.


SchSeba (approved):
/lgtm

PR #354, opened 2020-11-06
Bug 1894861: Switch to the PAO namespace used in the docs and make the namespaces configurable

fedepaol:
When installing an operator, the user can choose the namespace where to install
it. The common behaviour is to use the one described in the docs, which is different
for the one being used here.

On top of that, we are making the namespaces related to the other two operators
(ptp, sriov) configurable for what is related to the tests implemented here.

Finally, the deploy part is updated to create and use the new default namespace
for PAO.


fedepaol:
/retitle Bug 1894861: Switch to the PAO namespace used in the docs and make the namespaces configurable

fedepaol:
Expected failures, will converge when https://github.com/openshift-kni/performance-addon-operators/pull/437 will be merged

SchSeba (approved):
/lgtm

PR #356, opened 2020-11-09
Update latency test and provide documentation

cynepco3hahue:


cnf-tests/README.md
```diff
@@ -24,6 +24,20 @@ Assuming the kubeconfig file is in the current folder, the command for running t
...
+The `cnf-tests` image include the latency test that uses the `oslat` binary to verify the CPU latency. You can configure
+some important parameters via environment variables:
+
+- LATENCY_TEST_RUN - should be true, if you want to run the latency test(default false).
+- LATENCY_TEST_RUNTIME - how long do you want to run the `oslat` binary(default 5m).
+- OSLAT_MAXIMUM_LATENCY - what the maximum latency do you expect to have during the `oslat` run, the value should be greater than 0.
+- PERF_TEST_PROFILE - the performance profile name(default `performance`).
+
+```bash
+docker run --rm -v $KUBECONFIG:/kubeconfig -e KUBECONFIG=/kubeconfig -e LATENCY_TEST_RUN=true -e LATENCY_TEST_RUNTIME=600 -e OSLAT_MAXIMUM_LATENCY=20 -e PERF_TEST_PROFILE=<performance_profile_name> quay.io/openshift-kni/cnf-tests /usr/bin/test-run.sh -ginkgo.focus="\[performance\]\[config\]|\[performance\]\ Latency\ Test"
```
fedepaol, line 38: Would "just" Latency work? Or, can we think about a higher level label for this test?
I think it would be easier to document / use as opposed to the long string. It can be done in another PR, sorry if I did not think about it initially.

fedepaol (commented):
Added a comment.
Another thought: do you need to update the tests description?

cnf-tests/README.md
cynepco3hahue: Do you mean under the focus? It will run also some SCTP test that includes the latency word under test name.
In general it just a standard focus directive, it can be long but not too complicated. I think a customer will just run all tests, so it will not need to provide a focus in this case. I added this command mostly for the developer's usage.

cynepco3hahue:
> Another thought: do you need to update the tests description?

Already did it under the previous PR.

fedepaol:
> > Another thought: do you need to update the tests description?
> 
> Already did it under the previous PR.

Cool, thanks

cnf-tests/README.md
fedepaol: Yes, thanks for explaining. Agree, it does not make much sense to run latency only.

Maybe with that regard it may make sense to have this section about latency tests (the parameters apply to them even if you run them as part of the performance group), and a subsection that explains how to run them alone.
wdyt?

cynepco3hahue: Totally agree, will add a subsection under running performance tests.

cynepco3hahue: Done

fedepaol:
Thanks!

PR #357, opened 2020-11-09
Add xt_u32 feature tests to the test suite.

sabinaaledort:
xt_u32 is an iptables filtering kernel module.

This test is crearing a pod that has the xt_u32 module enbaled in it,
then running an iptable rule inside it to test the module works.

### DIFF
--- a/cnf-tests/Dockerfile
+++ b/cnf-tests/Dockerfile
@@ -57,9 +57,9 @@ RUN yum install -y numactl-devel make gcc && \
     make oslat && \
     cp oslat /oslat
 
-FROM centos:8
+FROM centos:7
 
-RUN yum install -y lksctp-tools iproute libhugetlbfs-utils libhugetlbfs tmux ethtool numactl-libs iptables
+RUN yum install -y lksctp-tools iproute libhugetlbfs-utils libhugetlbfs tmux ethtool ping numactl-libs
 
 RUN mkdir -p /usr/local/etc/cnf
 
--- a/cnf-tests/Dockerfile
+++ b/cnf-tests/Dockerfile
@@ -57,9 +57,9 @@ RUN yum install -y numactl-devel make gcc && \
     make oslat && \
     cp oslat /oslat
 
-FROM centos:7
+FROM centos:8
 
-RUN yum install -y lksctp-tools iproute libhugetlbfs-utils libhugetlbfs tmux ethtool ping numactl-libs
+RUN yum install -y lksctp-tools iproute libhugetlbfs-utils libhugetlbfs tmux ethtool numactl-libs iptables
 
 RUN mkdir -p /usr/local/etc/cnf
 
--- a/functests/test_suite_test.go
+++ b/functests/test_suite_test.go
@@ -16,7 +16,8 @@ import (
 	_ "github.com/openshift-kni/cnf-features-deploy/functests/dpdk" // this is needed otherwise the dpdk test won't be executed
 	_ "github.com/openshift-kni/cnf-features-deploy/functests/ptp"  // this is needed otherwise the ptp test won't be executed
 	"github.com/openshift-kni/cnf-features-deploy/functests/sctp"
-	_ "github.com/openshift-kni/cnf-features-deploy/functests/sctp" // this is needed otherwise the sctp test won't be executed
+	_ "github.com/openshift-kni/cnf-features-deploy/functests/sctp"   // this is needed otherwise the sctp test won't be executed
+	_ "github.com/openshift-kni/cnf-features-deploy/functests/xt_u32" // this is needed otherwise the xt_u32 test won't be executed
 
 	_ "github.com/openshift-kni/performance-addon-operators/functests/1_performance" // this is needed otherwise the performance test won't be executed
 	_ "github.com/openshift-kni/performance-addon-operators/functests/4_latency" // this is needed otherwise the performance test won't be executed
@@ -124,6 +125,7 @@ var _ = AfterSuite(func() {
 		namespaces.DpdkTest,
 		sctp.TestNamespace,
 		sriovNamespaces.Test,
+		namespaces.XTU32Test,
 	}
 
 	for _, n := range nn {
--- a/functests/utils/namespaces/namespaces.go
+++ b/functests/utils/namespaces/namespaces.go
@@ -28,10 +28,18 @@ var SRIOVOperator = "openshift-sriov-network-operator"
 // PTPOperator is the namespace where the PTP Operator is installed
 var PTPOperator = "openshift-ptp"
 
+// XTU32Test is the namespace of xt_u32 test suite
+var XTU32Test string
+
 func init() {
 	DpdkTest = os.Getenv("DPDK_TEST_NAMESPACE")
 	if DpdkTest == "" {
 		DpdkTest = "dpdk-testing"
+	}
+
+	XTU32Test = os.Getenv("XT_U32_TEST_NAMESPACE")
+	if XTU32Test == "" {
+		XTU32Test = "xt-u32-testing"
 	}
 
 	if performanceOverride, ok := os.LookupEnv("PERFORMANCE_OPERATOR_NAMESPACE"); ok {
--- a/functests/xt_u32/xt_u32.go
+++ b/functests/xt_u32/xt_u32.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	igntypes "github.com/coreos/ignition/config/v2_2/types"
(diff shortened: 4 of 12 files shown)


functests/test_suite_test.go
```diff
@@ -17,7 +17,8 @@ import (
 	_ "github.com/openshift-kni/cnf-features-deploy/functests/ptp"  // this is needed otherwise the ptp test won't be executed
 	"github.com/openshift-kni/cnf-features-deploy/functests/sctp"
 	_ "github.com/openshift-kni/cnf-features-deploy/functests/sctp" // this is needed otherwise the sctp test won't be executed
-
+	"github.com/openshift-kni/cnf-features-deploy/functests/xt_u32"
+	_ "github.com/openshift-kni/cnf-features-deploy/functests/xt_u32"                // this is needed otherwise the xt_u32 test won't be executed
```
fedepaol, line 21: The line above is enough to import the tests. This is needed only if you are not importing any symbol from the package.

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,223 @@
...
+	hasNonCnfWorkers   bool
+)
+
+func init() {
+	roleWorkerCNF := os.Getenv("ROLE_WORKER_CNF")
+	if roleWorkerCNF != "" {
+		xt_u32NodeSelector = fmt.Sprintf("node-role.kubernetes.io/%s=", roleWorkerCNF)
+	}
+
+	hasNonCnfWorkers = true
```
fedepaol, line 43: This was for the negative test, but I don't see it here.
I think this is not needed.

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,223 @@
...
+		Expect(err).ToNot(HaveOccurred())
+
+		if xt_u32NodeSelector == "" {
+			xt_u32NodeSelector, err = findXT_U32NodeSelector()
+			Expect(err).ToNot(HaveOccurred())
+		}
+	})
+
+	BeforeEach(func() {
+		if !hasNonCnfWorkers {
```
fedepaol, line 66: This means that if there are no noncnfworkers, all the tests will be skipped (beforeEach runs for each one of them), which is not what we want.

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,223 @@
...
+		} else {
+			Expect(len(filtered)).To(BeNumerically(">", 0))
+		}
+		testNode = filtered[0].ObjectMeta.Labels[hostnameLabel]
+	})
+
+	Context("Validate the module is enabled and works", func() {
+		It("Should create an iptables rule inside a pod that has the module enabled", func() {
+			By("Create a pod")
+			pod := xt_u32TestPod("xt-u32", testNode, TestNamespace, []string{"sleep inf"})
```
fedepaol, line 91: Since this is a one use, it may make sense to wrap the command in the pod definition function.
Same with the namespace.

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,223 @@
...
+				res, err := client.Client.Pods(TestNamespace).Get(context.Background(), xt_u32Pod.Name, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				return res.Status.Phase
+			}, 3*time.Minute, 1*time.Second).Should(Equal(k8sv1.PodRunning))
+
+			By("Check that xt_u32 module is loaded")
+			cmd := []string{"lsmod"}
+			out, err := pods.ExecCommand(client.Client, *xt_u32Pod, cmd)
+			Expect(err).ToNot(HaveOccurred())
+			res := strings.Contains(out.String(), "xt_u32")
```
fedepaol, line 106: You can use the `ContainSubstring` assertion here

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,223 @@
...
+			if !res {
+				err = errors.New("xt_u32 module is not loaded")
+				Expect(err).ToNot(HaveOccurred())
+			}
+
+			By("Create an iptables rule within the pod")
+			cmd = []string{"iptables", "-t", "filter", "-A", "INPUT", "-i", "eth0", "-m",
+				"u32", "--u32", "6 & 0xFF = 1 && 4 & 0x3FFF = 0 && 0 >> 22 & 0x3C @ 0 >> 24 = 8",
+				"-j", "DROP"}
+			_, err = pods.ExecCommand(client.Client, *xt_u32Pod, cmd)
```
fedepaol, line 116: Can you add the output as a description in case of failure?
I think it will be helpful to triage errors.

fedepaol:
@sabinaaledort I think it makes sense to add the new namespace to the list of those dumped by the reporter https://github.com/openshift-kni/cnf-features-deploy/blob/dbef580f7b5d738b97520641d3979a7ef05ba900/functests/utils/reporter.go#L32

functests/xt_u32/xt_u32.go
yuvalk: but adding one can be a good defensive approach
otherwise we might miss changes like the module loading by default (due to some RHEL/RHCOS change or if OKD/OCP is ever supported on other oses)
it is less important, but still something we'd like to check for.

so I guess it's ok to remove from this commit, but add on a separate commit that will add the negative test.

functests/xt_u32/xt_u32.go
yuvalk: I'd also add a descriptive info print explaining what this line does
so it's less cryptic 
IIUC it drops icmp ping (type=8, echo request) msgs from any source:
ip protocol = 1 (ICMP) && not fragmented && icmp type = 8 (echo request)

yuvalk: btw, seems like this will work on IPv4 only
I guess command itself will still work on IPv6.
(so no err will occur)

functests/xt_u32/xt_u32.go
fedepaol: Agree, also the SCTP approach of having the value injected from outside revealed to be a bit cluncky (especially when somebody else needs to run these tests), so in that case automatic discovery of non SCTP / XT_U32 workers would be better

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,223 @@
...
+			}
+
+			return xt_u32NodeSelector, nil
+		}
+	}
+
+	return "", errors.New("Cannot find a machine configuration that enables XT_U32")
+}
+
+func enablesXT_U32(mc mcfgv1.MachineConfig) (bool, error) {
```
yuvalk, line 149: IIUC this function only try to see if it's configure or not, returning true/false, based on that
so the name is confusing
is_xt_u32_enabled is probably better

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,223 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	return "", errors.New("Cannot find XT_U32NodeSelector")
+}
+
+func xt_u32TestPod(name, node, namespace string, args []string) *k8sv1.Pod {
```
yuvalk, line 190: since that a privileged pod, I clearly state that in the function name,
so anyone using it, will know that.
we probably need a naming convention for privileged operations too?

functests/xt_u32/xt_u32.go
fedepaol: But with camelCase, please :-)

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,223 @@
...
+
+	execute.BeforeAll(func() {
+		err := namespaces.Create(TestNamespace, client.Client)
+		Expect(err).ToNot(HaveOccurred())
+
+		err = namespaces.Clean(TestNamespace, "xt-u32", client.Client)
+		Expect(err).ToNot(HaveOccurred())
+
+		if xt_u32NodeSelector == "" {
+			xt_u32NodeSelector, err = findXT_U32NodeSelector()
```
yuvalk, line 60: not sure, but shouldnt there be some distinction between discovery mode (and non-discovery mode)?
ie create the MC if needed?
@fedepaol

fedepaol: The MC creation is intended to be the "installation of the feature". That's the approach we took with SCTP.
With that, it does not make much distinction between discovery and non discovery, apart from skipping or failing.
@sabinaaledort this reminds me that it would make sense to extend the validation suite (https://github.com/openshift-kni/cnf-features-deploy/blob/73e520f860d35a62e34c1e770a383e29b9c45dc7/validationsuite/cluster/validation.go#L171 ) the same way we did for SCTP

functests/xt_u32/xt_u32.go
sabinaaledort: i can rename the function to xt_u32PrivilegedTestPod
what privileged operations are you referring to?

fedepaol:
@sabinaaledort the linter failed: check the ci lane
Also, this needs rebase

cnf-tests/Dockerfile
```diff
@@ -59,7 +59,7 @@ cp oslat /oslat
 
 FROM centos:7
 
-RUN yum install -y lksctp-tools iproute libhugetlbfs-utils libhugetlbfs tmux ethtool ping numactl-libs
+RUN yum install -y lksctp-tools iproute libhugetlbfs-utils libhugetlbfs tmux ethtool ping numactl-libs iptables
```
fedepaol, line 62: Do we need this also for the Dockerfile.openshift?

functests/test_suite_test.go
```diff
@@ -124,6 +125,7 @@ var _ = AfterSuite(func() {
 		namespaces.DpdkTest,
 		sctp.TestNamespace,
 		sriovNamespaces.Test,
+		xt_u32.TestNamespace,
```
fedepaol, line 128: This should be namespaces.XT_U32Test (also, the name without _ as per the linter suggestion)

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/namespaces"
+	utilNodes "github.com/openshift-kni/cnf-features-deploy/functests/utils/nodes"
+	"github.com/openshift-kni/cnf-features-deploy/functests/utils/pods"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	k8sv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const hostnameLabel = "kubernetes.io/hostname"
+const TestNamespace = "test-xt-u32"
```
fedepaol, line 27: You can remove this and use the one under namespaces

functests/xt_u32/xt_u32.go
```diff
@@ -0,0 +1,208 @@
...
+
+const xt_u32LoadPath = "/etc/modules-load.d/xt_u32-load.conf"
+
+var (
```
fedepaol, line 31: If it's a single var, you can do it in one line `var ... string`

fedepaol:
Another thought: if this does not have any bare metal dependencies, we can deploy & run the tests as part of the u/s CI.
You need to add the feature name to the default list here https://github.com/openshift-kni/cnf-features-deploy/blob/cc193e43e1d6901efdd195e7b51458dac5900cbc/Makefile#L2

functests/utils/namespaces/namespaces.go
```diff
@@ -28,12 +28,20 @@ var SRIOVOperator = "openshift-sriov-network-operator"
...
+
 func init() {
 	DpdkTest = os.Getenv("DPDK_TEST_NAMESPACE")
 	if DpdkTest == "" {
 		DpdkTest = "dpdk-testing"
 	}
 
+	XTU32Test = os.Getenv("XT_U32_TEST_NAMESPACE")
+	if XTU32Test == "" {
+		XTU32Test = "test-xt-u32"
```
yuvalk, line 42: can you change it to be consistent with the above?
ie - xt-u32-testing

PR #360, opened 2020-11-11
Bug 1895871: DPDK test: don't backup and create sriov configuration in regular mode.

fedepaol:
In regular mode it's safe to assume there are no extra configurations in the cluster. Not having to backup and restore will make the test cycle faster, and will reduce the chance of not restoring the configuration because of errors in the middle.


fedepaol:
/hold
still to validate

SchSeba (approved):
/lgtm

/hold

please remove the hold after validation

SchSeba (approved):
/lgtm

/hold cancel

PR #363, opened 2020-11-16
Fix dpdk discovery mode.

fedepaol:
When the discovery of sriov on one nodes of the cluster fails, the discovery function fails even when the discovery is performed successfully on at least one node.


SchSeba (commented):
just a nit can you please change the return from `DpdkResources` to `*DpdkResources` so the return can be nil?

PR #366, opened 2020-11-19
Bug 1890089: Update to latest sriov operator tests.

fedepaol:
Vendor latest sriov tests, which include the fix for Bug #1890089


SchSeba (commented):
/lgtm

PR #367, opened 2020-11-19
Bug 1896798: Make it possible to override the namespace used by the operators, plus switch the pao default one.

fedepaol:
When installing an operator, the user can choose the namespace where to install it. The common behaviour is to use the one described in the docs, which is different for the one being used here.
On top of that, we are making the namespaces related to the other two operators (ptp, sriov) configurable for what is related to the tests implemented here.
The deploy part is updated to create and use the new default namespace for PAO.
We update the pao test to contain the fix.


SchSeba (commented):
/lgtm

PR #368, opened 2020-11-20
Dpdk tests: add a test that runs a pod using netdevice together with dpdk.

fedepaol:
This uses the sriov device partitioning, assigning some vfs to the dpdk payload and some
others to pods using net devices.

Also, changes the num of hugepages to 5 as 4 are needed to the two dpdk pods, one to the regular pod.


functests/dpdk/dpdk.go
```diff
@@ -144,135 +126,204 @@ var _ = Describe("dpdk", func() {
...
+	Context("VFS split for dpdk and netdevice", func() {
+		BeforeEach(func() {
+			if discovery.Enabled() {
+				Skip("Split VF test disabled for discovery mode")
+			}
+		})
+		execute.BeforeAll(func() {
+			createSriovPolicyAndNetworkShared()
+			var err error
+			dpdkWorkloadPod, err = createDPDKWorkload(nodeSelector,
```
SchSeba, line 291: should you clean the old pods before?

SchSeba (commented):
one comment I think we should have some clean between the different parts

functests/dpdk/dpdk.go
fedepaol: That was done inside `createSriovPolicyAndNetworkShared` / `createSriovPolicyAndNetworkDPDKOnly` (as it was before).
Will move the clean function outside, I agree it will improve readability.

functests/dpdk/dpdk.go
```diff
@@ -144,135 +126,206 @@ var _ = Describe("dpdk", func() {
...
+				out, err = pods.GetLog(dpdkWorkloadPod)
+				Expect(err).ToNot(HaveOccurred())
+				return out
+			}, 8*time.Minute, 1*time.Second).Should(ContainSubstring(LOG_ENTRY),
+				"Cannot find accumulated statistics")
+			checkRxTx(out)
+		})
+
+		It("Run a regular pod using a vf shared with the dpdk's pf", func() {
+			podDefinition := pods.DefineWithNetworks(namespaces.DpdkTest, []string{"test-regular-network"})
```
SchSeba, line 324: nit: can use add the network name as `const` and reuse it in the `createSriovPolicyAndNetworkShared` function?

SchSeba (commented):
just a small nit and we can merge this PR thanks for the great job!

fedepaol:
/hold
blocking this until 4.8 branch is available

SchSeba (approved):
/lgtm
/approve

PR #379, opened 2020-11-30
Fix bug in xt_u32 is ready script

sabinaaledort:
The MachineConfig name was updated to load-xt-u32-module as it can't include '_' character


sabinaaledort:
will be part of xt_u32 test pr
https://github.com/openshift-kni/cnf-features-deploy/pull/357

PR #380, opened 2020-12-02
Bug 1898531: Update the vendored ptp tests.

fedepaol:
This changes the configuration to use l2, making it compatible with ipv6 / ipv4 clusters.


fedepaol:
/override ci/prow/e2e-gcp 

gcp passed once, upstream CI does not run ptp.

PR #382, opened 2020-12-04
RHACM manifests to use deploy folder to install CNF components in managed clusters

alosadagrande:
These are the manifests required by ACM to install CNF components stored in [deploy folder|https://github.com/openshift-kni/cnf-features-deploy/tree/release-4.6/feature-configs/deploy/]

Applying these manifests into an ACM hub cluster will install all CNF components (dpdk, sctp, ptp, sriov and pao) in any targeted cluster (spoke) applying the "kustomization" file stored in the operator's deploy folder.

The point here is to have something similar to a single point of truth for installing CNF components. Based on the definitions folder already created and maintained, you choose whether you want to deploy using kubectl/oc (customize) or RHACM, but the base is the same.

Note that, since I just tested on clusters (hub and spoke) running on OCP 4.6 and installing CNF components 4.6, the merge request targets release-4.6 branch. If required, I can take a look at 4.7 version of OCP & CNF components and then merged it into the current master/main.

The proposed folder is feature-configs/ but I am not sure if makes more sense in the cn-ran overlay effort (cc/ @yuvalk @vitus133)

### DIFF
--- a/feature-configs/rhacm-manifests/operator-install/dpdk/dpdk-build-install.yaml
+++ b/feature-configs/rhacm-manifests/operator-install/dpdk/dpdk-build-install.yaml
@@ -1,0 +1,34 @@
+apiVersion: app.k8s.io/v1beta1
+kind: Application
+metadata:
+  name: dpdk-build-install
+  namespace: dpdk
+spec:
+  componentKinds:
+  - group: apps.open-cluster-management.io
+    kind: Subscription
+  descriptor: {}
+  selector:
+    matchExpressions:
+    - key: app
+      operator: In
+      values:
+      - dpdk-build-install
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: Subscription
+metadata:
+  name: dpdk-subscription-4.6
+  namespace: dpdk
+  labels:
+    app: dpdk-build-install
+  annotations:
+    apps.open-cluster-management.io/git-path: feature-configs/deploy/dpdk
+    apps.open-cluster-management.io/git-branch: release-4.6
+spec:
+  name: dpdk-subscription-4.6
+  channel: dpdk/cnf-features-deploy
+  placement:
+    placementRef:
+      kind: PlacementRule
+      name: placement-cluster-cnf21
--- a/feature-configs/rhacm-manifests/operator-install/dpdk/placement-cluster-cnf21.yaml
+++ b/feature-configs/rhacm-manifests/operator-install/dpdk/placement-cluster-cnf21.yaml
@@ -1,0 +1,16 @@
+apiVersion: apps.open-cluster-management.io/v1
+kind: PlacementRule
+metadata:
+  name: placement-cluster-cnf21
+  namespace: dpdk
+spec:
+  clusterSelector:
+    matchExpressions:
+      - key: name
+        operator: In
+        values:
+        - cnf21
+      - key: ocpversion
+        operator: In
+        values:
+        - 4.6.6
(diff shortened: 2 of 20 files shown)


alosadagrande:
A proper README file is required, but first, let's see your opinions.

feature-configs/rhacm-manifests/operator-install/dpdk/dpdk-build-install.yaml
```diff
@@ -0,0 +1,34 @@
...
+apiVersion: apps.open-cluster-management.io/v1
+kind: Subscription
+metadata:
+  name: dpdk-subscription-4.6
+  namespace: dpdk
+  labels:
+    app: dpdk-build-install
+  annotations:
+    apps.open-cluster-management.io/git-path: feature-configs/deploy/dpdk
+    apps.open-cluster-management.io/git-branch: release-4.6
```
yuvalk, line 27: let's do master and then port back if we want
which also imply let's focus here on u/s versions

feature-configs/rhacm-manifests/operator-install/dpdk/placement-cluster-cnf21.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: placement-cluster-cnf21
+  namespace: dpdk
+spec:
+  clusterSelector:
+    matchExpressions:
+      - key: name
+        operator: In
+        values:
+        - cnf21
```
yuvalk, line 12: what's cnf21?
is that just to serve as an example?

alosadagrande: cnf21 is the name (see the key field in the `placementRule`) of the targetted cluster. Yes, it is an example of `placementRule`, since this object is required to let ACM know which cluster or clusters is/are going to install the CNF components.

feature-configs/rhacm-manifests/operator-install/dpdk/dpdk-build-install.yaml
alosadagrande: The point of doing master is that the maintained deploy folder in this branch points to a 4.7 snapshot index of the performance addon operator, see the [catalogSource](https://github.com/openshift-kni/cnf-features-deploy/blob/master/feature-configs/deploy/performance/operator_catalogsource.yaml) while PTP or SRIOV are still targeting the official 4.6 release.

Currently, RHACM runs on OpenShift 4.6, so I do not know if it is possible to run PAO 4.7 on an OpenShift cluster 4.6 .

What we can do is move the PR to the master branch, but RHACM Git/Subscription manifests still pointing to the release-4.6 branch of deploy/ folder as it is right now and as it has been verified.

fedepaol:
@alosadagrande do you expect to run cnf-tests against a cluster configured with this layer?

If so, would it make sense to add the `is_ready.sh` file so we can ensure the cluster is ready before running the tests?

feature-configs/rhacm-manifests/operator-install/dpdk/dpdk-build-install.yaml
fedepaol: Ah, just noticed this points to release-4.6
What we normally do, is to have master pointing to the last working version until the "right version" is available.
This is the reason why the sriov-operator installed is still 4.6. I'd second @yuvalk approach of pushing to master and backporting. If we don't do that, I am afraid of missing pieces in the process.

PR #383, opened 2020-12-07
Disable chronyd when using ptp

SchSeba:
This commit add a machine config to disable chronyd under the ptp kustomize

### DIFF
--- a/feature-configs/cn-ran-overlays/ran-profile/cluster-config/du-common/ptp/disable-chronyd.yaml
+++ b/feature-configs/cn-ran-overlays/ran-profile/cluster-config/du-common/ptp/disable-chronyd.yaml
@@ -1,0 +1,35 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    # Pay attention to the node label, create MCP accordingly
+    machineconfiguration.openshift.io/role: worker-cnf
+  name: disable-chronyd
+spec:
+  config:
+    ignition:
+      version: 2.2.0
+    systemd:
+      units:
+        - contents: |
+            [Unit]
+            Description=NTP client/server
+            Documentation=man:chronyd(8) man:chrony.conf(5)
+            After=ntpdate.service sntp.service ntpd.service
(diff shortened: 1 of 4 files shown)


SchSeba:
@vitus133 please take a look

vitus133:
@SchSeba 
Thanks, got it.

feature-configs/cn-ran-overlays/ran-profile/cluster-config/du-common/ptp/disable-chronyd.yaml
```diff
@@ -0,0 +1,35 @@
...
+            Documentation=man:chronyd(8) man:chrony.conf(5)
+            After=ntpdate.service sntp.service ntpd.service
+            Conflicts=ntpd.service systemd-timesyncd.service
+            ConditionCapability=CAP_SYS_TIME
+
+            [Service]
+            Type=forking
+            PIDFile=/run/chrony/chronyd.pid
+            EnvironmentFile=-/etc/sysconfig/chronyd
+            ExecStart=/usr/sbin/chronyd $OPTIONS
```
fedepaol, line 26: Is $OPTIONS being set somewhere?

SchSeba: nope this is the file that exist in RHCOS I just copy the file here the important part is the `enable:false` all the other is just to don't change the file content

fedepaol: understood, thanks

fedepaol:
@SchSeba mind adding some context on why we are disabling chronyd?

Also, would it make sense to have it part of the ptp deployment instad of the profile / the demo layer?
Which boils down to: does it make sense to expect this to be deployed and should the validation step check for it?

SchSeba:
I put it here because I don't want to ptp deployment to reboot the node (when we apply the MC) also if the user deploy ptp it may take time until he apply the policy and the node may lose time sync.

PR #384, opened 2020-12-08
Add gatekeeper testing

marzianor9:


functests/utils/dynamic/dynamic.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+}
+
+// Resource returns a namespaceable interface for the given resource
+func (c *Client) Resource(resourceName string) clientGoDynamic.NamespaceableResourceInterface {
+	return c.resource[resourceName]
+}
+
+// Assert returns a boolean to indicate the existence of an interface for the given resource name
+func (c *Client) Assert(resourceName string) bool {
```
mmirecki, line 60: Note that this only checks that you have added something to the dict.
Also `c.dynamicInterface.Resource(gvr)` will not raise an error when the resource does not exist (I think it does not even return nil, but a client for a non existent resource).
Maybe in the assert you can check if the resource actually exists by trying to list it:


```
 func (c *Client) Assert(resourceName string) bool {
	clientForResource, ok := c.resource[resourceName]
       	crds, err := clientForResource.List(metav1.ListOptions{})
        return err == nil
}
```

functests/utils/dynamic/dynamic.go
```diff
@@ -0,0 +1,63 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	clientGoDynamic "k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// Client provides a struct for interacting with the openshift api
+type Client struct {
+	dynamicInterface clientGoDynamic.Interface
+	resource         map[string]clientGoDynamic.NamespaceableResourceInterface
```
mmirecki, line 16: maybe rename: clientForResource?

functests/utils/dynamic/dynamic.go
```diff
@@ -0,0 +1,63 @@
+package dynamic
```
mmirecki, line 1: "dynamic" will cause it to be confused with "k8s.io/client-go/dynamic"

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+var _ = Describe("gatekeeper", func() {
+	// we make the interface key to a var for runtime key check
+	// if we used the string directly a typo will lead to a runtime error
+	// typos in var names will make a compile error
+	assignMetadata := "assignmetadata"
+	pods := "pods"
+	// create the dynamic client
+	client := dynamic.New("")
```
mmirecki, line 38: On first look I assumed this is the standard dynamic client (of the client-go libs). Maybe we could go with something like testClient to avoid confustion (and packge naming colision)?

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,151 @@
...
+				},
+			},
+		),
+	)
+
+	DescribeTable("Do not mutate existing metadata info",
+		func(amInfo assignMetadataInfo, resource *unstructured.Unstructured) {
+			amDefinition := assignMetadataDefinition(&amInfo)
+
+			_, err := client.Resource(assignMetadata).Create(context.Background(), amDefinition, metav1.CreateOptions{})
```
mmirecki, line 93: `client.resource[assignMetadata]` already contains this client. Either use that one, or don't precreate them in the client.

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,168 @@
...
+	pods := "pods"
+
+	// create a gvr list for to resources we want an interface with
+	gvrList := []schema.GroupVersionResource{
+		{
+			Group:    "mutations.gatekeeper.sh",
+			Version:  "v1alpha1",
+			Resource: assignMetadata,
+		},
+		{
```
fedepaol, line 45: I'd use the regular client for this

fedepaol (commented):
Looks like a good start!
Added a few comments

functests/gatekeeper/resources.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// Resources is a list of unstructured resources for the gatekeeper test
+var Resources []unstructured.Unstructured
+
+func init() {
+	Resources = []unstructured.Unstructured{
```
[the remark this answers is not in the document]
marzianor9: when it's unstructured we are able the get a general resource test, and since this test is for labels/annotations we get free polymorphism.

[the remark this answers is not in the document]
fedepaol: Ah, nice!
Thanks for explaining, makes sense to me.

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+			},
+			{
+				kind:  "annotations",
+				key:   "mutated",
+				value: "true",
+			},
+		}...)...,
+	)
+
+	DescribeTable("should not mutate existing metadata info(labels/annotations)",
```
SchSeba, line 87: you can combine the two describe tables I think to use with 4 tests

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+			},
+			{
+				kind:  "annotations",
+				key:   "changed",
+				value: "true",
+			},
+		}...)...,
+	)
+
+	It("should apply mutations by order", func() {
```
SchSeba, line 125: The previews comments about Before and After also using the GK API scheme is related to this test

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+		amB := assignMetadataDefinition(amDefB)
+		amB, err := client.Resource("assignmetadata").Create(context.Background(), amB, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		defer func() {
+			err := client.Resource("assignmetadata").Delete(context.Background(), amB.GetName(), metav1.DeleteOptions{})
+			Expect(err).ToNot(HaveOccurred())
+		}()
+
+		By("Creating test-pod-b")
+		testPodB := GetTestPod(TestingNamespace)
```
SchSeba, line 147: I think we already have a create pod function no?

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+		}()
+
+		By("Asserting that test-pod-version-0 was labeled with mutation-version: 0")
+		value, ok, err := unstructured.NestedString(testPodA.Object, "metadata", amDef.metadata.kind, amDef.metadata.key)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ok).To(BeTrue())
+		Expect(value).To(Equal("0"))
+
+		By("Updating assignMetadata to mutation-version: 1")
+		err = unstructured.SetNestedField(am.Object, "1", "spec", "parameters", "assign", "value")
```
SchSeba, line 240: you know the object type you are working with here so lets use the object and not the `unstructured`

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+		testPod.SetName("test-pod-after-delete")
+		testPod.Unstructured, err = client.Resource("pods").Namespace(testPod.GetNamespace()).Create(context.Background(), testPod.Unstructured, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Asserting that test-pod-after-delete was not labeld by the deleted policy")
+		_, ok, err = unstructured.NestedString(testPod.Object, "metadata", amDef.metadata.kind, amDef.metadata.key)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ok).To(BeFalse())
+
+		By("Deleting test-pod-after-delete")
```
SchSeba, line 317: this is not part of the test should be in the afterEach clean

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+		err = client.Resource("pods").Namespace(testPod.GetNamespace()).Delete(context.Background(), testPod.GetName(), metav1.DeleteOptions{})
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("should be able to match by any match category", func() {
+		By("Creating the test namespaces")
+		includedNamepsace := &unstructured.Unstructured{
```
SchSeba, line 324: this is a namespace object just use the namespace schema

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+			Expect(value).To(Equal("true"))
+		}
+	})
+})
+
+func createTableEntries(suffix string, metaInfos ...metadataInfo) (entries []TableEntry) {
+	for _, metaInfo := range metaInfos {
+		objects := GetNamespaceScopeTestObjects(TestingNamespace)
+		for _, object := range objects {
+			object.SetName(fmt.Sprintf("%s-%s-%s", object.GetName(), metaInfo.kind[:3], suffix))
```
SchSeba, line 592: what this `metaInfo.kind[:3]` means?

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+						metadata: metaInfo,
+					},
+					object,
+				))
+		}
+	}
+	return entries
+}
+
+func assignMetadataDefinition(amInfo *assignMetadataInfo) *unstructured.Unstructured {
```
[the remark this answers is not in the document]
marzianor9: we created the dynamic client because gatekeeper didn't have the API struct we needed in the repo, and we were not sure if they want to add them.

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+	client, err := dynamicInterface.New("", gvrList...)
+	if err != nil {
+		log.Fatalf("Unable to create the dynamic client: %v", err)
+	}
+
+	DescribeTable("should be able to add metadata info(labels/annotations)",
+		func(amInfo assignMetadataInfo, object TestObject) {
+			amObject := assignMetadataDefinition(&amInfo)
+
+			amObject, err := client.Resource("assignmetadata").Create(context.Background(), amObject, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
marzianor9: i didn't want to use 2 clients in the same test suite, and the dynamic client allows for more generic tests.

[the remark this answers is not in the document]
marzianor9: and i was unable to add the schema for this object to the client

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+	}
+
+	objects := GetAllTestObjects(TestingNamespace)
+	for _, object := range objects {
+		gvrList = append(gvrList, object.GroupVersionResource())
+	}
+
+	// create the dynamic client
+	client, err := dynamicInterface.New("", gvrList...)
+	if err != nil {
```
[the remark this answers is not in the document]
marzianor9: ?i can't create a variable that will be used in execute.BeforeAll, i will need to make it global, should i do it?

functests/gatekeeper/gatekeeper.go
marzianor9: i needed a customized pod for the tests, also i needed it to be unstructured so that tests can be generic.

functests/gatekeeper/gatekeeper.go
marzianor9: the gatekeeper repo does not support this.

functests/gatekeeper/gatekeeper.go
marzianor9: do we want to use 2 clients?

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+	DescribeTable("should not mutate existing metadata info(labels/annotations)",
+		func(amInfo assignMetadataInfo, object TestObject) {
+			amDefinition := assignMetadataDefinition(&amInfo)
+
+			amDefinition, err := client.Resource("assignmetadata").Create(context.Background(), amDefinition, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			defer func() {
+				err := client.Resource("assignmetadata").Delete(context.Background(), amDefinition.GetName(), metav1.DeleteOptions{})
+				Expect(err).ToNot(HaveOccurred())
+			}()
```
mmirecki: We have quite a lot of these defers, so maybe we can do:

```
func deleteResource(resource string, name string) func() {
   return func() {
	err := client.Resource(resource).Delete(context.Background(), name, metav1.DeleteOptions{})
  }
}

```
and then use it like:
`
defer deleteResource("assignmetadata", amDefinition.GetName())()()
`

With 19 defer statements we should save more then 50 lines of code.

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+			err := client.Resource("pods").Namespace(testPodA.GetNamespace()).Delete(context.Background(), testPodA.GetName(), metav1.DeleteOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+		}()
+
+		By("Asserting that test-pod-a was labeled")
+		value, ok, err = unstructured.NestedString(testPodA.Object, "metadata", amDefA.metadata.kind, amDefA.metadata.key)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ok).To(BeTrue())
+		Expect(value).To(Equal(amDefA.metadata.value))
```
mmirecki: This test could pass if mutators are not ordered always the same, but ordered by order of adding them.
Maybe we can check if adding first A then B also produces the same result, or add a 3rd mutator which would be applied last, but not be applied before mutation-a.

The test would like like:

```
apply mutation-b 
create pod 
check mutation-b is applied

apply mutation-a
create pod 
check mutation-a is applied

apply mutation-c
create pod 
check mutation-a is applied (NOT C)
```
For mutator ordering check:
https://github.com/open-policy-agent/gatekeeper/blob/master/pkg/mutation/assignmeta_mutator.go#L69
https://github.com/open-policy-agent/gatekeeper/blob/master/pkg/mutation/assign_mutator.go#L55

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,637 @@
...
+					},
+				},
+			},
+		}
+		_, err = client.Resource("assignmetadata").Create(context.Background(), namespaceSelected, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		defer func() {
+			err := client.Resource("assignmetadata").Delete(context.Background(), namespaceSelected.GetName(), metav1.DeleteOptions{})
+			Expect(err).NotTo(HaveOccurred())
+		}()
```
mmirecki: It should be more readable if we defined the Assigns in one place and then created them in a loop. Something like:

```
allScopeSelected = {...}
namespaceScopeIncluded = {...}
namespaceScopeIncluded= {...}
...

mutations = { allScopeSelected,  namespaceScopeIncluded, namespaceScopeIncluded}

for mutation ;= range mutatinos {
    client.Resource("assignmetadata").Create(context.Background(), namespaceSelected, metav1.CreateOptions{})
    defer func(){}
}
```

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,644 @@
...
+		Expect(err).ToNot(HaveOccurred())
+
+		labels = clusterObject.GetLabels()
+
+		value, ok := labels["all-selected"]
+		Expect(ok).To(Equal(true))
+		Expect(value).To(Equal("true"))
+
+		value, ok = labels["namespace-included"]
+		Expect(ok).To(Equal(false))
```
oribon, line 612: _, ok = ... is enough here

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,644 @@
...
+
+		value, ok = labels["cluster-selected"]
+		Expect(ok).To(Equal(true))
+		Expect(value).To(Equal("true"))
+
+		value, ok = labels["namespace-selected"]
+		Expect(ok).To(Equal(false))
```
oribon, line 619: same here

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,644 @@
...
+		Expect(value).To(Equal("true"))
+
+		value, ok = labels["namespace-included"]
+		Expect(ok).To(Equal(false))
+
+		value, ok = labels["cluster-selected"]
+		Expect(ok).To(Equal(false))
+
+		value, ok = labels["namespace-selected"]
+		Expect(ok).To(Equal(false))
```
oribon, line 640: same :)

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,644 @@
...
+		By("Asserting that includedNamespaced object mutations were applied")
+		includedNamespacedObjectKey, err := k8sClient.ObjectKeyFromObject(includedNamespacedObject)
+		Expect(err).NotTo(HaveOccurred())
+		err = client.Get(context.Background(), includedNamespacedObjectKey, includedNamespacedObject)
+		Expect(err).NotTo(HaveOccurred())
+
+		labels = includedNamespacedObject.GetLabels()
+
+		value, ok = labels["all-selected"]
+		Expect(ok).To(Equal(false))
```
oribon, line 651: same :)

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,644 @@
...
+
+		value, ok = labels["namespace-included"]
+		Expect(ok).To(Equal(true))
+		Expect(value).To(Equal("true"))
+
+		value, ok = labels["cluster-selected"]
+		Expect(ok).To(Equal(false))
```
oribon, line 658: same :)

oribon (commented):
should be more readable not assigning value if it is not used

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,684 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		namespacesUsed := []string{"mutation-included", "mutation-excluded", "gk-test-object"}
+		for _, namespace := range namespacesUsed {
+			if namespaces.Exists(namespace, client) {
+				err := client.Namespaces().Delete(context.Background(), namespace, metav1.DeleteOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				namespaces.WaitForDeletion(client, namespace, time.Minute)
+			}
+		}
```
mmirecki, line 42: If you care about test time, then maybe move the WaitForDeletion after all namespaces are deleted: 

```
		for _, namespace := range namespacesUsed {
			if namespaces.Exists(namespace, client) {
				err := client.Namespaces().Delete(context.Background(), namespace, metav1.DeleteOptions{})
			}
		}

		for _, namespace := range namespacesUsed {
			if namespaces.Exists(namespace, client) {
				namespaces.WaitForDeletion(client, namespace, time.Minute)
			}
		}
```

marzianor9: the namespaces i delete are used only on the last test, all other times we only delete the pods from the TestingNamespace.

mmirecki: If time is not of essence we can skip this

marzianor9: fixed

functests/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,691 @@
...
+
+		for _, namespace := range namespacesUsed {
+			if namespaces.Exists(namespace, client) {
+				err := client.Namespaces().Delete(context.Background(), namespace, metav1.DeleteOptions{})
+				Expect(err).ToNot(HaveOccurred())
+			}
+		}
+
+		for _, namespace := range namespacesUsed {
+			if namespaces.Exists(namespace, client) {
```
mmirecki, line 45: We can skip `if namespaces.Exists(namespace, client)`, this is done in WaitForDeletion

PR #386, opened 2020-12-08
use the same interface for both ptp profiles

SchSeba:
this commit change the grandmaster profile to use the same profile
as the salve one


fedepaol:
@SchSeba mind commenting the rationale behind this change in the commit?
Apart from this, looks good to me

PR #388, opened 2020-12-10
Bug 1902118: force the gorilla/websocket version to 1.4.2

fedepaol:
This is to fix CVE-2020-27813, although no websocket is used by the image and this is an indirect dependency.


SchSeba (approved):
/lgtm

PR #389, opened 2020-12-13
Use standalone kustomize tool

vitus133:
Will download and use the latest
kustomize tool version  instead
of the one integrated in the OC_TOOL

As of v3.7.0 Kustomize supports a special type of kustomization that allows one to define reusable pieces of configuration logic that can be included from multiple overlays.
Components come in handy when dealing with applications that support multiple optional features and you wish to enable only a subset of them in different overlays, i.e., different features for different environments or audiences.

It is not clear when kustomize 3.x will be integrated into the kubectl, current supported version is 2.x
The integration is tracked here: https://github.com/kubernetes/kubectl/issues/818

### DIFF
--- a/hack/feature-deploy.sh
+++ b/hack/feature-deploy.sh
@@ -1,7 +1,26 @@
 #!/bin/bash
 
+export KUSTOMIZE_DIR="${KUSTOMIZE_DIR:-/tmp}"
+export KUSTOMIZE_BIN=$KUSTOMIZE_DIR/kustomize
+export KUSTOMIZE_VERSION=3.8.8
+
+# expect oc to be in PATH by default
+export OC_TOOL="${OC_TOOL:-oc}"
+
 set -e
+
 . $(dirname "$0")/common.sh
+
+
+# Function: download the specified version of the Kustomize binary to the $KUSTOMIZE_DIR
+function get_kustomize_binary (){
+    (pushd $KUSTOMIZE_DIR
+    curl -m 600 -s "https://raw.githubusercontent.com/\
+kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"\
+  | bash -s $KUSTOMIZE_VERSION
+    popd)
+}
+
 
 if [ "$FEATURES_ENVIRONMENT" == "" ]; then
 	echo "[ERROR]: No FEATURES_ENVIRONMENT provided"
@@ -13,8 +32,11 @@ if [ "$FEATURES" == "" ]; then
 	exit 1
 fi
 
-# expect oc to be in PATH by default
-export OC_TOOL="${OC_TOOL:-oc}"
+if [ ! -f $KUSTOMIZE_BIN ]; then
+  echo "Downloading the Kustomize tool"
+  get_kustomize_binary
+fi
+
 
 # Deploy features
 success=0
@@ -38,9 +60,13 @@ do
     set +e
     # be verbose on last iteration only
     if [[ $iterations -eq $((max_iterations - 1)) ]] || [[ -n "${VERBOSE}" ]]; then
-      ${OC_TOOL} apply -k "$feature_dir"
+      # ${OC_TOOL} apply -k "$feature_dir"
+      # TODO: revert to the above after https://github.com/kubernetes/kubectl/issues/818 is merged:
+      $KUSTOMIZE_BIN build "$feature_dir" | ${OC_TOOL} apply -f -
     else
-      ${OC_TOOL} apply -k "$feature_dir" &> /dev/null
+      # ${OC_TOOL} apply -k "$feature_dir" &> /dev/null
+      # TODO: revert to the above after https://github.com/kubernetes/kubectl/issues/818 is merged:
+      $KUSTOMIZE_BIN build "$feature_dir" | ${OC_TOOL} apply -f - &> /dev/null
     fi
 
     # shellcheck disable=SC2181


hack/feature-deploy.sh
```diff
@@ -38,9 +54,9 @@ do
     set +e
     # be verbose on last iteration only
     if [[ $iterations -eq $((max_iterations - 1)) ]] || [[ -n "${VERBOSE}" ]]; then
-      ${OC_TOOL} apply -k "$feature_dir"
+      $KUSTOMIZE_BIN build "$feature_dir" | ${OC_TOOL} apply -f -
```
fedepaol, line 57: Can you add a TODO to revert this to oc once the kustomize feature is in?

hack/feature-deploy.sh
```diff
@@ -13,6 +24,11 @@ if [ "$FEATURES" == "" ]; then
 	exit 1
 fi
 
+if ! stat $KUSTOMIZE_BIN &> /dev/null; then
```
fedepaol, line 27: you can use
`if [ -f $KUSTOMIZE_BIN ]; then`

fedepaol:
Thanks for the pr, left a few comments.

hack/feature-deploy.sh
```diff
@@ -1,5 +1,16 @@
...
+export KUSTOMIZE_BIN=$KUSTOMIZE_DIR/kustomize
+
+# Function: download the latest Kustomize binary to /tmp
+function get_kustomize_binary (){
+    pushd /tmp
+    curl -s "https://raw.githubusercontent.com/\
+kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+    popd
+}
+
```
yuvalk, line 13: this whole function should move after the common part

hack/feature-deploy.sh
```diff
@@ -1,5 +1,16 @@
 #!/bin/bash
 
+export KUSTOMIZE_DIR="${KUSTOMIZE_DIR:-/tmp}"
+export KUSTOMIZE_BIN=$KUSTOMIZE_DIR/kustomize
+
+# Function: download the latest Kustomize binary to /tmp
+function get_kustomize_binary (){
+    pushd /tmp
```
yuvalk, line 8: shouldnt that use $KUSTOMIZE_DIR ?

hack/feature-deploy.sh
```diff
@@ -1,5 +1,16 @@
...
+    curl -s "https://raw.githubusercontent.com/\
+kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+    popd
```
yuvalk, line 11: i'd say opening Parentheses and cd'ing is more error resilient:
```
(cd /tmp; curl ...)
```

in the former case, if the curl command fail (and set -e is enabled)
you'll be left in that directory instead of returning

hack/feature-deploy.sh
yuvalk: that TODO comment can link to
https://github.com/kubernetes/kubectl/issues/818
when that it closed, it'll get merged in the next oc too

hack/feature-deploy.sh
```diff
@@ -1,5 +1,16 @@
 #!/bin/bash
 
+export KUSTOMIZE_DIR="${KUSTOMIZE_DIR:-/tmp}"
+export KUSTOMIZE_BIN=$KUSTOMIZE_DIR/kustomize
+
+# Function: download the latest Kustomize binary to /tmp
+function get_kustomize_binary (){
+    pushd /tmp
+    curl -s "https://raw.githubusercontent.com/\
```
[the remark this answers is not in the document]
yuvalk: can pass the desired version as parameter to this script

yuvalk, line 9: I think we should add a
```
-m 600
```
so that curl will never hang

yuvalk: BTW: as a result of this review
I've opened https://github.com/kubernetes-sigs/kustomize/pull/3357
so the downloaded script here will also behave nicely

IMHO 10minutes is a sane timeout

hack/feature-deploy.sh
vitus133: Hi @yuvalk!
If I move function `get_kustomize_binary` after the common part, I will have to encapsulate the common part into the `main` function. And even then main should be the bottom most function, according to this: [Google's shell style guide](https://google.github.io/styleguide/shell.xml).

yuvalk: not sure what do you mean by `main` function, we dont have have a `main` in this script.

common is just [this](https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/common.sh) script 
so I'm not sure what's the problem with moving the function adter the line that source it
the added value is the `set -e` which I dont think we should have any code without it
and even in special cases, it's better to explicitly `set +e` (like we are doing elsewhere [here for example](https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/feature-deploy.sh#L38)

yuvalk: I merely meant that you should move the function definition after the line that source the common.sh

vitus133: Ah sorry I misunderstood it. Now I got  what you mean by the common part.

yuvalk:
> Will download and use the latest
> kustomize tool version instead
> of the one integrated in the OC_TOOL
> 
> As of v3.7.0 Kustomize supports a special type of kustomization that allows one to define reusable pieces of configuration logic that can be included from multiple overlays.
> Components come in handy when dealing with applications that support multiple optional features and you wish to enable only a subset of them in different overlays, i.e., different features for different environments or audiences.
> 
> It is not clear when kustomize 3.x will be integrated into the kubectl, current supported version is 2.x
> The integration is tracked here: [kubernetes/kubectl#818](https://github.com/kubernetes/kubectl/issues/818)

and one last small request - can you add the 2 last paragraphs you wrote into the commit msg?
so it will also explain the 'why' of this change

vitus133:
@yuvalk @fedepaol Looks like all the tests are passing now!
Could you guys please move it forward?

fedepaol:
/hold
so @yuvalk can approve too, but it's good on my side

Issue #374, opened 2020-11-27
CNF-Tests: The CI checks only for missing tests, not for tests that are removed


fedepaol:
When we remove a test, there is no check that forces us to remove it from the description too.

PR #393, opened 2020-12-21
added test removed from test suite case

oribon:
added to the fill subcommand check of removed tests that are still in the jsons 
https://github.com/openshift-kni/cnf-features-deploy/issues/374

### DIFF
--- a/cnf-tests/docgen/cmd/fill.go
+++ b/cnf-tests/docgen/cmd/fill.go
@@ -28,6 +28,8 @@ type TestSuite struct {
 }
 
 const emptyPlaceHolder = "XXXXXXXX"
+const errMissing = "Found tests with no description"
+const errRemoved = "Found tests that were removed"
 
 var (
 	junit       string
@@ -78,11 +80,21 @@ func fill(xmlFile, descriptionsFile string) {
 
 func fillDescriptions(fileName string, tests TestSuite, currentDescriptions map[string]string) error {
 	missing := false
+	removed := false
+	allTests := make(map[string]bool)
 	for _, t := range tests.Testcase {
 		if _, ok := currentDescriptions[t.Name]; !ok {
 			currentDescriptions[t.Name] = emptyPlaceHolder
-			fmt.Printf("The test %s do not have a valid description\n", t.Name)
+			fmt.Printf("The test %s does not have a valid description\n", t.Name)
 			missing = true
+		}
+		allTests[t.Name] = true
+	}
+	for k := range currentDescriptions {
+		if _, ok := allTests[k]; !ok {
+			delete(currentDescriptions, k)
+			fmt.Printf("The test %s was removed from the test suite\n", k)
+			removed = true
 		}
 	}
 	jsonData, err := json.MarshalIndent(currentDescriptions, "", "    ")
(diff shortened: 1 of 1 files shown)


yuvalk:
please squash 
and improve the commit msg
maybe something like
`added the ability to test for removed cases`
otherwise, it looks like this commit add a test that was removed by mistake

cnf-tests/docgen/cmd/fill.go
```diff
@@ -78,21 +78,49 @@ func fill(xmlFile, descriptionsFile string) {
...
-		return fmt.Errorf("Found tests with no description")
+		return fmt.Errorf(errMissing)
+	}
+	if removed {
+		return fmt.Errorf(errRemoved)
 	}
 	return nil
 }
+
+func sliceContains(sl []string, s string) bool {
```
fedepaol, line 117: This function does more than the name suggests: it also deletes the string if it is contained.
Can you name it accordingly?

fedepaol: Also, what's the point of deletion?

cnf-tests/docgen/cmd/fill.go
```diff
@@ -78,21 +78,49 @@ func fill(xmlFile, descriptionsFile string) {
...
+			removed = true
+		}
 	}
 	jsonData, err := json.MarshalIndent(currentDescriptions, "", "    ")
 	err = ioutil.WriteFile(fileName, jsonData, os.ModePerm)
 	if err != nil {
 		return fmt.Errorf("failed to open the descriptions file %v", err)
 	}
-
+	if missing && removed {
```
fedepaol, line 105: you can use a switch here instead of a sequence of ifs
```
    switch {
        case missing && removed:
             return blablabla
```

cnf-tests/docgen/cmd/fill.go
oribon: no need to iterate over tests that were already found in the json

fedepaol: still, I don't get it. Why you remove from `allTests` if it's throw-away (and not exported outside)?

cnf-tests/docgen/cmd/fill.go
```diff
@@ -78,21 +78,50 @@ func fill(xmlFile, descriptionsFile string) {
 
 func fillDescriptions(fileName string, tests TestSuite, currentDescriptions map[string]string) error {
 	missing := false
+	removed := false
+	errMissing := "Found tests with no description"
```
fedepaol, line 82: you can move these error descriptions down, near where you use them. It will make it easier to find them

cnf-tests/docgen/cmd/fill.go
```diff
@@ -78,21 +78,49 @@ func fill(xmlFile, descriptionsFile string) {
...
+			fmt.Printf("The test %s was removed from the test suite\n", k)
+			removed = true
+		}
 	}
 	jsonData, err := json.MarshalIndent(currentDescriptions, "", "    ")
 	err = ioutil.WriteFile(fileName, jsonData, os.ModePerm)
 	if err != nil {
 		return fmt.Errorf("failed to open the descriptions file %v", err)
 	}
+	errMissing := "Found tests with no description"
```
SchSeba, line 103: nit: lets just configure this as const so we don't need to allocate in runtime

cnf-tests/docgen/cmd/fill.go
```diff
@@ -78,21 +78,49 @@ func fill(xmlFile, descriptionsFile string) {
...
+	errRemoved := "Found tests that were removed"
```
SchSeba, line 104: nit: lets just configure this as const so we don't need to allocate in runtime

cnf-tests/docgen/cmd/fill.go
```diff
@@ -78,21 +78,49 @@ func fill(xmlFile, descriptionsFile string) {
...
+	case removed:
+		return fmt.Errorf(errRemoved)
+	default:
+		return nil
+	}
+}
 
-	if missing {
-		return fmt.Errorf("Found tests with no description")
+func findAndRemoveElement(sl []string, s string) bool {
```
SchSeba, line 117: please lets use a map and just go over the list one time in the insert time then we can just search with the key

SchSeba (commented):
nice job! just a few comments

PR #394, opened 2020-12-21
Add is_ready.sh script for ran-profile deployment

juphoff:
This commit adds a script to check if a ran-profile has been successfully deployed.

### DIFF
--- a/feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
+++ b/feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
@@ -1,14 +1,3 @@
 #!/bin/bash
 
-# Check for updated worker-cnf node
-
-oc wait mcp/worker-cnf --for condition=updated --timeout 1s
-
-
-# Check for real-time kernel.
-
-WORKER_NODE=$(oc get nodes --selector="node-role.kubernetes.io/worker-cnf" --output="name")
-
-echo "worker-cnf node is: ${WORKER_NODE}"
-
-oc describe "${WORKER_NODE}" | grep "Kernel Version:.*\.rt[0-9]*\..*\.x86_64"
+oc wait ${CNF_MCP:-"mcp/worker-cnf"} --for condition=updated --timeout 1s
--- a/feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
+++ b/feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# Check for updated worker-cnf node
+
+oc wait mcp/worker-cnf --for condition=updated --timeout 1s
+
+
+# Check for real-time kernel.
+
+WORKER_NODE=$(oc get nodes --selector="node-role.kubernetes.io/worker-cnf" --output="name")
+
+echo "worker-cnf node is: ${WORKER_NODE}"
+
+oc describe "${WORKER_NODE}" | grep "Kernel Version:.*\.rt[0-9]*\..*\.x86_64"
--- a/feature-configs/cn-ran-overlays/ran-profile/cluster-config/du-common/ptp/ptpconfig-grandmaster.yaml
+++ b/feature-configs/cn-ran-overlays/ran-profile/cluster-config/du-common/ptp/ptpconfig-grandmaster.yaml
@@ -7,7 +7,7 @@ spec:
   profile:
   - name: "grandmaster"
 # The interface name is hardware-specific    
-    interface: "ens1f0"
+    interface: "ens4f0"
     ptp4lOpts: "-2 --summary_interval 0"
     phc2sysOpts: "-a -r -r -n 24"
     ptp4lConf: |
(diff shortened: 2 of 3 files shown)


feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# Check for updated worker-cnf node
+
+oc wait mcp/worker-cnf --for condition=updated --timeout 1s
```
yuvalk, line 5: that assumes the mcp name is worker-cnf.
we are already kustomizing that to something else (worker-du) in some cases.
and changing the script for that would be difficult.
I suggest adding an environment variable

so 
```
oc wait ${CNF_MCP} --for condition ...
```

I'll actually open a separate PR about it, because the fix need to be cross the whole repo

feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
```diff
@@ -0,0 +1,14 @@
...
+
+
+# Check for real-time kernel.
+
+WORKER_NODE=$(oc get nodes --selector="node-role.kubernetes.io/worker-cnf" --output="name")
+
+echo "worker-cnf node is: ${WORKER_NODE}"
```
yuvalk, line 12: echoing in a `is_ready.sh` script will spam the log.
please remove it

yuvalk, line 12: there might be more than one worker in that pool

feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
```diff
@@ -0,0 +1,14 @@
...
+
+oc describe "${WORKER_NODE}" | grep "Kernel Version:.*\.rt[0-9]*\..*\.x86_64"
```
yuvalk, line 14: I think that checking just the kernel is not enough.
IIRC - there are 2 reboots
one for PAO 
and another for SRIOV

not sure if we even control the order
but need to check for both

you could actually search for the vfs in the same describe command output

juphoff: According to Vitaly, the necessary reboots are done when the worker reports 'updated', which happens before this. I verification of the RT kernel as an additional, post-update sanity check--something that's a necessary but not sufficient indication of a successful update.

yuvalk: sriov daemon can reboot the node on it's own
https://github.com/openshift/sriov-network-operator/blob/master/pkg/daemon/daemon.go#L513

juphoff: But that still wouldn't cause the previous check--that the worker-cnf node has completed its update--to pass, right?

In the QE pipeline, this profile deployment and subsequent verification isn't performed until after all the CNF operators, such as sriov, have been applied and verified.

juphoff: I'm going to move this Kernel-version check out of the is_ready.sh script and into the QE Jenkins pipeline along with other tests that go beyond just ensuring the update is done.

feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
juphoff: I've changed this to use ${CNF_MCP:-"mcp/worker-cnf"}, i.e. to use the environment variable you recommended, with a compatibility default to worker-cnf if unset.

feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
juphoff: Removed.

juphoff: I have removed this check altogether and will move it to the QE pipeline, where it will be accompanied by other checks & tests. It wasn't necessary here.

feature-configs/cn-ran-overlays/ran-profile/is_ready.sh
juphoff: I have removed this check and will do it as part of the QE pipeline's other checks & tests.

yuvalk: yes, but the problem remian.
I verified this.
sriov operator will reboot the node in some cases, which are:
- mlx: every time to change the max number of VFs
- intel: just to enable iommu in the linux cmd for the first time
and we can't know/control the order
so I think, for cn-ran, is_ready,sh should do 2 things:
1. wait for the mcp (as you already did)
2. verify vfs exists in oc describe node

for #2 there, maybe there's a better way? @SchSeba

juphoff: Oh, duh, I missed the vfs part. Thanks. Investigating...

Issue #373, opened 2020-11-27
CNF-Tests: when generating the tests markdown, the order is shuffled every time


fedepaol:
When we generate the doc, the order in which the tests are written in the markdown changes every time.

PR #395, opened 2020-12-22
Fix docgen shuffling order of markdown everytime

oribon:
problem was caused because the sort happened before cleaning the string

https://github.com/openshift-kni/cnf-features-deploy/issues/373


cnf-tests/docgen/cmd/generate.go
```diff
@@ -124,10 +124,11 @@ func descriptionsToList(descriptions map[string]string, matcher func(string) boo
 		if !matcher(n) {
 			continue
 		}
+		n = sanitizeName(n)
```
fedepaol, line 127: Can you use a different name here?

Something like `sanitizedName` := sanitazeName(n)
Reusing the variable may cause a bit of confusion

PR #396, opened 2020-12-22
DPDK: Change the hugepages number from 4 to 5

fedepaol:
We are now running an extra pod and we need to allow pages for that too.


SchSeba (approved):
/lgtm

fedepaol:
overriding.
Failures are related to performance tests, not to this change

PR #397, opened 2020-12-22
Added to ci-job comparison of newly generated docs to existing docs

oribon:
Adds the ability to check if the docs(TESTLIST.md) are up to date with the tests.

Should not be merged before https://github.com/openshift-kni/cnf-features-deploy/pull/395.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ govet:
 	# Disabling GO111MODULE just for go vet execution
 	GO111MODULE=off go vet github.com/openshift-kni/cnf-features-deploy/functests/...
 
-ci-job: gofmt golint govet check-tests-nodesc
+ci-job: gofmt golint govet check-tests-nodesc validate-test-list
 
 feature-deploy:
 	FEATURES_ENVIRONMENT=$(FEATURES_ENVIRONMENT) FEATURES="$(FEATURES)" hack/feature-deploy.sh
@@ -78,3 +78,7 @@ check-tests-nodesc:
 generate-cnf-tests-doc:
 	@echo "Generating cnf tests doc"
 	hack/generate-cnf-docs.sh
+
+validate-test-list:
+	@echo "Comparing newly generated docs to existing docs"
+	hack/compare-gen-md.sh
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -78,3 +78,7 @@ check-tests-nodesc:
 generate-cnf-tests-doc:
 	@echo "Generating cnf tests doc"
 	hack/generate-cnf-docs.sh
+
+compare-gen-md:
+	@echo "Comparing newly generated docs to existing docs"
+	hack/compare-gen-md.sh
```
fedepaol, line 84: nit: new line

Makefile
```diff
@@ -78,3 +78,7 @@ check-tests-nodesc:
 generate-cnf-tests-doc:
 	@echo "Generating cnf tests doc"
 	hack/generate-cnf-docs.sh
+
+compare-gen-md:
```
fedepaol, line 82: this can be "validate-test-list" or something like that

hack/compare-gen-md.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
+
+MDFILE="TESTLIST.md"
+
+mv cnf-tests/${MDFILE} cnf-tests/${MDFILE}.old
+
+. $(dirname "$0")/generate-cnf-docs.sh
```
fedepaol, line 9: Why not generating a new one in a different path (maybe _cache which is in .gitignore)?
This won't mess up with eventual changes, and will avoid us committing the .old file (which I will be guilty for sure :P)

hack/compare-gen-md.sh
```diff
@@ -0,0 +1,11 @@
...
+
+diff cnf-tests/${MDFILE} cnf-tests/${MDFILE}.old > /dev/null 2>&1
```
fedepaol, line 11: nit: new line

fedepaol:
@oribon the "ci" lane failed here, not sure if because of the markdown comparison.
If that is the case, some output would be needed otherwise we would not understand what is going on

fedepaol:
@oribon would it make sense to warn the user and suggest him the command to re-generate the tests doc?

PR #399, opened 2020-12-23
Restructure RAN profile: add CU, extendability

vitus133:
This commit restructures the profile manifests to enable
    creation of additional worker node types, such as DU-LDC
    and CU-UP. Detailed instructions on how this profile can be
    deployed, customized and scaled are provided in the README.
    There are also changes addressing the following:
    - is_ready.sh is provided per unit type
    - the structure is using a single basic profile, adding
    customizations on top of it for each node type
    - support of temporary patches (disable chronyd and
    custom realtime kernel)
    - improved README


vitus133:
@yuvalk @fedepaol @SchSeba 
I'd appreciate your comments
Thanks!

PR #400, opened 2020-12-23
Added vrf plus sriov integration test. Updated vrf test

kononovn:


functests/vrf/vrf.go
```diff
@@ -244,3 +188,63 @@ func paramInParamList(param string, paramRange []string) error {
 	}
 	return fmt.Errorf("error: wrong parameter %v", param)
 }
+
+func defineDesсribeTable(node string, ipStack string, namespace string, apiclient *client.ClientSet, nodeName string, testNetworkBlue string, testNetworkRed string) {
```
fedepaol, line 192: I think testNetworkBlue can be named in a better way.
Like, blueVRFNetworkName or something like that. Also, this is a generic function but you are already assuming the value in the parameter name.
Maybe it's better to switch to firstVRFNetwork, secondVRFNetwork or something along that line?

functests/vrf/vrf.go
```diff
@@ -244,3 +188,63 @@ func paramInParamList(param string, paramRange []string) error {
...
+		Skip("Skipping IPv4 test. Cluster supports IPv6 protocol")
+	} else if ipStack == ipStackIPv4 && net.ParseIP(podClientVRFRedOverlappingIP).To4() != nil {
+		podClientVRFBlueIPAddress = "10.255.255.1"
+		podServerVRFBlueIPAddress = "10.255.255.2"
+	} else {
+		Skip("Unpsupported protocol parameter")
+	}
+
+	By("Create VRFs client/server pods")
+	podClientIpamConfig := fmt.Sprintf(`[{"name": "%s", "mac": "%s", "ips": ["%s/24"]}, {"name": "%s", "mac": "%s", "ips": ["%s/24"]}]`,
```
fedepaol, line 210: this is more than the ipam config, I Think this is the podNetworkDefinition. You can omit pod and just call it clientNetworkDefinition

functests/vrf/vrf_sriov.go
```diff
@@ -0,0 +1,176 @@
...
+		metaPluginRed := fmt.Sprintf(`{"type": "vrf", "vrfname": "%s"}`, VRFRedName)
+		err = createSriovNetwork(sriovclient, sriovDevice, testNetworkRed, sriovNamespaces.Test, namespaces.SRIOVOperator, resourceNameVrf, ipam, metaPluginRed)
+		Expect(err).ToNot(HaveOccurred())
+		Eventually(func() error {
+			netAttDef := &netattdefv1.NetworkAttachmentDefinition{}
+			return sriovclient.Get(context.Background(), runtimeclient.ObjectKey{Name: testNetworkRed, Namespace: sriovNamespaces.Test}, netAttDef)
+		}, 60*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
+
+		metaPluginBlue := fmt.Sprintf(`{"type": "vrf", "vrfname": "%s"}`, VRFBlueName)
+		err = createSriovNetwork(sriovclient, sriovDevice, testNetworkBlue, sriovNamespaces.Test, namespaces.SRIOVOperator, resourceNameVrf, ipam, metaPluginBlue)
```
fedepaol, line 86: You can leverage the CreateSriovNetwork from the sriov package. It takes an optional slice of functions for changing the network before creating it. This avoids the need for another function:

func CreateSriovNetwork(clientSet *testclient.ClientSet, intf *sriovv1.InterfaceExt, name string, namespace string, operatorNamespace string, resourceName string, ipam string, options ...SriovNetworkOptions) error {

options is a type SriovNetworkOptions func(*sriovv1.SriovNetwork)
so you can do:
sriovOptions := func(network *sriovv1.SriovNetwork) {
    network.MetaPlugins = blablabla
}

and pass it as the last argument to the above function

functests/vrf/vrf_sriov.go
```diff
@@ -0,0 +1,176 @@
...
+		DescribeTable("Integration: SRIOV, IPAM: static, Interfaces: 1, Scheme: 2 Pods 2 VRFs OCP Primary network overlap",
+			func(node string, ipStack string) {
+				defineDesсribeTable(node, ipStack, sriovNamespaces.Test, apiclient, nodesList[0], testNetworkBlue, testNetworkRed)
+			},
+			Entry(describe, sameNode, ipStackIPv4),
+		)
+	})
+})
+
+func createSriovPolicy(sriovclient *sriovtestclient.ClientSet, sriovDevice *sriovv1.InterfaceExt, testNode string, numVfs int, resourceName string) {
```
fedepaol, line 114: this is the dpdk implementation.
In case of intel it will create a vfio one, which is not good for what we want to do.
Can you just use the one exported by the sriov operator tests?

functests/vrf/vrf.go
kononovn: I switched most of the var names to *firstVRF*/*secondVRF*, please review it again

PR #401, opened 2020-12-24
Ptp slavery

yuvalk:
eradicate slavery from ptp


vitus133:
@yuvalk 
I think it's now called "subordinate" and not "client"
The "mastery" still remains

yuvalk:
> @yuvalk
> I think it's now called "subordinate" and not "client"
> The "mastery" still remains

https://sourceforge.net/p/linuxptp/mailman/message/37183551/

yuvalk:
assuming we want to wait with this till patches are accepted in linuxptp
and actually until they are accepted into 
I'll close and reopen if/when appropriate

Issue #372, opened 2020-11-27
CNF tests: when looking for a valid mcp for sctp, the tests only check MatchExpressions and not MatchLabels


fedepaol:
If there's a machine config that enables sctp, but it's selected by a machine config pool via "matchexpressions", then the discovery part of the tests do not find it and the test fails.

PR #403, opened 2020-12-27
Find SCTP MachineConfigPool also via MatchLabels

oribon:
https://github.com/openshift-kni/cnf-features-deploy/issues/372

### DIFF
--- a/functests/sctp/sctp.go
+++ b/functests/sctp/sctp.go
@@ -598,6 +598,14 @@ func findSCTPNodeSelectorByMCLabel(mcLabel string) (string, error) {
 				}
 			}
 		}
+		for _, v := range mcp.Spec.MachineConfigSelector.MatchLabels {
+			if v == mcLabel {
+				for key, label := range mcp.Spec.NodeSelector.MatchLabels {
+					newSCTPNodeSelector := key + "=" + label
+					return newSCTPNodeSelector, nil
+				}
+			}
+		}
 	}
 
 	return "", errors.New("Cannot find SCTPNodeSelector")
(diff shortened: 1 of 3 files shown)


yuvalk:
@fedepaol 
maybe it's a better idea altogether to start a job/pod and search for a node with the module?
this way is more generic and can fit outside OCP as well

SchSeba:
/lgtm

holding until we branch this is for an SNO.
/hold

fedepaol:
> @fedepaol
> maybe it's a better idea altogether to start a job/pod and search for a node with the module?
> this way is more generic and can fit outside OCP as well

Agree for outside OCP, but as a short term solution I'd leave as it is (and address all / as much dependencies from MCP all together). If we only change this bit, there will be still other places around that look for MCs.
Also, start a pod may not scale if the cluster has 10000 nodes and only one has the module loaded (while this gives you back a selector).

yuvalk:
> > @fedepaol
> > maybe it's a better idea altogether to start a job/pod and search for a node with the module?
> > this way is more generic and can fit outside OCP as well
> 
> Agree for outside OCP, but as a short term solution I'd leave as it is (and address all / as much dependencies from MCP all together). If we only change this bit, there will be still other places around that look for MCs.
> Also, start a pod may not scale if the cluster has 10000 nodes and only one has the module loaded (while this gives you back a selector).

the other, longer term, option is to enhance ptp-operator to report ptp functionality as a node resource

fedepaol:
Just thinking, do we need this also for the ux_32 module? I think it suffers from the same issue

PR #408, opened 2021-01-07
deploy: always use the documented name

fromanirh:
per https://docs.openshift.com/container-platform/4.6/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html#installing-the-performance-addon-operator_cnf-master

the name we should use is 'openshift-performance-addon-operator'


fromanirh:
/hold
let's evaluate impacts (if any) on cnf-tests

fromanirh:
we will reopen if/when needed

PR #414, opened 2021-01-13
Fix bug in cluster validation findMachineConfigPool function

sabinaaledort:


yuvalk:
@fedepaol do we need a bz to backport to 4.7?

yuvalk:
I think this also calls for running cnf-tests, when testing cnf-tests, in a cluster with multiple MCPs
we might need to think of other cluster resources we should create, to validate better that cnf-tests is working properly

fedepaol:
thanks for finding (& fixing) this!
/approve

fedepaol:
> @fedepaol do we need a bz to backport to 4.7?

Yes, was thinking the same, and potentially back to 4.6 (but for that I'd maybe wait if this is experienced by somebody, the workaround is to skip the validation step).

yuvalk:
> > @fedepaol do we need a bz to backport to 4.7?
> 
> Yes, was thinking the same, and potentially back to 4.6 (but for that I'd maybe wait if this is experienced by somebody, the workaround is to skip the validation step).

I think anyone running this on a 'real' cluster, would most probably have multiple MCPs and then be affected by that.
since the RAN profile is expected to be used on 4.6 as well, I'd say 4.6 is imminent. 
so maybe smarter to get ahead of it actually happening to someone?

fedepaol:
> > > @fedepaol do we need a bz to backport to 4.7?
> > 
> > 
> > Yes, was thinking the same, and potentially back to 4.6 (but for that I'd maybe wait if this is experienced by somebody, the workaround is to skip the validation step).
> 
> I think anyone running this on a 'real' cluster, would most probably have multiple MCPs and then be affected by that.
> since the RAN profile is expected to be used on 4.6 as well, I'd say 4.6 is imminent.
> so maybe smarter to get ahead of it actually happening to someone?

Agree.

PR #419, opened 2021-01-19
Makefile rule: deploy and validate only

fedepaol:
The reference layers (i.e. the ran overlays) require capabilities not available on cloud providers. However, deploying the manifests without errors and checking that the required operators are running has validation value. With this, we add a new validate-on-ci rule that runs only the validation suite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 #TODO add default features here
 export FEATURES?=sctp performance xt_u32
+export SKIP_TESTS?=
 IMAGE_BUILD_CMD ?= "docker"
 
 # The environment represents the kustomize patches to apply when deploying the features
@@ -32,13 +33,21 @@ deps-update:
 
 functests: 
 	@echo "Running Functional Tests"
-	FEATURES="$(FEATURES)" hack/run-functests.sh
+	SKIP_TESTS="$(SKIP_TESTS)" FEATURES="$(FEATURES)" hack/run-functests.sh
+
+#validate is intended to validate the deployment as a whole, not focusing
+# but eventually skipping
+validateonly: 
+	@echo "Validating only"
+	SKIP_TESTS="$(SKIP_TESTS)" DONT_FOCUS=true TEST_SUITES="validationsuite" hack/run-functests.sh
 
 functests-on-ci: setup-test-cluster feature-deploy feature-wait functests
 
 origin-tests:
 	@echo "Running origin-tests"
 	ORIGIN_TESTS_FILTER="$(ORIGIN_TESTS_FILTER)" hack/run-origin-tests.sh
+
+validate-on-ci: setup-test-cluster feature-deploy feature-wait validateonly
 
 gofmt:
 	@echo "Running gofmt"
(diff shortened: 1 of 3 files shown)


fedepaol:
/hold
need to validate

yuvalk:
> /hold
> need to validate

need to validate the validator :rofl:

Makefile
```diff
@@ -32,10 +33,18 @@ deps-update:
...
 functests: 
 	@echo "Running Functional Tests"
-	FEATURES="$(FEATURES)" hack/run-functests.sh
+	SKIP_TESTS="$(SKIP_TESTS)" FEATURES="$(FEATURES)" hack/run-functests.sh
+
+#validate is intended to validate the deployment as a whole, not focusing
+# but eventually skipping
+validateonly: 
+	@echo "Validating only"
+	SKIP_TESTS="$(SKIP_TESTS)" DONT_FOCUS=true TEST_SUITES="validationsuite" hack/run-functests.sh
```
yuvalk, line 42: That means I can't validate on partial feature set?

hack/run-functests.sh
```diff
@@ -22,14 +22,24 @@ export CLEAN_PERFORMANCE_PROFILE="false"
...
 export LATENCY_TEST_RUN=${LATENCY_TEST_RUN:-false}
 
-if [ "$FEATURES" == "" ]; then
-	echo "[ERROR]: No FEATURES provided"
-	exit 1
+echo "Running local tests"
+
+
+if [ "$DONT_FOCUS" == true ]; then
+	echo "DONT_FOCUS enabled, not focusing"
```
yuvalk, line 29: meaning it will run all the tests?
let's be explicit about it
so ppl dont have to go hunt for ginkgo documentation on how the tool behave

Makefile
fedepaol: Yup. As the comment says, the idea when validating a "profile" is that you want to test all the features and in case you skip some (because the number of skips is expected to be very limited).

I felt this was a good compromise with the current FEATURES usage.

hack/run-functests.sh
fedepaol: Yep, make sense.
Will change it

hack/run-functests.sh
```diff
@@ -22,14 +22,24 @@ export CLEAN_PERFORMANCE_PROFILE="false"
 # Latency tests env variables
 export LATENCY_TEST_RUN=${LATENCY_TEST_RUN:-false}
 
-if [ "$FEATURES" == "" ]; then
-	echo "[ERROR]: No FEATURES provided"
-	exit 1
+echo "Running local tests"
+
+
+if [ "$DONT_FOCUS" == true ]; then
```
SchSeba, line 28: should we put a default in this file for the `$DONT_FOCUS` variable?

fedepaol: if the variable is not defined, the condition above will be false anyway.
If you think it would be more readable, I'll add it

hack/run-functests.sh
fedepaol: Done

SchSeba (approved):
/lgtm

PR #420, opened 2021-01-20
Add origin-tests

sabinaaledort:
ose-tests verify the behavior of an OCP cluster by running remote tests against the
cluster API that exercise functionality.


yuvalk:
Sabina, see:
https://github.com/openshift-kni/cnf-features-deploy/blob/f49d21e8fadcb78d6fc739e9daaa61f7003787df/external-tests/origintests/test.sh#L11
which is a script to extract the ose_tests as a binary.
this is done to allow us to run ose-tests in u/s ci, which runs inside a pod (so cant use podman there)

and also:
https://github.com/openshift-kni/cnf-features-deploy/blob/a3a571509e1fb1a46d272ccccb57d27708e2f868/hack/run-functests.sh#L79
which is an option to call this ose-tests script from run-functests

Now, IMHO, your approach is cleaner,
but let's work on converging it all into a single place.
ie - add an option into your script to do that extraction.
then I think we're ok with cleaning up the old method.

fedepaol:
> Sabina, see:
> 
> https://github.com/openshift-kni/cnf-features-deploy/blob/f49d21e8fadcb78d6fc739e9daaa61f7003787df/external-tests/origintests/test.sh#L11
> 
> which is a script to extract the ose_tests as a binary.
> this is done to allow us to run ose-tests in u/s ci, which runs inside a pod (so cant use podman there)
> 
> and also:
> 
> https://github.com/openshift-kni/cnf-features-deploy/blob/a3a571509e1fb1a46d272ccccb57d27708e2f868/hack/run-functests.sh#L79
> 
> which is an option to call this ose-tests script from run-functests
> 
> Now, IMHO, your approach is cleaner,
> but let's work on converging it all into a single place.
> ie - add an option into your script to do that extraction.
> then I think we're ok with cleaning up the old method.

agree

hack/run-ose-tests.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
+
+export CONTAINER_MGMT_CLI="${CONTAINER_MGMT_CLI:-docker}"
+export TESTS_REPORTS_PATH="${TESTS_REPORTS_PATH:-/tmp/artifacts/}"
+export TESTS_TYPE="${TESTS_TYPE:-openshift/conformance/parallel}"
+export failed=false
```
yuvalk, line 8: why is this exported?
I think you can remove the export

hack/run-ose-tests.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
+
+export CONTAINER_MGMT_CLI="${CONTAINER_MGMT_CLI:-docker}"
+export TESTS_REPORTS_PATH="${TESTS_REPORTS_PATH:-/tmp/artifacts/}"
+export TESTS_TYPE="${TESTS_TYPE:-openshift/conformance/parallel}"
```
yuvalk, line 7: -> TESTS_FILTER maybe?

fedepaol: or, OSE_TESTS_FILTER

Makefile
```diff
@@ -36,6 +36,10 @@ functests:
 
 functests-on-ci: setup-test-cluster feature-deploy feature-wait functests
 
+origin-tests:
+	@echo "Running origin-tests"
+	TESTS_TYPE="$(TESTS_TYPE)" hack/run-origin-tests.sh
```
yuvalk, line 41: TESTS_TYPE is never used
I think you forgot to rename it to `ORIGIN_TESTS_FILTER`

README.md
```diff
@@ -46,6 +46,17 @@ We expect to have a section of [the test suite](functests/test_suite_test.go) na
...
 
+### origin-tests
+
+Verifies behavior of an OCP cluster by running remote tests against the cluster API that exercise functionality.
+These tests may be disruptive.
+
+Running a dockerized version of origin-tests from [quay.io/openshift/origin-tests](quay.io/openshift/origin-tests).
+The full test suite can be found at [https://github.com/openshift/openshift-tests](https://github.com/openshift/openshift-tests).
+
+- run `TESTS_TYPE=openshift/conformance/serial make origin-tests`.
```
yuvalk, line 57: TESTS_TYPE -> ORIGIN_TESTS_FILTER

hack/run-origin-tests.sh
```diff
@@ -0,0 +1,82 @@
...
+failed=false
+
+echo "Running $ORIGIN_TESTS_FILTER tests"
+
+mkdir -p "$TESTS_REPORTS_PATH"
+
+mkdir -p _cache/
+cp -f "$KUBECONFIG" _cache/kubeconfig
+
+# The cluster infrastructure provider, will automatically default to the correct value.
```
yuvalk, line 48: what's a provider?

hack/run-origin-tests.sh
```diff
@@ -0,0 +1,82 @@
...
+if [ -n "$TEST_PROVIDER" ]; then
+  echo "Provider: $TEST_PROVIDER"
+  provider="--provider "${TEST_PROVIDER}""
+fi
+
+if [ "$ORIGIN_TESTS_IN_CONTAINER" == "true" ]; then
```
yuvalk, line 54: I think we should default to running in pdoman container

hack/run-origin-tests.sh
```diff
@@ -0,0 +1,82 @@
...
+            rm -rf ${skopeo_workdir}
+            break
+        fi
+    done
+}
+
+CONTAINER_MGMT_CLI="${CONTAINER_MGMT_CLI:-docker}"
+TESTS_REPORTS_PATH="${TESTS_REPORTS_PATH:-/tmp/artifacts/}"
+
+ORIGIN_TESTS_IMAGE="${ORIGIN_TESTS_IMAGE:-quay.io/openshift/origin-tests:4.8}"
```
yuvalk, line 35: can we take the version from common.sh?
I think it worthwhile changing this across the board (where possible)

yuvalk: it'll make it (much) easier and less error prone to update the version

hack/run-origin-tests.sh
```diff
@@ -0,0 +1,82 @@
...
+            mv "$testsbin" _cache/tools/openshift-tests
+            chmod +x _cache/tools/openshift-tests
+            rm -rf ${skopeo_workdir}
+            break
+        fi
+    done
+}
+
+CONTAINER_MGMT_CLI="${CONTAINER_MGMT_CLI:-docker}"
+TESTS_REPORTS_PATH="${TESTS_REPORTS_PATH:-/tmp/artifacts/}"
```
yuvalk, line 33: this will not run over cnf-tests?

sabinaaledort: no, the directory is created with '-p' flag and we just get one junit file in there. for example, 'junit_e2e_20210124-133651.xml'

hack/run-ose-tests.sh
yuvalk: they call it now origin-tests
because it's no longer ose..

hack/run-origin-tests.sh
sabinaaledort: do you mean to add ORIGIN_TESTS_IMAGE to common.sh?

hack/run-origin-tests.sh
sabinaaledort: the cluster type, i will update the comment

hack/run-origin-tests.sh
yuvalk: no, just something like
`OCP_VERSION=4.8`
and then use it here and everywhere else that need a version

fedepaol: Great idea!
I'd leave this for a separate PR though, wdyt?
Also, there are a couple of places where it would not immediate to use the env variable (although we could set it in the image). 

A couple that come into my mind are: 
- the json containing the mirroring images
- the default tag used by the tests in go code

I'd track this somewhere

yuvalk: let's introduce the variable here
and track the work to convert other places in a different PR.
but at least we'll have the system to do that inplace.

fedepaol: agree

hack/run-origin-tests.sh
```diff
@@ -0,0 +1,84 @@
...
+
+. $(dirname "$0")/common.sh
+
+skopeo_workdir=$(mktemp -d)
+function finish_skopeo {
+  rm -rf ${skopeo_workdir}
+}
+trap finish_skopeo EXIT
+
+function get_ose_tests_binary {
```
yuvalk, line 11: get_ose_tests_binary -> get_openshift_tests

PR #422, opened 2021-01-21
Leverage the validate test suite when waiting

fedepaol:
depends on https://github.com/openshift-kni/cnf-features-deploy/pull/419

Currently we use a is_ready.sh file per feature. This tampers the  "kustomize only" spirit of the deployments, plus makes things tricky when dealing with all-in-one deployments such as ran profiles.
    
Here we change the waiting script to leverage the validation step of cnf-tests. On top of that, we also extend the validation suite to check if a tag for the dpdk imagestream is present, and to verify that the sriov status is not in progress.

In a separate commit we also add an extra layer for validating the ran profile on cloud environments


hack/run-functests.sh
```diff
@@ -58,10 +67,10 @@ if [ "$TESTS_IN_CONTAINER" == "true" ]; then
   -v $(pwd)/_cache/:/kubeconfig:Z \
   -v $TESTS_REPORTS_PATH:/reports:Z \
   ${env_vars} \
```
SchSeba, line 69: should you add the `TEST_SUITES` also here?

fedepaol: it won't hurt.
I expect the three suites to run consistently when somebody runs `make functests`, but I can add that in case it's useful in the future (plus, we would have a consistent api outside)

fedepaol: Just noticed I mis-understood you comment. Yes, you are right.
Just fixed it

SchSeba (approved):
/lgtm
/approve

PR #426, opened 2021-01-31
origin-tests in a disconnected environment

sabinaaledort:


Makefile
```diff
@@ -40,6 +40,10 @@ origin-tests:
 	@echo "Running origin-tests"
 	ORIGIN_TESTS_FILTER="$(ORIGIN_TESTS_FILTER)" hack/run-origin-tests.sh
 
+mirror-origin-tests:
```
yuvalk, line 43: can we also have a target that does both?
mirror, then run tests in disocnnected mode from that repo

hack/mirror-origin-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+
+skopeo_workdir=$(mktemp -d)
+mapping_file=$(mktemp)
+function finish {
+  rm -rf ${skopeo_workdir}
+  rm -f ${mapping_file}
+}
+trap finish EXIT
+
+MIRROR_ORIGIN_TESTS_OC_TOOL="${MIRROR_ORIGIN_TESTS_OC_TOOL:-oc}"
```
yuvalk, line 13: can this be a 'global' common in common.sh?
I think it is useful in many scripts.

hack/mirror-origin-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+MIRROR_ORIGIN_TESTS_CONTAINER_MGMT_CLI="${MIRROR_ORIGIN_TESTS_CONTAINER_MGMT_CLI:-docker}"
```
yuvalk, line 14: same goes for this one. it's illogical to set docker for one script and podman (or other) for others.

hack/mirror-origin-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+
+MIRROR_ORIGIN_TESTS_IN_CONTAINER="${MIRROR_ORIGIN_TESTS_IN_CONTAINER:-true}"
```
yuvalk, line 16: even this one can be made more global,
setting if we want to run things in containers or not

hack/mirror-origin-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+MIRROR_ORIGIN_TESTS_IMAGE="${MIRROR_ORIGIN_TESTS_IMAGE:-quay.io/openshift/origin-tests:$OCP_VERSION}"
```
yuvalk, line 17: this one can be shared between the run and mirror
so they both use the same parameter name.

hack/mirror-origin-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+      get_openshift_tests_binary $MIRROR_ORIGIN_TESTS_IMAGE ${skopeo_workdir}
+  fi
+
+  mkdir -p _cache/
+
+  ORIGIN_TESTS_CMD="_cache/tools/openshift-tests"
+fi
+
+$ORIGIN_TESTS_CMD images --to-repository $ORIGIN_TESTS_REPOSITORY > ${mapping_file}
+if [[ $? != 0 ]]; then
```
yuvalk, line 51: note that you are running with `-e` script will stop before this command

hack/mirror-origin-tests.sh
yuvalk, line 13: wait actually, I think this is the first time we ever introduce a dependency on OC into cnf-features-deploy
do we want that?
I understand this is just in order to do the actual mirroring

maybe we should add a test if it doesnt exist
or we can even help people get it from:
https://github.com/openshift/okd/releases/latest
I'd at least suggest that in the error msg if it doesnt exists

unfortunately RH doesnt make it easy enough to automate the d/l of the latest oc

sabinaaledort: we are using oc in the setup-test-cluster.sh script also:
https://github.com/openshift-kni/cnf-features-deploy/blob/5f35374d8a50fc3db838ac5f55a9f612698a7607/hack/setup-test-cluster.sh#L7

yanirq:
Can we have a description in the PR ?

yuvalk (changes requested):
> Use global environment variables from common.sh script 

That's a WHAT msg and not a WHY

Makefile
```diff
@@ -46,6 +46,10 @@ wait-and-validate:
 functests-on-ci: setup-test-cluster feature-deploy feature-wait functests
 
 origin-tests:
+	@echo "Running origin-tests"
+	TESTS_IN_CONTAINER=true ORIGIN_TESTS_FILTER="$(ORIGIN_TESTS_FILTER)" hack/run-origin-tests.sh
+
+skopeo-origin-tests:
```
yuvalk, line 52: seems like you are adding a makefile target in this commit

hack/common.sh
```diff
@@ -19,3 +19,8 @@ export OCP_VERSION="${OCP_VERSION:-4.8}"
 
 export TESTS_REPORTS_PATH="${TESTS_REPORTS_PATH:-/tmp/artifacts/}"
 
+# expect oc to be in PATH by default
+export OC_TOOL="${OC_TOOL:-oc}"
+
+export CONTAINER_MGMT_CLI="${CONTAINER_MGMT_CLI:-docker}"
```
yuvalk, line 25: @fedepaol 
probably about time to change default to podman :-)

hack/origin-tests-common.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
```
yuvalk, line 3: I think it's better to keep the sourcing of common explicit in each script
it also allow for the hypothetical use of the function here, without the common.sh if ever needed

sabinaaledort: but we need OCP_VERSION for ORIGIN_TESTS_IMAGE (line 28) which is set in common.sh

Makefile
sabinaaledort: i just added TESTS_IN_CONTAINER=true because we changed the default to false in the common.sh

yuvalk:
so the only thing I'm thinking - is maybe we need to add a CI lane that test this?

Makefile
fedepaol: Isn't this default different from the cnf-tests one? If so, can we stick with the same default (not in container)?
Switching everything to "IN_CONTAINER" is fine as well

hack/common.sh
fedepaol: Lol I was young(er) :P

Makefile
yuvalk: skopeo-origin-tests run like cnf-tests (I guess you mean the `functests` target ?)
the origin-tests are just to be handy for devs, which I guess will prefer to run in container

you prefer to reverse it? have `origin-tests` run skopeo and add `origin-tests-podman` to run container?

we cant change `functests` otherwise we'll need to introduce something like `functests-on-ci` to use skopeo (bc CI is running in pod, and it is difficult to run podman in a pod)

fedepaol: So, the default for cnf-tests is run as a binary.
I'd like the default to be consistent (because consistency is good :-) ), so either both sticking to IN_CONTAINER=false or switching the default IN_CONTAINER to true. If we do the latter, we need to change how u/s CI is called.

yuvalk: see that we have two targets, one with containers and the other with the binary

fedepaol: Sorry, I read in a rush.
Works for me

PR #428, opened 2021-02-01
Ptp socket test

oribon:
Add a test to check if pods on the ptp slave node can communicate with the ptp UDS socket https://github.com/openshift/ptp-operator/pull/91

Depends on https://github.com/openshift-kni/cnf-features-deploy/pull/427 https://github.com/openshift-kni/cnf-features-deploy/pull/424


cnf-tests/testsuites/e2esuite/ptp/ptp.go
```diff
@@ -137,6 +138,79 @@ var _ = Describe("ptp", func() {
...
+			}
+		})
+
+		AfterEach(func() {
+			err := namespaces.Clean(ptpLinuxDaemonNamespace, "testpod-", client.Client)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		var _ = Context("Run pmc in a new pod on the slave node", func() {
+			It("Create the pod on the slave node", func() {
```
marzianor9, line 163: It() should contain what It(the component tested) should do or is expected to do.
to log steps you can use the By() function.

example:
It("should be able to communicate using a uds",func(){
    By("Creating a pod")
    /////////
    By("Executing pmc")
    /////////
})

cnf-tests/testsuites/pkg/pods/pods.go
```diff
@@ -76,6 +76,14 @@ func RedefinePodWithNetwork(pod *corev1.Pod, networksSpec string) *corev1.Pod {
 	return pod
 }
 
+// RedefineAsPrivileged updates the pod definition to be privileged
+func RedefineAsPrivileged(pod *corev1.Pod) *corev1.Pod {
+	pod.Spec.Containers[0].SecurityContext = &corev1.SecurityContext{}
+	b := true
+	pod.Spec.Containers[0].SecurityContext.Privileged = &b
```
marzianor9, line 83: doesn't it make only the first container privileged, instead of the whole pod?

oribon: It does. Like some other Redefine/Define functions in the file it changes only the first container so I went with it. Also from what I see there isn't a `privileged` parameter for a Pod, and I'm not sure updating all of the pod's containers is necessary here. btw I took it from here https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/test/util/pod/pod.go#L56

cnf-tests/testsuites/e2esuite/ptp/ptp.go
oribon: thanks updated :)

cnf-tests/testsuites/e2esuite/ptp/ptp.go
```diff
@@ -137,6 +138,82 @@ var _ = Describe("ptp", func() {
...
+				ptpPods, err := client.Client.Pods(ptpLinuxDaemonNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app=linuxptp-daemon"})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(len(ptpPods.Items)).To(BeNumerically(">", 0))
+				var ptpSlavePod v1core.Pod
+				for _, pod := range ptpPods.Items {
+					if podRole(pod, slaveLabel) {
+						ptpSlavePod = pod
+					}
+				}
+				Expect(ptpSlavePod).ToNot(BeZero())
```
SchSeba, line 173: zero? this is a pod object. can you have a boolean?

cnf-tests/testsuites/e2esuite/ptp/ptp.go
```diff
@@ -137,6 +138,82 @@ var _ = Describe("ptp", func() {
...
+					},
+				}
+				pod, err := client.Client.Pods(ptpLinuxDaemonNamespace).Create(context.Background(), podDefinition, metav1.CreateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				err = pods.WaitForCondition(client.Client, pod, v1core.ContainersReady, v1core.ConditionTrue, 3*time.Minute)
+				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() int {
+					buf, _ := pods.ExecCommand(client.Client, *pod, []string{"pmc", "-u", "-f", "/var/run/ptp4l.0.config", "GET CURRENT_DATA_SET"})
+					Expect(buf.String()).ToNot(ContainSubstring("failed to open configuration file"), "ptp config file is not shared between pods")
+					return strings.Count(buf.String(), "offsetFromMaster")
```
SchSeba, line 209: please add a comment showing the output as an example for documentation.
also explain why we need to see it 2 times

cnf-tests/testsuites/pkg/pods/pods.go
SchSeba, line 83: you can use the pointer package from k8s

cnf-tests/testsuites/pkg/pods/pods.go
```diff
@@ -76,6 +76,14 @@ func RedefinePodWithNetwork(pod *corev1.Pod, networksSpec string) *corev1.Pod {
 	return pod
 }
 
+// RedefineAsPrivileged updates the pod definition to be privileged
+func RedefineAsPrivileged(pod *corev1.Pod) *corev1.Pod {
+	pod.Spec.Containers[0].SecurityContext = &corev1.SecurityContext{}
```
SchSeba, line 81: this will override the securityContext please check before that if it's already there just add the true

cnf-tests/testsuites/pkg/pods/pods.go
```diff
@@ -76,6 +76,14 @@ func RedefinePodWithNetwork(pod *corev1.Pod, networksSpec string) *corev1.Pod {
 	return pod
 }
 
+// RedefineAsPrivileged updates the pod definition to be privileged
+func RedefineAsPrivileged(pod *corev1.Pod) *corev1.Pod {
```
SchSeba, line 80: extend this to also have container name and if it's empty it apply to the first container inside the pod

SchSeba (commented):
this looks great just a few comments

cnf-tests/testsuites/pkg/pods/pods.go
```diff
@@ -76,6 +91,20 @@ func RedefinePodWithNetwork(pod *corev1.Pod, networksSpec string) *corev1.Pod {
 	return pod
 }
 
+// RedefineAsPrivileged updates the pod definition to be privileged
+func RedefineAsPrivileged(pod *corev1.Pod, containerName string) (*corev1.Pod, error) {
+	c := containerByName(pod, containerName)
+	if c == nil {
+		return pod, goerr.New(fmt.Sprintf("Container with name: %s not found in pod.", containerName))
```
SchSeba, line 98: please use fmt.Error not need to import a new package

PR #435, opened 2021-02-10
Add a copy of the container-mount-hiding proof-of-concept

lack:


yuvalk:
I'm very interested to see how CI react to that
@lack can you add your feature here
https://github.com/openshift-kni/cnf-features-deploy/blob/master/Makefile#L2
so it will actually be tested with?

lack:
@yuvalk : There we go!  After a bunch of misfires (and one actual issue with my previous implementation and OVN), this passes all our CI :)

Makefile
```diff
@@ -1,5 +1,5 @@
 #TODO add default features here
-export FEATURES?=sctp performance xt_u32
+export FEATURES?=sctp performance xt_u32 cmns
```
yuvalk, line 2: I understand why you added that here
but probably need to remove from default before merging

yuvalk:
@lack can you add description into the commit msg itself as of why are we adding it, benefits, etc

yuvalk:
@lack and also change the name of the feature to something a little bit more descriptive

lack:
Note: The current commit history has the `container-mount-namespace` feature enabled by default in the top-level Makefile so the CI infrastructure takes it through its paces.  I plan on removing this from the default feature list prior to merge here, so please do not merge yet.

cnf-tests/testsuites/validationsuite/cluster/validation.go
```diff
@@ -195,85 +195,25 @@ var _ = Describe("validation", func() {
...
-						for _, value := range mcpItem.Spec.MachineConfigSelector.MatchLabels {
-							if value == name {
-								return true, &mcpItem
-							}
-						}
+		matchSCTPMachineConfig := func(ignitionConfig *igntypes.Config, _ *clientmachineconfigv1.MachineConfig) bool {
+			if ignitionConfig.Storage.Files != nil {
+				for _, file := range ignitionConfig.Storage.Files {
+					if file.Path == "/etc/modprobe.d/sctp-blacklist.conf" {
+						return true
```
lack, line 202: This will look a lot cleaner after PR #454 is merged.

hack/common.sh
```diff
@@ -2,14 +2,18 @@
...
 cd "$(dirname "$0")/.."
 
 function finish {
-    popd
+    popd >&2
 }
 trap finish EXIT
 
+function debug {
+    echo "$@" >&2
```
yuvalk, line 14: can you do `echo "DEBUG: @0"`
so we can easily filter and distinguish between debug and errors?

lack:
@yuvalk : All CI passing now, and I've removed the new feature from the Makefile for now, too.

Anything else we need here?

yuvalk:
I think we'd better enable it in the default `FEATURES` string
so it'll be tested for every PR

lack:
Another CI failure related to GCP leases
/test e2e-gcp-ovn

PR #440, opened 2021-02-11
Skip ptp tests setup in single node environment

mmirecki:


fedepaol:
@mmirecki can you expand the commit message with the reason of the skipping?
Also, the ci lane is failing, I checked and I think the reason is because you are accessing the api in init().
The order of various `init` is not guaranteed, so the client might not be ready. On top of that, this would make dryRun fail.
Can you move it to beforeAll ?

PR #442, opened 2021-02-16
Remove node-selector network-sriov.capable: "true" for ran profile

josephdrichard:


fedepaol:
/lgmt
/approve

With this, we "try" to configure all the worker-cnf nodes, but the scheduling will make sure the workloads will run only on the nodes where the device is really available.

fedepaol:
/hold

On a second thought, @josephdrichard can you elaborate the commit message?
Putting the reason why the line is not needed anymore, and what was wrong with that, and why it will work even without that.

PR #448, opened 2021-02-24
Kubernetes reporter: dump also the init containers

fedepaol:
Currently we dump logs only iterating the containers in a pod. Here we
also take the logs from init containers.


SchSeba (commented):
/lgtm
/approve

PR #452, opened 2021-03-01
gatekeeper operator catalog deployment

marzianor9:


SchSeba:
/hold
until validation

SchSeba (commented):
@marzianor9 just a general question there is a reason you check the pods and not just the deployment obect the replica == ready ?

marzianor9:
@SchSeba i wanted to check on the deployment, but to my understanding "readyReplicas" on deployment does not equal running pods, if it does i would be happy to change that.

SchSeba (approved):
/lgtm
/approve

feature-configs/deploy/gatekeeper/subscription.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: gatekeeper-operator-product
+  namespace: openshift-operators
```
SchSeba, line 5: @mmirecki why we use this namespace?

mmirecki: @SchSeba this is the default ns we install to when installing from operatorhub

PR #454, opened 2021-03-04
Refactor some common logic in the cluster validation suite

lack:
The mechanisms for checking MC and MCP was duplicated for both the sctp
and xt_u32 tests, and will be reused for other MC-related deployments in
the future.  This abstraction makes it easier to manage and reuse these
checks.

### DIFF
--- a/cnf-tests/testsuites/validationsuite/cluster/validation.go
+++ b/cnf-tests/testsuites/validationsuite/cluster/validation.go
@@ -195,85 +195,25 @@ var _ = Describe("validation", func() {
 	})
 
 	Context("sctp", func() {
-		findSCTPMachineConfig := func(mcl []clientmachineconfigv1.MachineConfig) (bool, *clientmachineconfigv1.MachineConfig) {
-			for _, mc := range mcl {
-				if mc.Spec.Config.Raw == nil {
-					continue
-				}
-				ignitionConfig := igntypes.Config{}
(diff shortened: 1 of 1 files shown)


cnf-tests/testsuites/validationsuite/cluster/validation.go
```diff
@@ -40,6 +40,77 @@ func init() {
 	machineConfigPoolNodeSelector = fmt.Sprintf("node-role.kubernetes.io/%s", roleWorkerCNF)
 }
 
+type MCMatcher func(*igntypes.Config, *clientmachineconfigv1.MachineConfig) bool
```
fedepaol, line 43: Can you put these utilities functions at the bottom of the file?

fedepaol, line 43: Saw both the matcher rely only on ignitionConfig, I'd remove the second parameter as nobody uses it

fedepaol (commented):
Thanks for doing this!

cnf-tests/testsuites/validationsuite/cluster/validation.go
lack: That's a future-facing hint to the matcher I'm about to add for my other PR, but it's true I could remove this parameter for now and re-add it in that PR.

lack: Done.

SchSeba (commented):
/lgtm
/approve

PR #456, opened 2021-03-09
Allow features to specify additional SKIP_TESTS

lack:
For compound features, such as cn-ran-profiles, the feature needs to
be able to specify if any of the default features aren't on by default
so the default validation suite (which is used for feature-wait) doesn't
attempt to wait for features that aren't present.

In other words, this allows us to stop setting SKIP_TESTS in the CI lane
repository, encapsulating the description of what is deployed (based on
feature-configs/*) and what is tested (based on cnf-tests/* and
test-overrides/*) exclusively in this single repo


yuvalk:
> For compound features, such as cn-ran-profiles, the feature needs to
> be able to specify if any of the default features aren't on by default
> so the default validation suite (which is used for feature-wait) doesn't
> attempt to wait for features that aren't present.

adding a non yaml file in the kustomization directory is awkward and non-intuitive
maybe, a better option for when running in discovery mode, is to allow (via a flag) cnf-tests to read the configuration off a on-cluster ConfigMap

then you can add it to the kustomization directory like any other object and it get stored in the cluster, so customer can repeat cnf-tests in a reproducible manner

(next steps: store the results in cluster, create CRD and build a full operator :-D )

/cc @fedepaol

lack:
The non-yaml file is not in the same hierarchy as the kustomize files, but in a parallel hierarchy.

Though I do agree it is a bit cumbersome.

The problem with using something on-cluster is that the contents is needed as part of determining which tests to use as part of hack/feature-wait.sh. Maybe it would be fine because we allow failures temporarily as we wait for the cluster to get ready, but it feels fragile to rely on a resource from the cluster to determine how to check if the cluster is ready...

That said, maybe the ConfigMap idea is worth trying out to see if it meets the need.

fedepaol:
> adding a non yaml file in the kustomization directory is awkward and non-intuitive

Agree, and it took a bit of effort to get rid of the is_ready.sh

> maybe, a better option for when running in discovery mode, is to allow (via a flag) cnf-tests to read the configuration off a on-cluster ConfigMap
>

Note that this is only the upstream CI, where we run only the validation step because we can't run much more.
 
> then you can add it to the kustomization directory like any other object and it get stored in the cluster, so customer can repeat cnf-tests in a reproducible manner
> 

This could be an option, as long as it's part of EVERY deployment. What I don't want is the configmap to be a part of the CI configuration that is deployed in the cluster, nor for it to be a must have in order to run cnf-tests

> (next steps: store the results in cluster, create CRD and build a full operator :-D )
> 
> /cc @fedepaol

Overall I am a bit torn. The configmap seems a bit too far (ie, it requires a lot of extra logic to parse it and to instruct cnf-tests, and at this point it must end up in the container itself). If this feels too hacky (maybe it only suffers from discoverability?) we can keep the current solution. I kind of like the fact that the CI runner is explicit about what it's doing, but I don't have a strong opinion against this (maybe replacing the file with a json with all the combinations?)

lack:
@fedepaol I'd be happy with a single file that lists the specific overrides instead of a heirarchy of files that contain the same thing.

After looking through the test-running scripts, I was curious if it's worthwhile having the ability to reset or append the following things:
- Tests that should be skipped
- Tests that should be focused
- External tests that should also be run

lack:
@fedepaol 
> It worked so far, and if we don't find any major blockers I'd avoid making the logic more complex.

I totally agree.  Do you prefer my file-per-feature mechanism here, or would you rather a single file with some sort of `featurename: value` mapping?

lack:
@yuvalk : Closing this on your recommendation. We can revisit this (or something else that does the same thing) later, once PR #457 lands.

PR #457, opened 2021-03-10
Create 5G reference network

vitus133:
This commit contains manifests for deploying the 5G RAN components on Kubernetes clusters.
It complements the existing DU profile by extending beyond separate units, approaching closer to the structure maintained by a "typical" mobile network operator.
The manifest structure defined here should become a single source of truth for RAN configurations deployed with Advanced Cluster Management for Kubernetes or additional orchestration tools.


yrobla (commented):
My main concern here is how are we going to reconciliate that with policies. The way how we apply policies is a bit different: we build a bit of a lego... we have base policies, profile policies and per-site policies. And we apply those as separated components... so for example, a kustomization.yaml from a site, that points to base... isn't clearly translated into policies.
What are your thoughts about how to reconcile that?

vitus133:
> My main concern here is how are we going to reconciliate that with policies. The way how we apply policies is a bit different: we build a bit of a lego... we have base policies, profile policies and per-site policies. And we apply those as separated components... so for example, a kustomization.yaml from a site, that points to base... isn't clearly translated into policies.
> What are your thoughts about how to reconcile that?

vitus133:
Hi William,
Thanks for this review!
Please find below my comments.

>  * When describing the DU or CU as real-time or non-real-time, to avoid confusion with a RAN Telco audience, it will be more accurate to point out that the RT or non-RT is the characterization of the RHCOS Kernel and not the RAN components. When it comes to O-RAN architecture, there are elements that are referred as near-realtime or non-realtime RICs, which has nothing to do with the Kernel running mode or the workload itself but have direct integrations with these.

Yes, you are right. Rephrased as follows:
"From the high level blocks composing RAN (CU, DU, RU), only CU and DU can be virtualized and implemented as cloud-native functions.
CU / DU split is driven mostly by latency and traffic volume optimizations. 
A DU can be seen as an application performing intensive digital signal processing, data transfer and algorithmic tasks in constrained time budgets. DU implementations often use hardware acceleration (FPGA, GPU, eASIC) for DSP processing offload. There are also software-only implementations (FlexRAN), based on AVX-512 instructions."

>     * The "typical RAN" deployment is operated & managed by different teams. Even when there will be a common base, the defintion of the clusters could be on a different repo (basically mono-repo vs multi-repo configurations). It might be good to either describe this as an option or call this out as something out of scope.
> 
Added this to non-goals:
"- There is no attempt to evaluate repository organization approaches (i.e.monorepo vs. multi-repo). The approach might be different for every particular mobile network operator. This project contains both profiles and clusters definitions in a single repository, but this can be changed by replacing the direct links with links to other Git repositories."

>     * The use of `kubernetes.io/...` label namespace is reserved for Kubernetes, we should use a different label namespace (e.g. ran.openshift.io/...)

Somebody told me that using "ran.openshift.io" will have hard time to be accepted by the upstream community, because it's openshift specific.
Can we just use "ran" as a prefix for all labels unrelated to "node-role.kubernetes.io"?

>     * The labeling scheme decoupling features from node roles should be defined. Following the label namespace example we can consider labels like this to enable a feature on a node irrespective of its MCP
> 
> 
> ```shell
> # Example of labels for a DU node
> ran.openshift.io/n3k=""
> ran.openshift.io/sriov=""
> ran.openshift.io/local-storage-operator=""
> ran.openshift.io/ptp=""
> 
> # Example of labels for CU node
> ran.openshift.io/sriov=""
> ran.openshift.io/local-storage-operator=""
> ran.openshift.io/ocpvirt=""
> ```

I'm not sure I'm getting all the reasons behind this need. If I take SR-IOV as an example, why would I need ran.openshift.io/sriov="" general label, if I already have more specific label, e.g. ran.kubernetes.io/worker-du-large-otwaon1234rd-m1?

>     * We might need to expand explaining the HW profile definition as it is not generic like described in the README.md. The hardware model with specific network and storge layout + bios configuration should be defined as node "profiles". The Telco will have very few of these at the beginign and only add very few of them per year (usually aligned to the release of new CPU models). Each one of these will have minimum two BIOS profiles, one with HT enabled, other with HT disabled. The preferred method in production is HT enabled. The BIOS configs are specific to the vendor. Because of this, as part of the preparaiton steps, the specific hardware sku + layout + BIOS config needs to be defined as a "profile" which is going to be used to reference specific PAO profiles, etc.
> 
> If we take a theoretical SuperMicro SKU of profile like the one described in the README.md, it will be something like this:
> 	
> CPU 	Intel Xeon Gold series
> RAM 	96GB, DDR4
> Disks 	1TB /dev/nvme0n1
> NUMA 0 	CPUs:
> 0 - 51 (with HT)
> 0 - 26 (without HT)
> Device name 	Model 	Comment
> eno1 	10GB-T 	On-board NIC, not used
> eno2 	10GB-T 	On-board NIC, not used
> ens1f0 	XXV710 	Midhaul-A networks
> ens1f1 	XXV710 	Midhaul-B networks
> ens3f0 	N3000 	Fronthaul-1, PTP1
> ens3f1 	N3000 	Fronthaul-2, PTP2 (future)
> n/a 	N3000 	Accelerator (this can represent up to 8 devices. This is determined by FEC configs and antenna layout)
> 
> Building the profile defintion should account for a setup like:
> 	Comment
> SMCI 	SuperMicro (their ticker)
> Hardware Config 	00
> Profile0 	BIOS config with HT
> Profile1 	BIOS config without HT
> 
> This results in a minimum of two resulting hardware profiles: `smci00-profile0`, `smci00-profile1` which are then used in the site-specific configurations

I totally agree, but I must admit I fell for the peer pressure here. One of the purposes of pushing to this repo is to connect to CI pipelines, which run on a generic / different clusters and / or GCP... 
Another point of view was that our work here should be unrelated to any specific hardware, because each customer will bring its own supported models, so the amount of supported models will only add up over time. And we don't intend to maintain a list of certified hardware models... Or do we now?

 
>     * The MCP naming scheme should be defined at least as a reference for the structure

There is MCP example in each of the building blocks here. I will elaborate  on considerations for dividing the nodes to different MCPs and naming examples. Is this what you mean?
 
>     * The use of large, medium, small is not appropriate for these scenarios as what is "large" today, will be "small" tomorrow and that will cause challenges on the operation of the deployments as different clusters might end up with different definitions of "small/med/large". That is another reason why the definition of a hardware profile and reference to those
>
Agree, WIP

>     * Each DU and CU instance is in its own namespace, the layout and description of the SR-IOV node policies should account for that
> 
Is it really configured per **instance**? And by instance do you mean node?
I can easily imagine that for FEC, but in a data center, wouldn't there be a pool of nodes with identical SR-IOV node policies?

>     * When it comes to the PAO, in the same cluster, on different locations or MCPs, the same hardware profile may have different PAO (usually because of different TopologyManager policies or different core reservations). The model should account for this.

Yep, this is already a case. I will add an explanation to the data center readme file.

>     * The PTP config is site-specific as parameters may be different

Will add an example to PTP customizations
> 
> We should probably have a deeper discussion on the various considerations for each type of operators. For example, depending on the RAN vendor or RAN version on a particular site, we will have OCP Virt or not. The structure should account for that.

Totally, and I'm happy to have this discussion started.
I left several things out of scope, trying to first get the structure reviewed and accepted, so later it will be possible to make incremental additions. 
Thanks again for the comments!

williamcaban:
Agree with the label namespace. What if we follow RFC 6761 (Section 6.5) that reserves "example.com/.net/.org" for documentation and follow the pattern with `ran.example.com/...`  Will that be more appropriate for upstream and in general?

WRT the SR-IOV labels it is how the Operator works, there is a label it matches to enable. the daemonset and the other is for the node policy to apply the actual config. So, in the example above `ran.example.com/sriov=''` will be used by the daemonset, while the other for the creation and configuration of the actual VFs. 

Thanks for all the updates.

vitus133:
Hi William,
> Agree with the label namespace. What if we follow RFC 6761 (Section 6.5) that reserves "example.com/.net/.org" for documentation and follow the pattern with `ran.example.com/...` Will that be more appropriate for upstream and in general?
Yes, I think it's a good solution that should hurt nobody. Changing that.

> WRT the SR-IOV labels it is how the Operator works, there is a label it matches to enable. the daemonset and the other is for the node policy to apply the actual config. So, in the example above `ran.example.com/sriov=''` will be used by the daemonset, while the other for the creation and configuration of the actual VFs.
Got it, will add the label. 
> Thanks for all the updates.
Thanks a lot!

vitus133:
> Agree with the label namespace. What if we follow RFC 6761 (Section 6.5) that reserves "example.com/.net/.org" for documentation and follow the pattern with `ran.example.com/...` Will that be more appropriate for upstream and in general?

Sure, will change to ran.example.com
> WRT the SR-IOV labels it is how the Operator works, there is a label it matches to enable. the daemonset and the other is for the node policy to apply the actual config. So, in the example above `ran.example.com/sriov=''` will be used by the daemonset, while the other for the creation and configuration of the actual VFs.

Thanks, got it. Will add the feature labels
> 
> Thanks for all the updates.

Thanks a lot!

feature-configs/5g-ref-nw/clusters/otwaon1234rd/cu-up/README.md
```diff
@@ -0,0 +1,10 @@
+# Data center CU-UP
+This is an example of a data center CU-UP configurations
+## Structure
+The structure is identical to [du-large-m1](../otwaon1234rd/du-large-m1).
+
+## Labeling
+In addition to having the master role, the node must be labeled as follows:
+- `node-role.kubernetes.io/worker-cu-up-otwaon1234rd=""` - used for machine config pool
+- `ptp/slave=""` - used as node selector for PTP
```
kenyis, line 9: I do not think we would enable PTP in the general CU example.  I do think we would enable NTP.  Should we add an NTP provisioning via MCO?

feature-configs/5g-ref-nw/clusters/otwaon1234rd/cu-up/nw-config/sriov/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
+# Base profile for the du-large flavor
```
kenyis, line 1: We should clean-up this comment.  This is cu-large.

feature-configs/5g-ref-nw/clusters/otwaon1234rd/cu-up/role-hardware/performance/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+# Base profile for du-large flavor
```
kenyis, line 1: Cleanup the comment.  This is for the CU.

feature-configs/5g-ref-nw/clusters/otwaon1234rd/cu-up/role-hardware/sctp/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+# Base profile for du-large flavor
```
kenyis, line 1: Again please clean up the comment.

feature-configs/5g-ref-nw/clusters/otwaon1234rd/remote-sites/otwaon23456rw/nw-config/sriov/sriov-networks.yaml
```diff
@@ -0,0 +1,26 @@
...
+  resourceName: du_fec_mh_otwaon23456rw
+---
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetwork
+metadata:
+  name: sriov-nw-du-fec-mh-u
+  namespace: openshift-sriov-network-operator
+spec:
+  vlan: 101
+  resourceName: du_fec_mh_otwaon23456rw
```
kenyis, line 26: We need to look at this part.  I think we need a new namespace since the N3000 does not use the generic SRIOV operator for configuration.  This is an example that I have seen:

apiVersion: sriovfec.intel.com/v1
kind: SriovFecClusterConfig
metadata:
  name: config
  namespace: n3000-operators-resources
spec:
  nodes:
    - nodeName: node1
      physicalFunctions:
        - pciAddress: 0000.1d.00.0
          pfDriver: pci-pf-stub
          vfDriver: vfio-pci
          vfAmount: 2
          bbDevConfig:
            n3000:
              # Network Type: either "FPGA_5GNR" or "FPGA_LTE"
              networkType: "FPGA_5GNR"
              # Programming mode: 0 = VF Programming, 1 = PF Programming
              pfMode: false
              flrTimeout: 610
              downlink:
                bandwidth: 3
                loadBalance: 128
                queues:
                  vf0: 16
                  vf1: 16
                  vf2: 0
                  vf3: 0
                  vf4: 0
                  vf5: 0
                  vf6: 0
                  vf7: 0
              uplink:
                bandwidth: 3
                loadBalance: 128
                queues:
                  vf0: 16
                  vf1: 16
                  vf2: 0
                  vf3: 0
                  vf4: 0
                  vf5: 0
                  vf6: 0
                  vf7: 0

feature-configs/5g-ref-nw/clusters/otwaon1234rd/remote-sites/otwaon23456rw/nw-config/sriov/sriov-node-policies.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetworkNodePolicy
+metadata:
+  name: sriov-nnp-du-fec-mh
+  namespace: openshift-sriov-network-operator
+# Adjust resourceName and nodeSelector  
+spec:
+  nodeSelector:
+    ran.kubernetes.io/worker-du-small-otwaon23456rw: ""
+  resourceName: du_fec_mh_otwaon23456rw
```
kenyis, line 21: Again fec uses a different operator.

feature-configs/5g-ref-nw/clusters/otwaon34567sno/nw-config/sriov/sriov-networks.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: SriovNetwork
+metadata:
+  name: sriov-nw-du-fec-mh-c
+  namespace: openshift-sriov-network-operator
+spec:
+  vlan: 100
+  resourceName: du_fec_mh_otwaon34567sno
+---
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetwork
```
kenyis, line 20: Need to change the API.  Note that we also need to deal with different operators for Mt Bryce vs N3000.

feature-configs/5g-ref-nw/profile-base/du-small/sriov/sriov-nw-du-fec-mh-u.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetwork
+metadata:
+  name: sriov-nw-du-fec-mh-u
+  namespace: openshift-sriov-network-operator
```
kenyis, line 5: Need to fix the operator.

feature-configs/5g-ref-nw/profile-base/du-small/sriov/sriov-nw-du-fec-mh-c.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetwork
+metadata:
+  name: sriov-nw-du-fec-mh-c
+  namespace: openshift-sriov-network-operator
```
kenyis, line 5: Need to fix the operator.

feature-configs/5g-ref-nw/profile-base/du-general/ptp/ptpconfig-slave.yaml
```diff
@@ -0,0 +1,117 @@
+apiVersion: ptp.openshift.io/v1
+kind: PtpConfig
+metadata:
+  name: slave
```
kenyis, line 4: Given this is fresh, can we use inclusive language?  T-SC here?

feature-configs/5g-ref-nw/profile-base/du-general/ptp/ptpconfig-grandmaster.yaml
```diff
@@ -0,0 +1,121 @@
+# The grandmaster profile is provided for testing only
+# It is not installed on production clusters
+apiVersion: ptp.openshift.io/v1
+kind: PtpConfig
+metadata:
+  name: grandmaster
```
kenyis, line 6: T-GM

feature-configs/5g-ref-nw/profile-base/du-general/ptp/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
+- ptpconfig-slave.yaml
```
kenyis, line 2: rename for inclusivity?

kenyis (changes requested):
A couple of questions:

1/  Do we disable NTP for the DU?
2/  How would we handle the N3000 bitstream flashing as a precondition?  Should we add a check?  Is that possible?

feature-configs/5g-ref-nw/README.md
```diff
@@ -0,0 +1,222 @@
+
+# __5G RAN reference structure__
+
+# Overview
+This folder contains the manifests for deploying the 5G RAN components on Kubernetes clusters. It complements the [`cn-ran-overlays`](../cn-ran-overlays), that defines a single DU profile, by addressing the manifests scalability aspects. The structure defined here can be extended beyond separate units, approaching closer to the structure maintained by a "typical" mobile network operator.
```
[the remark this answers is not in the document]
vitus133: What I had in mind is to deprecate cn-ran-overlays eventually, after the new folder can be connected to the CI. Still need to figure out how to handle the transition

> 1/ Do we disable NTP for the DU?

Yes, there is a "disable-chronyd" patch in the current profile that I missed to take in. Fixing that. 
@SchSeba, is there a plan to disable NTP from PTP operator?

> 2/ How would we handle the N3000 bitstream flashing as a precondition? Should we add a check? Is that possible?

There is a [PR-464](https://github.com/openshift-kni/cnf-features-deploy/pull/464/) adding N3000 operator deoplyment, will add the relevant changes to this PR too. 
Regarding the bitstream, I understand that we can't add it since it's proprietary. 
I think we should ask Intel whether they can release a freely open test bitstream, so anyone could load it just to test N3000 hardware is operational. @SchSeba, do you think they have something like that?

feature-configs/5g-ref-nw/clusters/otwaon1234rd/remote-sites/otwaon23456rw/nw-config/sriov/sriov-networks.yaml
vitus133: I think this is a naming problem and confusion that will be resolved after I fix all the comments made by William. 
Here 'fec' is Far Edge Cloud, related to the profile of a single-socket server.
In Intel's sriov fec objects, 'fec' means forward error correction.
So yes, need to fix, but not to n3000 config, just the sriovResourceName must become more clear

feature-configs/5g-ref-nw/profile-base/du-general/ptp/kustomization.yaml
vitus133: I know we are about to fix things for inclusivity. My thought here was the timing. Since we already have CI testing on the DU profile, it will take time until we upgrade the pipelines to do something meaningful tor all the building blocks we define here. In the meantime, @yuvalk  suggested that we might want to test that the new profile output is "object-exact" (not bit-exact, but produce the same objects with the same settings) as the old profile.
So what I did as a rule of thumb in this project, for everything related to the DU - keep it object-exact with the current DU profile.
This is a point I am happy to open for discussion :)

feature-configs/5g-ref-nw/README.md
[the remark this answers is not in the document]
vitus133: Added NTP disabling for DUs

feature-configs/5g-ref-nw/README.md
```diff
@@ -0,0 +1,222 @@
...
+# Building blocks
+## Deployment types
+### Data centers
+One of the desired effects of RAN transition to the Cloud is the ability to centralize its functions in data centers, utilizing standard IT infrastructure to provide pools of centralized RAN functions. 
+A data center deployment will typically contain one or more Kubernetes clusters hosting various types of mobile network workloads:
+- A pool of RAN CU-CP and / or CU-UP units
+- A pool of DU
+- Functions related to 5G core network
+- Management functions
+It is important to emphasize that any of the functions above are optional and totally depend on the mobile network operator preferences and requirements. The servers running the above workloads are assumed to have two CPU sockets (see [Assumptions on hardware types](#hw_types)
```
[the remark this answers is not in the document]
vitus133: fixed

feature-configs/5g-ref-nw/clusters/otwaon1234rd/cu-up/README.md
vitus133: fixed

feature-configs/5g-ref-nw/clusters/otwaon1234rd/cu-up/role-hardware/performance/kustomization.yaml
vitus133: fixed

feature-configs/5g-ref-nw/profile-base/du-small/sriov/sriov-nw-du-fec-mh-u.yaml
vitus133: Removed all the fec (far edge cloud) from the names to avoid confusion with forward error correction

feature-configs/5g-ref-nw/profile-base/du-small/sriov/sriov-nw-du-fec-mh-c.yaml
vitus133: Removed all the fec (far edge cloud) from the names to avoid confusion with forward error correction

feature-configs/5g-ref-nw/profile-base/cu-up/sriov/sriov-nnp-cu-up-numa0-bh-dpdk.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+# This works for Intel based NICs.
+# For Mellanox please change to:
+#     deviceType: netdevice
+#     isRdma: true
+  deviceType: vfio-pci
+  isRdma: false
+  nicSelector:
+# The exact physical function name must match the hardware used    
+    pfNames: ["ens1f0"]
```
lack, line 15: Since this is a template that *must* be specified with the right physical device name, perhaps the profile-base should have this commented out?

lack (commented):
I think this is a step in the right direction.  Good stuff!

feature-configs/5g-ref-nw/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+### __clusters__
+The [clusters](./clusters) folder contains the actual network tree of the cluster configurations. This is the only directory that is expected to be modified by partners for deployment or maintenance.
+#### On clusters and sites
+The scope of this work is limited to Kubernetes cluster configuration, so clusters are the main building blocks of this network model. However, a notion of sites can be also important to mobile network operators.
+
+The relation between clusters and sites can be complicated. There are large sites, such as data centers, that can accommodate several clusters. At the same time there could be clusters spanning over several physical sites with remote worker nodes, or geo-diverse deployments. There is no separate site notation in the structure above - site names are expected to be encoded in the cluster names. Remote worker node configurations can be optionally placed under the "remote-sites" subdirectory for clarity. In any case, application of a particular configuration to a particular cluster is expected to be handled by tools external to Kubernetes, on the upper orchestration level.
```
imiller0, line 112: The statement `site names are expected to be encoded in the cluster names` seems to imply that a site will always equal a cluster. I think this is misleading since a cluster can span many physical sites as you describe in the RWN case. For that reason I think we should avoid correlating sites and clusters. Something along the lines of:

> A cluster may serve one or more cell towers/locations/physical sites depending on a carrier's deployment model. Cluster naming is entirely under the control of the carrier but will be used by Zero Touch Provisioning to correlate the cluster to the configuration defined within this repository.

williamcaban: Because of the non-trivial relationship between a control plane of a cluster and sites reporting to it, using a `remote-sites` directory might not communicate the intended separation of concerns for the node pools. The same challenge will happen when a regional or local datacenter is used as an aggregation for multiple roles. In these setups, CU pools and DU pools can be in the same physical location using different types of pools (MCPs). In this case, the pools will be identified with different names (e.g. `ldc1-ran-cu-dell01-profile2` and `ldc1-ran-du-smci00-porfile0`) even when at the same physical location.

feature-configs/5g-ref-nw/operators/performance/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+resources:
+# PAO deployment
+## Uncomment for the upstream branch
+- ../../../deploy/performance
+## Uncomment for the release-4.6 branch
```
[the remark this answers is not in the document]
vitus133: For 4.6 and 4.7 we've got cn-ran-overlays. This work should potentially include all 4.8+ content. I think I'd rather remove the comment

feature-configs/5g-ref-nw/profile-base/cu-up/sriov/sriov-nnp-cu-up-numa0-bh-dpdk.yaml
vitus133: Good point, thanks

feature-configs/5g-ref-nw/profile-base/du-general/ptp/kustomization.yaml
ijolliffe: maybe capture as a TODO

williamcaban:
CU workload does not uses `idle=poll`, only DU workload on locations with the right cooling and power (usually DU pool configs) is where we can see `idle=poll`. Outside those, that setting is not a default.

_W

MarSik:
Also you can now add additional option for tuning clusters where all workload will be DPDK based (almost no netqueues needed) as we merged https://github.com/openshift-kni/performance-addon-operators/pull/572

The example is simple:

```
spec:
  ...
  net:
    userLevelNetworking: true
```

See https://github.com/openshift-kni/performance-addon-operators/blob/master/docs/performance_profile.md#net for more info

feature-configs/5g-ref-nw/profile-base/cu-up/sriov/sriov-nnp-cu-up-numa0-bh-dpdk.yaml
vitus133: @lack 
Actually after rethinking this, since we are going to certify hardware platforms, NIC device names will be the same in all the servers of the same platform. The only thing we are supposed to customize is VLANs. That's why I initially fixed the device names in the profile, and VLANs in the customizations.
Please tell me whether it makes more sense now

yuvalk: @vitus133 see
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/consistent-network-interface-device-naming_configuring-and-managing-networking
that means that certification will need to dictate what PCIe slots to use (otherwise names will be different)
so are we really dictating slot 1?

also, I guess this call for documentation (in README.md?) that this is a single port (no bonding) configuration, always using port#0 of the NIC. you are also dictating that

feature-configs/5g-ref-nw/README.md
vitus133: @wcabanba
In my example, the pools are identified by different names (du-dc-m1, and there could be du-dc-m2 and so on for all the variants). I added the RWN example in the `remote-sites` to emphasize that it is a different site. The RWN could have been sitting in the same directory from the  MCP perspective, as it belongs to the same cluster. But since the cluster was named using CLLI, which is a site attribute, this could have been logically wrong.
Does it make sense?

vitus133: > The statement `site names are expected to be encoded in the cluster names` seems to imply that a site will always equal a cluster. I think this is misleading since a cluster can span many physical sites as you describe in the RWN case. For that reason I think we should avoid correlating sites and clusters. Something along the lines of:
> 
> > A cluster may serve one or more cell towers/locations/physical sites depending on a carrier's deployment model. Cluster naming is entirely under the control of the carrier but will be used by Zero Touch Provisioning to correlate the cluster to the configuration defined within this repository.

@imiller0 thanks, fixing this text to be more palatable. Or do you think we should totally avoid the mention of sites in this structure? I remember we agreed on just the opposite in one of ztp working sessions, but I'm ready to reconsider since it will simplify things a lot

feature-configs/5g-ref-nw/README.md
```diff
@@ -0,0 +1,140 @@
...
+### Three-node clusters
+Three-node clusters are deployments where minimum three physical servers are used as both infrastructure nodes and worker nodes. Additional worker nodes can be configured if needed.
+This type of deployment is expected to run on the Far Edge in cases where single node deployment is not sufficient. The workloads can vary from a small DU pool, to a full RAN and core network setup, oriented to local MNO data centers or private network deployments.
+
+
+
+## Assumptions on RAN CNF requirements
+### <a name="cu_up_nw"></a>cu-up networks
+
+<img src="images/cu-up.png" width=800>
```
yuvalk, line 76: can you include an index with all the abbrevation used?
up?
nw?
nnp?

naming rules, how are they constructed?
it's almost impossible to understand for example what `sriov-nnp-cu-up-numa1-mh-u` means
I'll try my best guess here
sriov network-node-polocy central-unit user-plane(?) numa1 mid-haul u(?)

vitus133: Yes indeed. 
Will add the naming anyway :)

feature-configs/5g-ref-nw/profile-base/du-dual/performance/performance-profile.yaml
```diff
@@ -0,0 +1,20 @@
...
+    reserved: "0-1"
+  hugepages:
+    defaultHugepagesSize: 1G
+    pages:
+    - count: 32
+      size: 1G
+  realTimeKernel:
+    enabled: true
+  additionalKernelArgs:
+  - nosmt
```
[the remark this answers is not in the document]
vitus133: Thanks @MarSik I didn't know it changed for v2. Am I right that in v1 all these were set by PAO by default?

feature-configs/5g-ref-nw/README.md
vitus133: Added a section to the main README

feature-configs/5g-ref-nw/profile-base/du-single/performance/performance-profile.yaml
```diff
@@ -0,0 +1,20 @@
...
+    isolated: "2-25"
+    reserved: "0-1"
+  hugepages:
+    defaultHugepagesSize: 1G
+    pages:
+    - count: 16
+      size: 1G
+  realTimeKernel:
+    enabled: true
+  additionalKernelArgs:
```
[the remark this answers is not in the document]
vitus133: done

feature-configs/5g-ref-nw/profile-base/du-dual/performance/performance-profile.yaml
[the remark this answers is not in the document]
vitus133: done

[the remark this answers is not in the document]
MarSik: Nope, we have never set them by default. idle=poll is very heavy and has huge impact on power consumption and heat.

PR #460, opened 2021-03-10
Add tc bandwidth limitation MachineConfig demos

oribon:


SchSeba (commented):
/hold 

we need to wait until rhcos move to 8.4 with the tc binary

PR #462, opened 2021-03-11
VRF Test: reduce the permissions needed to use a VRF

fedepaol:
We are running the tests with priviledged pod, here we do what a user
would do, assigning the least permissions which are NET_RAW


SchSeba:
/hold

please add another commit to enable this tests on the u/s ci

PR #467, opened 2021-03-19
Update documentation about the 'container-mount-namespace' cnf-test

lack:


lack:
This should fix up the documentation fallout from PR #435

fedepaol:
@marzianor9 @evgenLevin  not sure we did the same for gk / n3000, but in case we need to
cc @mmirecki @SchSeba

PR #470, opened 2021-03-21
Add ZTP directory and initial infrastructure

ijolliffe:
Signed-off-by: ijolliff@redhat.com


ijolliffe:
@lack has some good ideas to simplify this - so he will pick up with a new PR

PR #472, opened 2021-03-23
Configure RAN owners

lack:
This adds the RAN team members to the default 'reviewers' group, and also defines a new 'RAN-approver' alias, which will have approver access to both feature-configs/cn-ran-overlays and the forthcoming feature-configs/5g-ref-nw DU profile areas.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -14,6 +14,15 @@ aliases:
     - vladikr
     - mmirecki
     - yuvalk
+    - ijolliffe
+    - yrobla
+    - serngawy
+    - imiller0
+    - lack
+    - aneeshkp
+    - josephdrichard
+    - vitus133
+    - kenyis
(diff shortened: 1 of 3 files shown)


lack:
Offline conversation with @yuvalk : I'll be retooling this PR shortly.

lack:
@yuvalk : Updated as you asked.  As you can see, openshift-ci-robot gets angry and adds the `do-not-merge/invalid-owners-file` label if you define an alias and try to use that new alias within the same PR :(

yuvalk:
> @yuvalk : Updated as you asked. As you can see, openshift-ci-robot gets angry and adds the `do-not-merge/invalid-owners-file` label if you define an alias and try to use that new alias within the same PR :(

so, I think this is because of the order by which the `approve` plugin works, it first read the OWNERS_ALIASES and only then apply the PR and checking the OWNERS file

PR #481, opened 2021-03-30
Add SNO systemd mount scanning fix

vitus133:
This commit adds systemd mount scanning fix link to the ran-general profile.
/assign @lack


feature-configs/5g-ref-nw/profile-base/ran-general/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
+- ../../../deploy/container-mount-namespace
 - sctp
```
lack, line 3: Why are we referring to the outer reference `../../../deploy/container-mount-namespace` but a local `sctp`?

feature-configs/5g-ref-nw/profile-base/ran-general/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
+- ../../../deploy/container-mount-namespace
```
lack, line 2: Adding this to profile-base/ran-general will add this to all DU profiles, not just SNO as the commit text implies, won't it?

For the record, I don't think that's a wrong thing to do, but I think the commit message should be more clear that this is adding this feature to all RAN profiles, not just SNO.

lack (changes requested):
Should we also rename the role-hardware/sctp -> ran-general in the non-SNO cases to keep things consistent?

PR #490, opened 2021-04-14
Add initial playbook for deploying AI operator

yrobla:
These playbooks deploy a provisioner cluster with
ipv6+disconnected features, that can be used to deploy AI
on top. Then it mirrors the needed images, and finally
deploys assisted installer operator.


yrobla:
contained invalid file

PR #493, opened 2021-04-18
Update k8s dependencies to 1.20.4

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>


fromanirh:
/lgtm
I think it's ok for master-only change.

fromanirh:
it may be not obvious from the commit message/PR description: we want to update PAO deps - to pull in tests and test fixes, but we need to bump k8s deps first: https://github.com/openshift-kni/cnf-features-deploy/pull/492

yuvalk:
> it may be not obvious from the commit message/PR description: we want to update PAO deps - to pull in tests and test fixes, but we need to bump k8s deps first: #492

Ok.
Do we havea BZ?
For the future, can we be more explicit in the commit msg?

adding the why helps understand the story

fromanirh:
> > it may be not obvious from the commit message/PR description: we want to update PAO deps - to pull in tests and test fixes, but we need to bump k8s deps first: #492
> 
> Ok.
> Do we havea BZ?
> For the future, can we be more explicit in the commit msg?
> 
> adding the why helps understand the story

Point taken. I didn't request that in the review and this is on me. We can and should do better. I will be stricter next time.

cynepco3hahue:
In general updating k8s dependencies should be some regular routine, we should be aligned with OpenShift repositories. It my bad that I did not specify that I had some dependency issues under the PAO.

PR #499, opened 2021-04-22
WIP - Add configmap ai

yrobla:


yrobla:
configmap cannot be mapped from infraenv, so this is not useful

PR #500, opened 2021-04-22
Add support for cards on secondary NUMA

oribon:
We need 5 hugepages for the tests so we configure 10 without specifying a NUMA node,
this way it will get spread equally 5 between the 2 default NUMAs to support any network devices on them


SchSeba:
@sabinaaledort please take a look on this PR we may need to change the performance profile in the CI

PR #505, opened 2021-04-28
Remove ztp-playbooks from cnf-features-deploy

yrobla:
This will be moved to ztp-cluster-deploy repo


yuvalk:
even after this PR, there are still some files under `ztp-playbooks` directory (like the install-ai-operator)
assuming this is intentional (is it?)
maybe you do want to keep the README file too?

serngawy (approved):

PR #507, opened 2021-04-28
Update baseline policies to match expected SNO DU configuration

imiller0:


ztp/ztp-ran-policy-generator/sourcePolicies/PerformanceProfile.yaml
```diff
@@ -3,8 +3,7 @@ kind: PerformanceProfile
 metadata:
   name: $name
 spec:
-  additionalKernelArgs:
-    - nosmt
+  additionalKernelArgs: []
```
serngawy, line 6: if it is an empty array, it could be better to remove it.

ztp/ztp-ran-policy-generator/ranGenTemplateExamples/site-du-sno-1-ranGen.yaml
```diff
@@ -49,17 +49,10 @@ sourceFiles:
...
       # ex: overriding the additionalKernelArgs values exist in the source policy file PerformanceProfile
-      additionalKernelArgs:
-      - "nosmt"
-      - "nmi_watchdog=0"
-      - "audit=0"
-      - "mce=off"
-      - "processor.max_cstate=1"
-      - "idle=poll"
-      - "intel_idle.max_cstate=0"
+      additionalKernelArgs: []
```
serngawy, line 52: you don't need to add additionalKernelArgs as it is already defined in source PerformanceProfile.yaml  as additionalKernelArgs: []

serngawy (changes requested):

ztp/ztp-ran-policy-generator/sourcePolicies/PerformanceProfile.yaml
imiller0: Definitely could remove it, but having the empty array is a nice placeholder to let the user know what can be updated if they want.

serngawy (approved):

serngawy (approved):

PR #510, opened 2021-04-30
Skip xt_u32 test in sno

mmirecki:


cnf-tests/testsuites/e2esuite/xt_u32/xt_u32.go
```diff
@@ -63,7 +69,7 @@ var _ = Describe("xt_u32", func() {
 		var testNode string
 
 		BeforeEach(func() {
-			if !hasNonCnfWorkers {
+			if !hasNonCnfWorkers || isSingleNode {
```
SchSeba, line 72: I think is better to create a different skip for the single node with a better message (test need multiple nodes or something like that)

SchSeba (commented):
Thanks for the PR just a small comment

cnf-tests/testsuites/e2esuite/xt_u32/xt_u32.go
```diff
@@ -66,6 +72,9 @@ var _ = Describe("xt_u32", func() {
 			if !hasNonCnfWorkers {
 				Skip("Skipping as no non-enabled nodes are available")
 			}
+			if isSingleNode {
```
yuvalk, line 75: Actually, this is literally not true for this test.
negative test need, as described above, a nonCnfWorker.
but one of that is enough.

in (telco) SNO, this should be skipped by the above `if !hasNonCnfWorkers` as SNO should be a CNF worker..

SchSeba: This was added so the user will not even need to add the environment variable

yuvalk: > This was added so the user will not even need to add the environment variable

what environment variable?

mmirecki: XT_U32TEST_HAS_NON_CNF_WORKERS
As I understand we need a non cnf worker node to run.

SchSeba: @mmirecki please move this if before the other one so we skip the test for the right reason

mmirecki: > @mmirecki please move this if before the other one so we skip the test for the right reason

Done

yuvalk: so why cant we just assign a default value to that XT_U32_TEST_HAS_NON_CNF_WORKERS ?
(true for all such negative tests)
otherwise, we might consider labeling tests 
so that we can label tests that are suitable for snos (or not suitable)
similarly to how this is done in origin-tests
@fedepaol

mmirecki: In SNO we want to skip regardless of what value user sets to this env var (unless we want to be able to run test which we know will fail)

SchSeba (commented):
/lgtm
/approve

PR #514, opened 2021-05-05
Allow running origin-tests on CI

sabinaaledort:
Running origin-tests on CI requires sourcing the ci-operator setup
to get the cloud provider environment varibales which describe the 
type of the cluster the CI is running on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,19 @@ origin-tests-disconnected-environment:
 		ORIGIN_TESTS_REPOSITORY="$(ORIGIN_TESTS_REPOSITORY)" ORIGIN_TESTS_FILTER="$(ORIGIN_TESTS_FILTER)" \
 		hack/run-origin-tests.sh
 
+origin-tests-on-ci:
+	export SHARED_DIR=/tmp/shared
+	rm -rf /tmp/gcloudconfig/
+	export TEST_CSI_DRIVER_MANIFEST=""
+	export TEST_TYPE=""
+	ci_operator_source_file="/go/src/github.com/openshift/release/ci-operator/step-registry/openshift/e2e/test/openshift-e2e-test-commands.sh"
+	source_file_cut_line=$(grep --text -n "function suite" $ci_operator_source_file | cut -f1 -d:)
+	source_file_cut_line=$(($source_file_cut_line-1))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,6 +69,19 @@ origin-tests-disconnected-environment:
 		ORIGIN_TESTS_REPOSITORY="$(ORIGIN_TESTS_REPOSITORY)" ORIGIN_TESTS_FILTER="$(ORIGIN_TESTS_FILTER)" \
 		hack/run-origin-tests.sh
 
+origin-tests-on-ci:
+	export SHARED_DIR=/tmp/shared
+	rm -rf /tmp/gcloudconfig/
+	export TEST_CSI_DRIVER_MANIFEST=""
+	export TEST_TYPE=""
+	ci_operator_source_file="/go/src/github.com/openshift/release/ci-operator/step-registry/openshift/e2e/test/openshift-e2e-test-commands.sh"
```
yuvalk, line 77: I hate this implicit dependency
if they change something in `openshift/release` we'll break.
cant we have the ci lane to be multi-steps and add it there?
@fedepaol

Makefile
```diff
@@ -69,6 +69,19 @@ origin-tests-disconnected-environment:
...
+	source_file_cut_line=$(grep --text -n "function suite" $ci_operator_source_file | cut -f1 -d:)
+	source_file_cut_line=$(($source_file_cut_line-1))
+	head --lines=+$source_file_cut_line $ci_operator_source_file >> temp_source_file.sh
+	source temp_source_file.sh
+	cd /go/src/github.com/openshift-kni/cnf-features-deploy/
+	CLUSTER_PROVIDER="${TEST_PROVIDER}" make skopeo-origin-tests
```
yuvalk, line 83: I think I found a better way
apparently we dont need `skopeo` as `oc adm` include this functionality:
https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/e2e/gcp/libvirt/setup/openshift-e2e-gcp-libvirt-setup-commands.sh#L36

@fedepaol - FYI

Makefile
fedepaol: Not sure I got the comment right, but the problem now is that the "test" step is all or none.
Ideally, we would like to have a "prepare for origin tests" step and a "test" step that we override, but that's not what we have today

PR #518, opened 2021-05-06
Add downward API verification

gkopels:


gkopels:
Opening again after rebase

PR #521, opened 2021-05-06
Update source policy custom resource reference

serngawy:
- Moving ztp/ztp-policy-generator/sourcePolicies to ztp/source-policy-crs
- Create links for custom resources under feature-config/deploy
- Update reference to source policy crs


imiller0 (commented):
/lgtm

PR #526, opened 2021-05-11
Add missing attributes to performance profile

imiller0:
The PerformanceProfile needs to have the machineConfigPool and
userLevelNetworking set.

### DIFF
--- a/ztp/source-policy-crs/PerformanceProfile.yaml
+++ b/ztp/source-policy-crs/PerformanceProfile.yaml
@@ -13,6 +13,10 @@ spec:
       - size: $size
         count: $count
         node: $node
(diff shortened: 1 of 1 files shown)


serngawy (approved):

yuvalk:
it will be nice if commit msg include the answer to: why is this needed?
BTW: since in this case it's 2 different answers (mcp selector needed because of the SNO bug/shortcoming, level networking is needed to improve performance by aligning reserved cpus and net queue sizes), i'd consider splitting into 2 commits.

BBTW: the mcp selector, hopefully can be removed once the issue is resolved, allowing for a more common/generic profile that can apply to RWNs too

PR #530, opened 2021-05-13

sabinaaledort:
dpdk test requires at least 5 hugepages to work

### DIFF
--- a/feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
+++ b/feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
@@ -18,5 +18,5 @@ spec:
   hugepages:
     pages:
     - size: "1G"
-      count: 4
+      count: 5
       node: 0


feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
```diff
@@ -18,5 +18,5 @@ spec:
   hugepages:
     pages:
     - size: "1G"
-      count: 4
```
SchSeba, line 21: for this one we need to change it to `10` and remove the node so we allocate 5 on every numa

yuvalk:
I had a simila PR#508. just mentioning it (I'll close it now)

yuvalk:
also worth mentioning in your commit msg, that this is done in favor of d/s. have no impact u/s (which doesnt uses the typical baremetal)

sabinaaledort:
Will open another PR on master and will backport to release-4.6

PR #531, opened 2021-05-13

sabinaaledort:
dpdk test requires now at least 10 hugepages to work.

Done in favor of d/s as the typical baremetal performance profile is not used u/s.


feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
```diff
@@ -18,5 +18,5 @@ spec:
   hugepages:
     pages:
     - size: "1G"
-      count: 5
+      count: 10
       node: 0
```
SchSeba, line 22: lets remove the node here to allow 5 on every numa (when we have 2)

PR #535, opened 2021-05-14
Bug 1957175: Replace github.com/satori/go.uuid with github.com/gofrs/uuid

fedepaol:
satori/go.uuid is not maintained anymore and it's maintained fork
is github.com/gofrs/uuid so we switch.


cgoncalves:
I believe it is not needed. I could not find any active branch using this dependency.

PR #539, opened 2021-05-18

sabinaaledort:
dpdk test requires now at least 10 hugepages to work.

Done in favor of d/s as the typical baremetal performance
profile is not used u/s.

Removing the node to allocate 5 hugepages on every NUMA
node (when we have 2).

### DIFF
--- a/feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
+++ b/feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
@@ -18,5 +18,4 @@ spec:
   hugepages:
     pages:
     - size: "1G"
-      count: 4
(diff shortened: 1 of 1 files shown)


feature-configs/typical-baremetal/performance/performance_profile.patch.yaml
```diff
@@ -18,5 +18,4 @@ spec:
   hugepages:
     pages:
     - size: "1G"
-      count: 4
-      node: 0
+      count: 10
```
SchSeba, line 21: please add a new line here

SchSeba (commented):
/lgtm
/approve

PR #540, opened 2021-05-19
a fancy verify commits makefile target

yuvalk:


hack/verify-commits.sh
```diff
@@ -0,0 +1,18 @@
...
+	fi
+}
+trap finish EXIT
+
+# list commits
+for commitish in $(git log --oneline origin/master..HEAD | cut -d' ' -f 1); do
+	commit_msg_filename=$(mktemp)
+	git log --format=%B -n 1 $commitish > $commit_msg_filename
+	if ! .githooks/commit-msg $commit_msg_filename; then
+		echo validatoin failed for $commitish
```
imiller0: typo (validation)

.githooks/commit-msg
```diff
@@ -0,0 +1,29 @@
...
+)
+exceptions=( go.mod go.sum LICENSE Makefile OWNERS OWNERS_ALIASES README.md )
+
+if ! [[ "$(head -1 $1)t sh" == *":"* ]]; then
+	echo ERROR: commit msg subject must include component name
+	exit 4
+fi
+msg_prefix=$(head -1 $1 | cut -d':' -f 1)
+for filename in $(git diff --name-only --cached); do
+	if [[ ! " ${exceptions[@]} " =~ " $filename " ]]; then
```
fedepaol, line 23: so if I understand this correctly, if a commit hits multiple components we require every component to be added right?
Would it be hard to take the root path from a file and use that automatically?

hack/verify-commits.sh
fedepaol, line 15: nit: validation

Makefile
```diff
@@ -118,6 +118,12 @@ validate-test-list:
 	@echo "Comparing newly generated docs to existing docs"
 	cnf-tests/hack/compare-gen-md.sh
 
+verify-commits:
```
fedepaol, line 121: This can go under the `ci-job` makefile rule

.githooks/commit-msg
yuvalk: not exactly, I'm checking for a single component
so unless the file is @exception, it will fail if you try for example to update hack and modify ztp in the same commit.
you'd need to break it into 2 atomic commits.

Makefile
yuvalk: ah, this is easier than adding an additional lane then, great, will do

yuvalk: done

.githooks/commit-msg
imiller0: There are times where a single logical change/update will span across boundaries. For example when updating the "DU profile" captured in ztp/source-policy-crs (contains symlinks to feature-configs/deploy/....) and ztp/ztp-policy-generator/ranPolicyGenTempExamples. I may need to update the template under the policy generator and have a corresponding change to one of the source policies which is under feature-configs .

yuvalk:
> There are times where a single logical change/update will span across boundaries. For example when updating the "DU profile" captured in ztp/source-policy-crs (contains symlinks to feature-configs/deploy/....) and ztp/ztp-policy-generator /ranPolicyGenTempExamples. I may need to update the template under the policy generator and have a corresponding change to one of the source policies which is under feature-configs .

Arguably, that means your commit is not atomic, and you'd have to split it (at least) into 2 commits, one for each component (ztp, features)

It is perfectly fine tough for a single PR to span multiple components, with multiple commits.

.githooks/commit-msg
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/sh
+
+declare -A components
+components=( \
+	["cnf-tests"]="cnf-tests","go.mod","go.sum" \
```
fedepaol, line 5: go.mod / go.sum may impact the policy generator too

yuvalk: so will keep them in exceptions list then

yuvalk: actually maybe better to add it in ztp too

yuvalk: and I'm contemplating the vendor directory
if it need its own component or be part of these 2 too.

yuvalk: hmm actually, the other way around
I'll keep the vendor component
but allow `vendor:` commits to touch cnf-tests and ztp too (so when updating vendoring, must do that with proper component but can also update relevant go files

yuvalk:
now a good time for this
I'll also add the lane on openshift/release

.githooks/commit-msg
```diff
@@ -0,0 +1,43 @@
...
+		is_prefix_found=false
+		IFS=', ' read -r -a prefixes <<< "${components[$msg_prefix]}"
+		for prefix in "${prefixes[@]}"; do
+			if [[ "$filename" == "$prefix"* ]]; then
+				is_prefix_found=true
+				break
+			fi
+		done
+		if [ "$is_prefix_found" = false ]; then
+			echo ERROR: $filename is not part of $msg_prefix component
```
fedepaol, line 39: can you add a dumb proof message?

Something like for file xxx your commit message must start with yyy , please change your commit message to zzz

yuvalk: well, at least it's telling you what's wrong..
I'll try to add that logic

yuvalk: there might be multiple matches

yuvalk: done

yuvalk:
I've also added a `make print-git-components`
to show all the existing components

yuvalk:
/retest
now that pr#590 is merged, we should be ok

yuvalk:
/retest
infra issue with gather-network step
https://steps.ci.openshift.org/reference/gather-network

PR #541, opened 2021-05-24
Validate sriov nodes are targeted by the performance profile

oribon:
Validates that the node picked is also configured by the performance profile.
Solves a problem where a node without hugepages is picked for the tests and pods
are stuck on pending


fedepaol:
Good catch!

/approve
/lgtm

PR #542, opened 2021-05-31
Changed name of dpdk validation suite

evgenLevin:
Changed the dpdk validation package name from `dpdk` to `dpdk imagestream` so that this validation test can be skipped.

### DIFF
--- a/cnf-tests/TESTLIST.md
+++ b/cnf-tests/TESTLIST.md
@@ -13,7 +13,7 @@ The validation tests are preliminary tests intended to verify that the instrumen
 | validation container-mount-namespace should have a container-mount-namespace machine config for worker | Check the presence of a machine config that enables container-mount-namespace on workers. | 
 | validation container-mount-namespace should have the container-mount-namespace machine config as part of the master machine config pool | Check if the container-mount-namespace machine config is used by the master machine config pool | 
 | validation container-mount-namespace should have the container-mount-namespace machine config as part of the worker machine config pool | Check if the container-mount-namespace machine config is used by the worker machine config pool | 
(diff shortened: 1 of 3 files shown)


fedepaol:
@evgenLevin what is the purpouse of this change? Specifically, the current code checks for the imagestream and if not present skips. Is there a bug with the current behaviour? Or is this for some different scenario?

evgenLevin:
> @evgenLevin what is the purpouse of this change? Specifically, the current code checks for the imagestream and if not present skips. Is there a bug with the current behaviour? Or is this for some different scenario?

I have a scenario with disconencted SNO , I would like to take the dpdk image from variables (IMAGE_REGISTRY=cnfdc8-installer.cnfdc8.t5g.lab.eng.bos.redhat.com:5000/rh-osbs/ DPDK_TESTS_IMAGE = dpdk-base: v4.8.0-5), not from  imagestream. But in this case, the test will fail due to cluster is disconnected, so the WA will skip the validation:
```
  should have a tag ready from the dpdk imagestream [It]
    /remote-source/app/cnf-tests/testsuites/validationsuite/cluster/validation.go:260
    dpdk imagestream has no tags
    Expected
        <int>: 0
    to be >
        <int>: 0
```
```
`Events:
  Type		Reason		Age			From			Message
  ----		------		----			----			-------
  Warning	InvalidOutput	14s (x13 over 25s)	build-controller	Error starting build: an image stream cannot be used as build output because the integrated container image registry is not configured`
`

fedepaol:
Yes but, according to the existing code 

https://github.com/openshift-kni/cnf-features-deploy/blob/06daf7743284fa11848aa7717ec2e99ffaed3c46/cnf-tests/testsuites/validationsuite/cluster/validation.go#L261

that means that a "s2i-dpdk-app" imagestream is available on the cluster, but with no tags. Is that the case? If so, I think the right thing is to remove it.

PR #548, opened 2021-06-04
Add ManagedCluster and gitOps Application CRs

serngawy:
- Add ACM ManagedCluster and KlusterletAddonConfig CRs
- Add gitOps Application CR for ACM & ArgoCD
- Refactor policy builder
- Remove fs links and combine subscription policies in one policy


ztp/source-policy-crs/PaoSubscription.yaml
```diff
@@ -1 +0,0 @@
-../../feature-configs/deploy/performance/operator_subscription.yaml
```
lack, line 1: You're replacing this symlink with a copy of the subscription, which means we now have 2 copies of this in the repo - Is there a reason we don't want to refer to what's in the main feature-configs/deploy area?

lack: (looks like this same duplication is repeated for ptp and sriov)

serngawy: yes, the issue is having the OperatorGroup in different file than the subscription make the policyGen generate 2 different ACM policy for each. which make the operator installation fail if the subscription get created before the OperatorGroup. So for 4.8 will make them in 1 file so policyGen will generate 1 ACM policy for them AND in 4.9 we can change the logic in policyGen to accept many files under same policy.

lack: Is there a reason we can't use the kustomize.yaml in the existing operator locations to combine what's there into one file?

Either that, or we should fix the source operators to be a single file.

I don't think duplicating this info is the right approach.

lack (changes requested):
Adding duplicate information for the operators seems like the wrong approach.

PR #553, opened 2021-06-09
Downward API disable Discovery Mode

gkopels:
The downward api test case has been disabled for discovery mode with the following code
			if discovery.Enabled() {
				Skip("Downward API test disabled for discovery mode")
			}


SchSeba (commented):
/lgtm
/approve

PR #563, opened 2021-06-17
Bump up the version of operators

fedepaol:
Master became 4.9, pin the last available version of operators.


yuvalk:
that should go into the 4.8 branch too

yuvalk:
and a note, I think this wont actually work (but is fine here), before 4.8 goes live

fedepaol:
Yep, let's put a 
/hold

PR #566, opened 2021-06-17
Add site config CRD

serngawy:

### DIFF
--- a/ztp/ran-site-plan-crd/site-config-cr-ex.yaml
+++ b/ztp/ran-site-plan-crd/site-config-cr-ex.yaml
@@ -1,0 +1,105 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: dummy-site-plan-sno-du
+---
+apiVersion: ran.openshift.io/v1
+kind: SiteConfig
+metadata:
+  name: "dummy-site-plan-sno-du"
+  namespace: "dummy-site-plan-sno-du"
+spec:
+  baseDomain: "ptp.lab.eng.bos.redhat.com"
+  pullSecretRef:
+    name: "name of the pullSecret secret"
+  releaseImageUrl: "quay.io/openshift-release-dev/ocp-release:4.8.0-rc.0-x86_64"
+  sshPublicKey: "ssh-rsa "
+  sshPrivateKeySecretRef:
+    name: "name of the private key secret"
+  clusters:
+  - clusterName: "du-sno1"
+    clusterType: sno
+    clusterProfile: du
+    clusterLabels:
+      - groups: "group-du-sno"
+      - common: true
+      - sites : "site-du-sno"
+    clusterNetwork:
+      - cidr: 10.128.0.0/14
+        hostPrefix: 23
+    machineNetwork:
+      - cidr: 10.16.231.0/24
+    serviceNetwork:
+      - 172.30.0.0/16
+    manifestsConfig:
+      ntpServer: "pool.ntp.org"            # required for chrony manifest
+    additionalNTPSources:
+      - NTP.server1
+      - 10.16.231.22
+    nodes:
+      - hostName: "node1"
+        bmcAddress: "idrac-virtualmedia+https://10.16.231.87/redfish/v1/Systems/System.Embedded.1"
+        bmcCredentialsName:
+          name: "name of bmcCredentials secret"
+        bootMACAddress: "00:00:00:01:20:30"
+        bootMode: "UEFI"
+        rootDeviceHints:
+          hctl: '1:2:0:0'
+        # userData contain the user data to be passed to the host before it boots
+        userData:
+          bootKey: value1
+        cpuset: "2-19,22-39"
+        nodeNetwork:
+          interfaces:
+            - name: eno1
+              macAddress: 00:00:00:01:20:30
+            - name: eth0
+              macAddress: 02:00:00:80:12:14
+            - name: eth1
+              macAddress: 02:00:00:80:12:15
+          config:
+            # Example for the nmstate config. The interface names & macAddresses must match the defined interfaces above.
+            interfaces:
+              - name: eno1
+                macAddress: 00:00:00:01:20:30
+                type: ethernet
+                ipv4:
+                  enabled: true
+                  dhcp: false
+                  address:
+                  - 10.16.231.3/24
+                  - 10.16.231.28/24
+                  - 10.16.231.31/24
+                ipv6:
+                  enabled: true
+                  dhcp: false
+                  address:
+                  - 2620:52:0:10e7:e42:a1ff:fe8a:601/64
+                  - 2620:52:0:10e7:e42:a1ff:fe8a:602/64
+                  - 2620:52:0:10e7:e42:a1ff:fe8a:603/64
+              - name: bond99
+                type: bond
+                state: up
+                ipv6:
+                  address:
+                  - ip:2620:52:0:1302::100
+                  prefix-length: 64
+                  enabled: true
+                  link-aggregation:
+                    mode: balance-rr
+                    options:
+                      miimon: '140'
+                    slaves:
+                    - eth0
+                    - eth1 
+            dns-resolver:
+              config:
+                server:
+                  - 10.19.42.41
+            routes:
+              config:
+                - destination: 0.0.0.0/0
+                  next-hop-address: 10.16.231.254
+                  next-hop-interface: eno1
+                  table-id: 254
--- a/ztp/ran-site-plan-crd/site-config-crd.yaml
+++ b/ztp/ran-site-plan-crd/site-config-crd.yaml
@@ -1,0 +1,304 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  # name must match the spec fields below, and be in the form: <plural>.<group>
+  name: siteconfigs.ran.openshift.io
+spec:
+  # group name to use for REST API: /apis/<group>/<version>
+  group: ran.openshift.io
+  scope: Namespaced
+  names:
+    # plural name to be used in the URL: /apis/<group>/<version>/<plural>
+    plural: siteconfigs
+    # singular name to be used as an alias on the CLI and for display
+    singular: siteconfig
+    # kind is normally the CamelCased singular type. Your resource manifests use this.
+    kind: SiteConfig
+    # shortNames allow shorter string to match your resource on the CLI
+    shortNames:
+    - siteconfig
+  # list of versions supported by this CustomResourceDefinition
+  versions:
+    - name: v1
+      # Each version can be enabled/disabled by Served flag.
+      served: true
+      # One and only one version must be marked as the storage version.
+      storage: true
+      # A schema is required
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                baseDomain:
+                  description: BaseDomain is the base domain to which the cluster should belong.
+                  type: string
+                pullSecretRef:
+                  description: Ref to the pullSecret secret that will be used when pulling images.
+                  type: object
+                  properties:
+                    name:
+                      description: "name of the pullSecret secret"
+                      type: string
+                releaseImageUrl:
+                  description: |
+                    releaseImageUrl is the url to the image containing metadata for all
+                    components that run in the cluster, and is the primary and best way
+                    to specify what specific version of OpenShift you wish to install.
+                  type: string
+                sshPublicKey:
+                  description: sshPublicKey will be added to all cluster hosts for use in debugging.
+                  type: string
+                sshPrivateKeySecretRef:
+                  description: |
+                    sshPrivateKey is the private SSH key to use for access to
+                    compute instances. This private key should correspond to
+                    the public key included in the sshPublicKey. The
+                    private key is used by Hive to gather logs on the target
+                    cluster if there are install failures. The SSH private
+                    key is expected to be in the secret data under the
+                    ssh-privatekey key.
+                  type: object
+                  properties:
+                    name:
+                      description: "name of the privateKey secret"
+                      type: string
+                clusters:
+                  description: List of clusters that will be installed in the site.
+                  type: array
+                  items:
+                    type: object
+                    properties:
+                      clusterName:
+                        description: |
+                          ClusterName is the friendly name of the cluster. It is used
+                          for subdomains, some resource tagging, and other instances
+                          where a friendly name for the cluster is useful.
+                        type: string
+                      baseDomain:
+                        description: BaseDomain is the base domain to which the cluster should belong.
+                        type: string
+                      clusterType:
+                        description: clusterType either standard or sno
+                        type: string
+                        default: standard
+                        enum:
+                          - standard
+                          - sno
+                      clusterProfile:
+                        description: |
+                          clusterProfile specify the day 0 configuration that will get applied on the cluster. Must be one
+                           of the following none, du, cu, cu-du.
+                        type: string
+                        default: none
+                        enum:
+                          - none
+                          - du
+                          - cu
+                          - cu-du
+                      clusterLabels:
+                        description: clusterLabels specify the the day2 configuration policies that will get applied on the cluster.
+                        type: array
+                        items:
+                          type: object
+                          properties:
+                            common:
+                              description: common used to apply the common policies to the cluster should be true|false.
+                              type: boolean
+                            groups:
+                              description: |
+                                groups used to apply the groups of policies to the cluster. the groups value must match the policies
+                                group name. multipale groups names cloud be added example group-du, group-du-sno, group-dell
+                              type: string
+                            sites:
+                              description: |
+                                sites used to apply specific policies to the cluster. if sites didn't set the siteConfig.metadata.name
+                                will be considered as the value for sites. sites value must match the policies match label for sites.
+                              type: string
+                      clusterNetwork:
+                        description: ClusterNetwork is the list of IP address pools for pods. Default is 10.128.0.0/14 and a host prefix of /23.
+                        type: array
+                        items:
+                          description: |
+                            ClusterNetworkEntry is a single IP address block for
+                            pod IP blocks. IP blocks are allocated with size 2^HostSubnetLength.
+                          type: object
+                          required:
+                            - cidr
+                          properties:
+                            cidr:
+                              description: CIDR is the IP block address pool.
+                              type: string
+                            hostPrefix:
+                              description: |
+                                HostPrefix is the prefix size to allocate to each
+                                node from the CIDR. For example, 24 would allocate
+                                2^8=256 adresses to each node. If this field is
+                                not used by the plugin, it can be left unset.
+                              type: integer
+                              format: int32
+                      machineNetwork:
+                        description: MachineNetwork is the list of IP address pools for machines.
+                        type: array
+                        items:
+                          description: MachineNetworkEntry is a single IP address block for node IP blocks.
+                          type: object
+                          required:
+                            - cidr
+                          properties:
+                            cidr:
+                              description: CIDR is the IP block address pool for machines within the cluster.
+                              type: string
+                      serviceNetwork:
+                        description: |
+                          ServiceNetwork is the list of IP address pools for services default is 172.30.0.0/16.
+                          currently only one entry is supported.
+                        type: array
+                        maxItems: 1
+                        items:
+                          type: string
+                      apiVIP:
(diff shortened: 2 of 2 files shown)


ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                          where a friendly name for the cluster is useful.
+                        type: string
+                      baseDomain:
+                        description: BaseDomain is the base domain to which the cluster should belong.
+                        type: string
+                      clusterProfile:
+                        description: |
+                          clusterProfile speicfy the day0 configuration that will get appllied on the cluster must be one
+                          of the following du-sno, du, cu, cu-du
+                        type: string
```
browsell, line 86: We are combining two things here that should be separated.
1) Cluster type - standard or sno
2) Role - du/cu 

What is a cu-du ?

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                          ClusterName is the friendly name of the cluster. It is used
+                          for subdomains, some resource tagging, and other instances
+                          where a friendly name for the cluster is useful.
+                        type: string
+                      baseDomain:
+                        description: BaseDomain is the base domain to which the cluster should belong.
+                        type: string
+                      clusterProfile:
+                        description: |
+                          clusterProfile speicfy the day0 configuration that will get appllied on the cluster must be one
```
browsell, line 84: See above comment

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                        description: APIVIP is the virtual IP used to reach the OpenShift cluster's API.
+                        type: string
+                      ingressVIP:
+                        description: IngressVIP is the virtual IP used for cluster ingress traffic.
+                        type: string
+                      manifestsConfig:
+                        description: day0 required manifestsConfig.
+                        type: object
+                        properties:
+                          ntpServer:
```
browsell, line 165: Is this a list of servers ?

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                          cpuset:
+                            description: cpuset required by workload partitioning manifest.
+                            type: string
+                          managementWorkloadName:
+                            description: managementWorkloadName required by workload partitioning manifest.
+                            type: string
+                      additionalNTPSources:
+                        description: |
+                          AdditionalNTPSources is a list of NTP sources (hostname or
+                          IP) to be added to all cluster hosts. They are added to any
```
browsell, line 177: How is this different than the ntp config above ?

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                        maxItems: 1
+                        items:
+                          type: string
+                      apiVIP:
+                        description: APIVIP is the virtual IP used to reach the OpenShift cluster's API.
+                        type: string
+                      ingressVIP:
+                        description: IngressVIP is the virtual IP used for cluster ingress traffic.
+                        type: string
+                      manifestsConfig:
```
lack, line 161: I'm not sure grouping these under "things we need for manifests" is right.  The cpuset info is needed for the workload partitioning, which happens to be a manifest, but it's also going to be needed to populate the day-2 performance profile.

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                releaseImageUrl:
+                  description: |
+                    releaseImageUrl is the url to the image containing metadata for all
+                    components that run in the cluster, and is the primary and best way
+                    to specify what specific version of OpenShift you wish to install.
+                  type: string
+                sshPublicKey:
+                  description: sshPublicKey will be added to all cluster hosts for use in debugging.
+                  type: string
+                sshPrivateKey:
```
lack, line 47: Sensitive items like this (and the BMC password, and pull secret too) should be references to secrets, and not stored plaintext.

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                        properties:
+                          ntpServer:
+                            description: ntpServer required by chrony manifest.
+                            type: string
+                          cpuset:
+                            description: cpuset required by workload partitioning manifest.
+                            type: string
+                          managementWorkloadName:
+                            description: managementWorkloadName required by workload partitioning manifest.
+                            type: string
```
lack, line 173: The only valid value for this is "management" right now.  I don't think we know enough about what other workloads might exist in the future to bother making this configurable yet.

But do we want a boolean to enable/disable management workload partitioning entirely? (@browsell ?)

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: I would prefer to see a single list of NTP servers.

lack (changes requested):

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: from config view: 
du-sno -> to use the master node label for policy match config
du -> to add MCP du then use node label for policy match config
cu -> to add MCP cu then node label for policy match config
cu-du -> to add MCPs cu and du; early we were discussing that in LDC one cluster could manage CU and DU

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: the idea is to create the secret , so the user provide data in base64 encoded txt then we create the secret and other stuff

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: we can level it up, just we said apply only day0 config till we figure out how we will make the link with day2 config

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: I think it should be just one item based on the example below 
https://github.com/redhat-ztp/ztp-cluster-deploy/blob/master/create-cluster-sno/roles/chrony_day0/templates/chrony.conf#L2

if it is a list will remove it but if it is one item then can we consider the first item from the additionalNTPSources list below ?

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: okay, we can remove it. Just want to avoid the need for it later

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: But wouldn't this mean the user is putting their passwords in clear text in git?

serngawy: well, it is not clear text it is base64 encoded :D but this is even how k8s store the secrets

lack: The difference though is that k8s restricts access to the secrets by RBAC roles; encouraging storing of secrets in git is bad security practice.

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: I think the desired input from a user would be a single list of NTP servers, and then we would just take the 1st one to put in the chrony.conf - Is that right @browsell ?

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: Agreed.  Maybe just start with a placeholder for performance tuning data, where for now the managementCpuset is the only entry; later we can add other entries?

For now, for SNO anyway, it's safe to assume the managementCpuset is the same in the whole cluster (because there's only ONE node in the cluster anyway.  But I'm curious, @browsell : Is this safe to assume that the performance data could be in the cluster area, or is it more appropriate to have it per-node in the node section?

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: Yeah, I think it's too soon to know enough about how this will be configurable in the future; let's delete it for now.

browsell: +1, delete for now

ztp/ran-site-plan-crd/site-config-crd.yaml
browsell: Performance configuration would be mapped to a performance profile which would map to a mcp. So for a standard cluster we can't assume the settings will be cluster wide.  I think we need another section here perhaps to group these by mcp

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: I agree, its just user will have to provide secret names instead of the bmc info.

ztp/ran-site-plan-crd/site-config-crd.yaml
browsell: Do we need this at all ? How is this different then L174 ?

serngawy: this is single server, L174 is a list. will remove it and consider the first item in the L174 list.

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: So will apply it per node then ?

serngawy: this is hard to be different per node/MCP for day0 . if that the case needs to be day2 ?

lack: That makes sense, actually; a single cluster can have a list of MCP roles, and a list of nodes, and each node refers to one of the MCP roles.

@serngawy : Would you object to introducing a "role" or "pool" concept where we can put the performance data?

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: - For the SSH key specifically, can we just generate a private key and inject the associated public key without having to specify any of it in git? (And we would also inject additional SSH public key(s) specified by the user for their own reasons)
- For the BMC password, we should definitely use a secret instead of putting that info in git.  With that password, an attacker would own the machine entirely.
- The pull-secret contents is also protected and should maybe be protected in a secret as well?

Maybe @yuvalk  and @dbaker-rh can chime in with their thoughts on the security aspects of this CRD, given that we expect end-users to check their associated CRs into git?

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: yes, we will do it but its not possible to have multiple instances of the same  day0 manifest per cluster. its only 1 ISO per cluster. I believe we should ask the AI guys to provide that in the future allowing multiple ISO per cluster to provision different nodes. will set the cpu.set per node for now as we just have 1 node->cluster

lack: 1 iso per cluster shouldn't matter - The differentiation should be at the manifest-per-role level, and you can inject the same kind of day-0 MC object for the 'master' role and a slightly different one for the 'worker' role, and also 'worker-du' or 'worker-cnf' or whatever in the future.  One ISO should be able to add all of them just fine simultaneously...

ztp/ran-site-plan-crd/site-config-crd.yaml
vitus133: +1 on having all keys and passwords as secret references.
BTW, in this case SSH public key appearing as a string is inherited (I believe) from AI CRD. Probably need to have this raised there? Or is it OK to list public keys in clear text?

vitus133 (commented):
Great stuff! 
It is present in some cases, but not all - I think it could be helpful to specify which items are required for every object. For the optional ones specify what happens when omitted.

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: Yes, public keys are safe to expose in clear text.

@serngawy : Can you help us understand why we need a private key in this information at all?

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: I'd vote to only include the values we know we need for now, until we define other profile types.

ztp/ran-site-plan-crd/site-config-cr-ex.yaml
```diff
@@ -0,0 +1,107 @@
...
+    registries.conf: |
+      'unqualified-search-registries = ["registry.fedoraproject.org", "registry.access.redhat.com", "registry.centos.org", "docker.io"]
+      [[registry]]
+        location = "quay.io/openshift-release-dev/ocp-release"
+        insecure = false
+        mirror-by-digest-only = true
+    ca-bundle.crt: |
+      '-----BEGIN CERTIFICATE-----
+      e*@mple
+      -----END CERTIFICATE-----
```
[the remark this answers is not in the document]
serngawy: okay, I will remove it as the registry config belong to the AgentserviceConfig which is part of the AI operator installation.

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                        description: BaseDomain is the base domain to which the cluster should belong.
+                        type: string
+                      clusterProfile:
+                        description: |
+                          clusterProfile speicfy the day0 configuration that will get appllied on the cluster must be one
+                          of the following stander, sno, du-sno, du, cu, cu-du
+                        type: string
+                        default: stander
+                        enum:
+                          - stander
```
lack, line 90: Typo: `standard`?

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                          ClusterName is the friendly name of the cluster. It is used
+                          for subdomains, some resource tagging, and other instances
+                          where a friendly name for the cluster is useful.
+                        type: string
+                      baseDomain:
+                        description: BaseDomain is the base domain to which the cluster should belong.
+                        type: string
+                      clusterProfile:
+                        description: |
+                          clusterProfile speicfy the day0 configuration that will get appllied on the cluster must be one
```
lack, line 85: Type: `specify`

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                        default: stander
+                        enum:
+                          - stander
+                          - sno
+                          - du-sno
+                          - du
+                          - cu
+                          - cu-du
+                      clusterLables:
+                        description: clusterLables specify the the day2 configuration policies that will get applied on the cluster.
```
lack, line 97: Typo: `clusterLabels`

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: To @browsell 's comment, should we split this into:

clusterType: [sno|standard]
clusterProfile: [du|cu|none|...]

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: Should this be removed?

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                                Address holds the URL for accessing the controller on the network. Must contain the redfish or the HW identification in the
+                                url for example redfish-virtualmedia+https://12.13.222.2/redfish/v1/Systems/System.Embedded.1
+                                example for Dell idrac-virtualmedia+https://10.16.231.87/redfish/v1/Systems/System.Embedded.1
+                              type: string
+                            bmcCredentialsName:
+                              description: Bmc credential secret info
+                              type: object
+                              properties:
+                                name:
+                                  description: "Name of the bmcCredentials secret"
```
lack, line 201: Should this specify the format of the bmcCredential secret contents?  I'm assuming it should contain the username and the password, but how should these be encoded in the secret?

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                              description: Select the method of initializing the hardware during boot. Defaults to UEFI.
+                              type: string
+                              default: UEFI
+                              enum:
+                                - UEFI
+                                - UEFISecureBoot
+                                - legacy
+                            cpuset:
+                              description: cpuset required by workload partitioning manifest.
+                              type: string
```
lack, line 217: Here's a regex pattern that represents valid input for a cpuset: `^[0-9]+([-,][0-9]+)*$`

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: I noticed that you have cpuset down below; can this whole section be removed now?

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,298 @@
...
+                              pattern: '[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}'
+                            bootMode:
+                              description: Select the method of initializing the hardware during boot. Defaults to UEFI.
+                              type: string
+                              default: UEFI
+                              enum:
+                                - UEFI
+                                - UEFISecureBoot
+                                - legacy
+                            cpuset:
```
lack, line 215: Eventually this will be specified per pool, not per node, but for SNO this is fine.

lack (changes requested):

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: done. separate them

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: done

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: we agreed to leave it as the chrony manifest required single item and the ntp below is list

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: will keep it per node for now

ztp/ran-site-plan-crd/site-config-crd.yaml
serngawy: done.

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: I thought we agreed below (line 172) that we would remove the single item and use the 1st entry in the list as the item for the chrony manifest?

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: Missed an additional typo in the description field too :)

lack (changes requested):
We're getting close!

ztp/ran-site-plan-crd/site-config-crd.yaml
```diff
@@ -0,0 +1,304 @@
...
+                    cluster if there are install failures. The SSH private
+                    key is expected to be in the secret data under the
+                    ssh-privatekey key.
+                  type: object
+                  properties:
+                    name:
+                      description: "name of the privateKey secret"
+                      type: string
+                clusters:
+                  description: List of clusters tha will be installed in the site.
```
lack, line 69:
```suggestion
                  description: List of clusters that will be installed in the site.
```

ztp/ran-site-plan-crd/site-config-crd.yaml
lack: Apparently in a side-conversation, @browsell and @serngawy agreed it's good as-is for now.  I'm cool with that too.

lack (approved):

PR #575, opened 2021-06-24
Update golang image due to a rate limit when pulling from docker.io

sabinaaledort:


cnf-tests/Dockerfile
```diff
@@ -1,12 +1,14 @@
-FROM golang:1.13 AS builder
+FROM quay.io/openshift/origin-base AS builder
```
fromanirh, line 1: this works, but I wonder if we can just use the golang builder image OCP uses - it should be easier to track

yuvalk: which is?

SchSeba: we didn't found any golang image that is why @sabinaaledort use the base one

yuvalk: `registry.svc.ci.openshift.org/openshift/release:golang-1.13`
but actually in many ocp componentes, I see golang-1.16, or 1.15, some random examples:
`FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder`
`FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9 as builder`

fromanirh: everything else equal, I'd prefer to reference a golang image - not `origin-base`, which is a bit too opaque IMO

yuvalk: the ones with openshift-4.X will need constant maintenance 
though we might want it anyhow, to keep in line with the OCP used golang version

fromanirh:
ping?

sabinaaledort:
will be fixed in https://github.com/openshift/release/pull/19776

PR #576, opened 2021-06-24
Bump sriov and ptp operator using release 4.9 branch

SchSeba:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -85,7 +85,7 @@ replace (
 
 // Test deps
 replace (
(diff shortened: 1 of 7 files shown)


SchSeba:
the failed test are not related to this change

/override ci/prow/e2e-gcp-ovn

yuvalk:
yes, but if you override, then this change doesnt get tested.

yuvalk:
we have these failures, cross-wide, I think we need to spend some time investigating more deeply why they happen 
I'm on it with @fedepaol

PR #587, opened 2021-07-01
[release-4.6] Add a ref field to the dpdk build config

SchSeba:
This is needed as there was a change in dpdk and test-pmd


SchSeba:
/retest 
docker rate limit

PR #582, opened 2021-06-29
Support running dpdk tests on secure boot environment

SchSeba:
This commit continue the work done in the sriov operator vendored tests
to support secure boot enabled environments.

Signed-off-by: Sebastian Sch <sebassch@gmail.com>

PR #589, opened 2021-07-04
DPDK discovery fails on missing sriov testing namespace

oribon:
https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/cb3b24b87a4de9e50f2b684b3a602dd6e657a8a1/test/util/cluster/cluster.go#L73
fails since the sriov-conformance-testing namespace isn't created by the suite


oribon:
https://github.com/openshift-kni/cnf-features-deploy/pull/582

PR #592, opened 2021-07-06
Gatekeeper policy

dbaker-rh:
Add demo/gatekeeper directory to deploy demo policies.
Add test cases to (manually) test those policies.

### DIFF
--- a/feature-configs/demo/gatekeeper/10_Constraints.yaml
+++ b/feature-configs/demo/gatekeeper/10_Constraints.yaml
@@ -1,4 +1,4 @@
-## This yaml *must* be loaded after 00_ConstraintTemplates.yaml, and after the gatekeeper operator
+## This yaml gets loaded after 00_ConstraintTemplates.yaml, and after the gatekeeper operator
 ## has had time to create the required CRD for these objects to be created.
 ##
 ## To apply these policies:
--- a/feature-configs/demo/gatekeeper/10_Constraints.yaml
+++ b/feature-configs/demo/gatekeeper/10_Constraints.yaml
@@ -6,7 +6,7 @@
 ##
 
 ---
-# Block global tolerations
+# The first restriction here is to block global tolerations, such as seen here:
 #
 # oc -n foo get pod/bar -o yaml
 #
@@ -36,7 +36,7 @@ spec:
 
 
 ---
-# Block specific tolerations
+# The second restriction is to block specific tolerations, such as seen here
 #
 # oc -n foo get pod/bar -o yaml
 #
--- a/feature-configs/demo/gatekeeper/10_Constraints.yaml
+++ b/feature-configs/demo/gatekeeper/10_Constraints.yaml
@@ -1,0 +1,73 @@
+## This yaml *must* be loaded after 00_ConstraintTemplates.yaml, and after the gatekeeper operator
+## has had time to create the required CRD for these objects to be created.
+##
+## To apply these policies:
+## $ oc label namespace/FOO admission.gatekeeper.sh/tolerations=enforce
+##
+
+---
+# Block global tolerations
+#
+# oc -n foo get pod/bar -o yaml
+#
+#   ...
+#     spec:
+#       tolerations:
+#       - operator: Exists
+#
+
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: K8sRestrictGlobalToleration
+metadata:
+  name: deny-global-tolerations
+spec:
+  match:
+    kinds:
+    - apiGroups: [""]
+      kinds: ["Pod"]
+    namespaceSelector:
+      matchExpressions:
+      - key: admission.gatekeeper.sh/tolerations
+        operator: In
+        values: [ "enforce" ]
+
+  parameters:
+    allowGlobalToleration: false
+
+
+---
+# Block specific tolerations
+#
+# oc -n foo get pod/bar -o yaml
+#
+#   ...
+#     spec:
+#       tolerations:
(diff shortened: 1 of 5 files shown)


yuvalk:
do we have a policy for the grace period too?

dbaker-rh:
No.  tGPS would require a mutation, not a validation, to be useful.

feature-configs/demo/gatekeeper/10_Constraints.yaml
```diff
@@ -0,0 +1,73 @@
+## This yaml *must* be loaded after 00_ConstraintTemplates.yaml, and after the gatekeeper operator
```
fedepaol, line 1: This is not necessarily true.
The way the deployment works is that we iterate over all the yamls until everything works, so dependencies on CRDs are satisfied. That is the safe for the operators creating CRDs

feature-configs/demo/gatekeeper/10_Constraints.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - key: admission.gatekeeper.sh/tolerations
+        operator: In
+        values: [ "enforce" ]
+
+  parameters:
+    allowGlobalToleration: false
+
+
+---
+# Block specific tolerations
```
fedepaol, line 39: At first glance, this seemed a leftover comment, while it's explaining what's below.
Mind making that more explicit?

fedepaol: Same above

feature-configs/demo/gatekeeper/sample-testcases.yaml
```diff
@@ -0,0 +1,49 @@
+---
```
fedepaol, line 1: How about extending the e2e tests with these?

feature-configs/demo/gatekeeper/sample-testcases.yaml
```diff
@@ -0,0 +1,49 @@
...
+# This should fail to create due to the specific toleration
+apiVersion: v1
+kind: Pod
+metadata:
+  generateName: test-fail-specific-
+spec:
+  containers:
+  - image: docker.io/redhat/ubi8:latest
+    name: demo
+    command: ["/bin/sh", "-c", "while :; do date; sleep 10; done"]
```
fedepaol, line 11: can't you just sleep as the entry point?

fedepaol:
@dbaker-rh left a couple of comments.
In general, this seems to deserve the status of feature that we deploy to a cluster more than a demo.
If we are suggesting customers to use these policies, I think it makes sense to e2e them so we know they don't regress. WDYT?

feature-configs/demo/gatekeeper/10_Constraints.yaml
dbaker-rh: I agree.  It's true, and it's also not relevant.  The yaml must be loaded after the first one, but because you repeat/retry it will retry and load afterwards.  Agree that my comment can be updated to clarify, though.

feature-configs/demo/gatekeeper/10_Constraints.yaml
dbaker-rh: comment updated.

feature-configs/demo/gatekeeper/sample-testcases.yaml
dbaker-rh: agree that this is needed -- with the 4.9 mixup with gatekeeper I was going to defer that.  If you have time to help walk me through adding the e2e I'd appreciate some help in getting it integrated properly.

feature-configs/demo/gatekeeper/sample-testcases.yaml
dbaker-rh: I always do a while ; date ; sleep so the logs show constant updates.  Agree that just a sleep by itself would be functionally similar.

dbaker-rh (commented):
Thanks for the comments Federico.

Strong agree - adding e2e and feature-ifying it would be a very good idea here.

feature-configs/demo/gatekeeper/sample-testcases.yaml
fedepaol: Fine by me

feature-configs/demo/gatekeeper/sample-testcases.yaml
fedepaol: Sure, let's try to carve some time next week.

feature-configs/demo/gatekeeper/sample-testcases.yaml
mmirecki: command: ["/bin/bash", "-c", "sleep INF"]

feature-configs/demo/gatekeeper/sample-testcases.yaml
```diff
@@ -0,0 +1,49 @@
+---
+# This should fail to create due to the specific toleration
+apiVersion: v1
+kind: Pod
+metadata:
+  generateName: test-fail-specific-
```
mmirecki, line 6: The pods should be in the "gatekeeper-demo" namespace due to the namespace selector on the constraint templates.

feature-configs/demo/gatekeeper/sample-testcases.yaml
mmirecki: Let me take a shot at adding this to cnf tests

SchSeba:
closing this one feel free to reopen :)

/close

PR #600, opened 2021-07-08
features: Bump the custom catalog for intel operator

SchSeba:
The 4.8 redhat certified is not available yet so we need to use the 4.7 one


SchSeba:
/retest

check if this is a flake

PR #602, opened 2021-07-09
fix CI commit verifier (hack/verify-commits.sh)

yuvalk:
the previous #540 was assuming CI normally clones github,
where infact, it have only the local master branch.

script now uses github APIs to get the right commit id.
(see commit msg for more info)

### DIFF
--- a/hack/verify-commits.sh
+++ b/hack/verify-commits.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e -o pipefail
+
 function finish {
 	if [ -f "$commit_msg_filename" ]; then
 		rm -f $commit_msg_filename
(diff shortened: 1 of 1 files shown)


yuvalk:
/assign @fedepaol 
this is the fix PR
note I had to do #603 separately, because according to ci-operator config, the project_image dockerfile is taken from branch and not from PR. https://github.com/openshift/ci-operator/blob/master/CONFIGURATION.md#build_rootproject_image
just mentioning it here, if anyone bump into this in the future

hack/verify-commits.sh
```diff
@@ -1,18 +1,62 @@
...
+function is_merge_commit {
+	commitid=$1
+	local sha="$1"
+	msha=$(git rev-list -1 --merges ${sha}~1..${sha})
+	[ -z "$msha" ] && return 1
+	return 0
+}
+
+if [[ -z "$UPSTREAM_COMMIT" ]]; then
+	if [[ ! -z "$CI" ]]; then
```
fedepaol, line 21: Where these two variables are coming from?
Also, should we fail if UPSTREAM_COMMIT is not defined and CI is not defined? In that case, we would not set latest_upstream_commit I think

yuvalk: in prow CI=true
the rest (UPSTREAM_COMMIT and UPSTREAM_BRANCH are made especially for users who want to tweak this script)

yuvalk: UPSTREAM_COMMIT take precedence over CI (and UPSTREAM_BRANCH). so if we ever have a desire to override it there, we can.
if it's not CI, then UPSTREAM_BRANCH can be used

I'm also defensively checking if branch/ commit exist
as well as that there is a diff

hack/verify-commits.sh
```diff
@@ -1,18 +1,62 @@
...
+
 # list commits
-for commitish in $(git log --oneline origin/master..HEAD | cut -d' ' -f 1); do
+for commitish in $(git log --oneline $latest_upstream_commit..HEAD | cut -d' ' -f 1); do
 	commit_msg_filename=$(mktemp)
-	git log --format=%B -n 1 $commitish > $commit_msg_filename
-	if ! .githooks/commit-msg $commit_msg_filename; then
-		echo validation failed for $commitish
-		exit 10
+	if [[ $(git rev-list --no-walk --count --merges $commitish) == 0 ]]; then
```
yuvalk, line 55: this is also pretty cool, to avoid merge commits, which doesnt have a component ..

hack/verify-commits.sh
fedepaol: Mind adding a comment for that? It's not entirely clear to the new reader

yuvalk:
/override ci/prow/e2e-gcp-ovn
problem there is
`Jul 13 14:01:04.512: [INFO]: run command 'oc [exec -i -n openshift-machine-config-operator -c machine-config-daemon --request-timeout 30 machine-config-daemon-skd2w -- /bin/bash -c taskset -pc 1]' (err=<nil>):
  stdout="" `
  which is a known issue we are separately fighting against.
overriding since this was the only problem there and is surely unrelated to this PR

PR #606, opened 2021-07-14
Bump k8s dependencies to 1.21

cynepco3hahue:
1. Bump k8s dependencies to 1.21
2. Bump PAO dependencies to master
3. Bump PTP dependencies to master
4. Update the code with relevant changes
5. Update cnf-tests Dockerfiles to use golang-1.16, needed because once local and the container has different versions it can fail the build under the container.


SchSeba (commented):
/lgtm

PR #609, opened 2021-07-16
Add missing settings in specs

yrobla:
Reconciled with the latest changes that
i used for an ipv6 cluster deployment.


ztp/source-cluster-crs/cluster-crs.yaml
```diff
@@ -84,6 +85,8 @@ spec:
     enabled: true
   searchCollector:
     enabled: false
+  workManager:
+    enabled: true
```
imiller0, line 89: I think that workManager must default to enabled (my clusters have it enabled although it is not in the applied CR), however I'm happy to have it explicitly defined here.

imiller0 (approved):
/lgtm

PR #610, opened 2021-07-16
Remove provisioning from ClusterDeployment

imiller0:
The provisioning section is only used if the CR does not have a
clusterInstallRef (AgentClusterInstall)

### DIFF
--- a/ztp/source-cluster-crs/cluster-crs.yaml
+++ b/ztp/source-cluster-crs/cluster-crs.yaml
@@ -48,9 +48,9 @@ spec:
           cluster-name: siteconfig.Spec.Clusters.ClusterName
   pullSecretRef:
     name: siteconfig.Spec.PullSecretRef.Name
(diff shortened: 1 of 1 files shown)


yuvalk:
/close
hopefully to stop this from continuously failng

vitus133:
I guess @imiller0 didn't know that component name now must appear in the commit message. I didn't know.
Will open a new PR
/close

vitus133:
[PR 613](https://github.com/openshift-kni/cnf-features-deploy/pull/613)

yuvalk:
> I guess @imiller0 didn't know that component name now must appear in the commit message. I didn't know.
> Will open a new PR
> /close

I've wrote an elaborated explanation (in the PR, commit and on slack)
I encourage you all to install the local commit hook which will catch these before pushing, actually before even committing, by running in your local dev env:
`make install-commit-hooks`

PR #630, opened 2021-07-30
ztp: vendor dependencies for policyGenerator

nishant-parekh:
Vendor dependencies for Policy Generator code to facilitate downstream build.

signed-off-by: nparekh@redhat.com


yuvalk:
IMHO
you dont need your own vendor dir
and you dont need your own go.sum and go.mod
if you'll delete the go.mod and go.sum, it'll use the project ones
then it'll also use the shared vendor dir
we dont need to hold 2 separate sets of external dependencies..

lack:
> IMHO
> you dont need your own vendor dir
> and you dont need your own go.sum and go.mod
> if you'll delete the go.mod and go.sum, it'll use the project ones
> then it'll also use the shared vendor dir
> we dont need to hold 2 separate sets of external dependencies..

I disagree; cnf-tests and ztp/.../policygenerator are two very different projects with very different purposes - I'm not sure what we gain by sharing dependencies.

yuvalk:
> I disagree; cnf-tests and ztp/.../policygenerator are two very different projects with very different purposes - I'm not sure what we gain by sharing dependencies.

all the rest of the tools shares this with no known problem.
it's easier and cleaner to manage that way.
for example fixing security issues that need vendoring update.

and if we'll ever hit a real issue, we can split later.

arguably, the PolicyGen deserves it's own repo (in ACM org?) and then reference it from ztp, then it can be packaged and managed independently. but that's a different discussion that go beyond the scope of this PR.

yuvalk:
This PR commits will need some cleanup. the end result looks fine, but the way there need cleaning

nishant-parekh:
> > I disagree; cnf-tests and ztp/.../policygenerator are two very different projects with very different purposes - I'm not sure what we gain by sharing dependencies.
> 
> all the rest of the tools shares this with no known problem.
> it's easier and cleaner to manage that way.
> for example fixing security issues that need vendoring update.
> 
> and if we'll ever hit a real issue, we can split later.
> 
> arguably, the PolicyGen deserves it's own repo (in ACM org?) and then reference it from ztp, then it can be packaged and managed independently. but that's a different discussion that go beyond the scope of this PR.

I agree that is in the works for 4.10, but for the time being we would want to maintain single source for external dependencies.

nishant-parekh:
> This PR commits will need some cleanup. the end result looks fine, but the way there need cleaning

I have updated the commits for the PR

lack:
> > I disagree; cnf-tests and ztp/.../policygenerator are two very different projects with very different purposes - I'm not sure what we gain by sharing dependencies.
> 
> all the rest of the tools shares this with no known problem.
> it's easier and cleaner to manage that way.
> for example fixing security issues that need vendoring update.
> 
> and if we'll ever hit a real issue, we can split later.

One practical implication today is around what commit prefix one should use when adding a dependency for ZTP.

There's also the matter of the different OWNERS in the two areas, meaning that the ZTP owners can't exclusively approve a ZTP change if that change includes dependencies.

serngawy (changes requested):
I'm not quit sure if this the right thing to do right now. that will affect the pre & post job container image as well as the full test process.  I will mark it as Request change till we finalize our thoughts on it.

lack:
> I'm not quit sure if this the right thing to do right now. that will affect the pre & post job container image as well as the full test process.  I will mark it as Request change till we finalize our thoughts on it. 

Can you elaborate a bit more?  This changes the build process a little (in that it builds from the vendored directories instead of downloading the dependencies each time), but the resulting binary should be identical.

yuvalk:
> I'm not quit sure if this the right thing to do right now. that will affect the pre & post job container image as well as the full test process. I will mark it as Request change till we finalize our thoughts on it.

Building without vendoring is very risky for reproducibility and thus discouraged (and even forbidden for d/s, supported, builds)

nishant-parekh:
> I'm not quit sure if this the right thing to do right now. that will affect the pre & post job container image as well as the full test process. I will mark it as Request change till we finalize our thoughts on it.

@serngawy as @yuvalk mentioned we need to do this to enable PolicyGenerator build downstream, the Policy Generator build is used in the d/s container image used for the pre and post sync jobs. The only change I see in containerFile upstream in copying both the ztp and top level vendor directory and changing the build command to include -mod=vendor, am I missing something?

lack:
> all the rest of the tools shares this with no known problem.

But until today, all the golang source is in cnf-tests and related by the fact that they're for testing and all managed under the same OWNERS user set.

ztp/ has a separate purpose, separate OWNERS, and separate commit tag prefix.

I don't want to fight that hard on this; I'd much rather have this in and move forward, but if you think the maintenance we'll save in having a shared vendor directory offsets the cost we'll incur by having ZTP changes split across two "ownership and commit prefix domains" in this repository, I'll defer to @yuvalk 's wisdom :)

nishant-parekh:
waiting on a go ahead from @vitus133 on this change.

vitus133 (approved):

fromanirh:
please fix the conflicts

ijolliffe (approved):
/lgtm

fromanirh (approved):
/approve
all the changes seems to be in the `ztp/`  subtree, so I don't see any problem from the larger project perspective.

PR #637, opened 2021-08-02
Enforce 'go vet' in both go projects

lack:
- ztp: fix all 'go vet' complaints
- ztp: add go vet hack enforcement script
- cnf-tests: add go vet hack enforcement script
- infra: enforce govet in cnf-tests and ztp-policy-generator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,8 +81,8 @@ golint:
 
 govet:
 	@echo "Running go vet"
-	# Disabling GO111MODULE just for go vet execution
-	GO111MODULE=off go vet github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites...
+	cnf-tests/hack/govet.sh
+	ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/hack/govet.sh
 
 verify-commits:
 	hack/verify-commits.sh
--- a/cnf-tests/hack/common.sh
+++ b/cnf-tests/hack/common.sh
@@ -2,8 +2,7 @@
 
 set -e
 
-pushd .
-cd "$(dirname "$0")/.."
+pushd "$(dirname "$0")/.."
 
 function finish {
     popd
--- a/cnf-tests/hack/govet.sh
+++ b/cnf-tests/hack/govet.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
+echo "INFO: Running govet on cnf-tests/testsuites/..."
+go vet ./testsuites/...
--- a/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/hack/common.sh
+++ b/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/hack/common.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+pushd "$(dirname "$0")/.."
+
+function finish {
+    popd
+}
+trap finish EXIT
+
+GOPATH="${GOPATH:-~/go}"
+export GOFLAGS="${GOFLAGS:-"-mod=vendor"}"
+
+export PATH=$PATH:$GOPATH/bin
+
+mkdir -p _cache
--- a/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/hack/govet.sh
+++ b/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/hack/govet.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+. $(dirname $0)/common.sh
+echo "INFO: Running go vet on ztp policy-generator"
+go vet ./...
--- a/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/utils/utils.go
+++ b/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/utils/utils.go
@@ -65,7 +65,7 @@ type AcmPolicy struct {
 
 type acmPolicySpec struct {
 	RemediationAction string                   `yaml:"remediationAction"`
-	Disabled          bool                     `yaml:"disabled`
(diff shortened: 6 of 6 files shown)


lack:
Proof of this same `go vet` script setup catching an issue in ZTP is exemplified in PR #636

This PR fixes that issue and applies the same scripts.

cnf-tests/hack/common.sh
```diff
@@ -2,8 +2,7 @@
 
 set -e
 
-pushd .
-cd "$(dirname "$0")/.."
+pushd "$(dirname "$0")/.."
```
lack, line 5: @yuvalk : I wasn't sure if this was okay... Was there a reason for doing `pushd .; cd foo` instead of just `pushd foo`?  I find the latter much clearer to debug.

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/hack/common.sh
```diff
@@ -0,0 +1,18 @@
...
+pushd "$(dirname "$0")/.."
+
+function finish {
+    popd
+}
+trap finish EXIT
+
+GOPATH="${GOPATH:-~/go}"
+# not yet, but soon...
+#export GOFLAGS="${GOFLAGS:-"-mod=vendor"}"
```
lack, line 14: @nishant-parekh : this should be uncommented once your vendor PR goes in, I think.

cnf-tests/hack/common.sh
yuvalk: I'm ok with your fix here.
just worried that the commit msg hook didnt catch it.

ah it's cnf-tests/hack/common.sh and not hack/common.sh (which have the same issue)
actually IDK why we have 2.
I'd prefer cnf-tests/hack/* to include `hack/common.sh`
if possible
@fedepaol

yuvalk: hmm actually, that'll require special treat for exactly that pushd :-/

Makefile
```diff
@@ -81,8 +81,8 @@ golint:
 
 govet:
 	@echo "Running go vet"
-	# Disabling GO111MODULE just for go vet execution
-	GO111MODULE=off go vet github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites...
+	cnf-tests/hack/govet.sh
+	ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/hack/govet.sh
```
yuvalk, line 85: why do you hide the ztp scripts dir so deep?

yuvalk: or to that matter the whole policyGenerator

cnf-tests/hack/govet.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
+echo "INFO: Running govet on cnf-tests/testsuites/..."
+go vet ./testsuites/...
```
yuvalk, line 5: that's missing the `GO111MODULE` disabling (??)

Makefile
yuvalk: weirdly, the previous statement here doesn't seems to work (weird, because I'd expect that to cause a CI error)

lack: Not sure.  But I'm not going to change that today :)

cnf-tests/hack/govet.sh
lack: It is, but I wasn't even sure why that was there in the first place.

This seems to run fine.

Makefile
lack: My guess is that for kustomize to find the plugin the executable has to be in the proper directory structure.  Maybe we could move the source out to a more manageable depth and just have a makefile that copies the binary deep into that tree?

@serngawy can you shed some light on this?  Does the source need to be that deep, or just the resulting executable?

cnf-tests/hack/govet.sh
lack: Also the old way seemed broken, as you noticed.

cnf-tests/hack/common.sh
lack: If you did `$(basename ${BASH_SOURCE[-1]})` instead of `$(basename $0)` maybe that would solve it?

lack: On the other hand, maybe having these sub-project hack dirs being completely independent may be wise if/when we repackage either of them separately?

Makefile
serngawy: for the kustomize plugin it needs to follow this directory structure to follow kustomize documentation 
https://kubectl.docs.kubernetes.io/guides/extending_kustomize/

lack: @serngawy : But all the source doesn't need to be there, right?  If we had the source in, say `ztp/PolicyGenerator` with a makefile that copied the executable into `ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator` that would work too, right?

serngawy: yes, the binary that matter.

lack: @serngawy  Is there a list of "We should do this some day" cleanup stuff for the policygenerator?  I'm not going to do that in this PR, obviously.

yuvalk:
/lgtm
I like it but want @fedepaol approval

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/hack/common.sh
nishant-parekh: I think this can be uncommented now.

lack: Exactly why I put the hold on it :)

nishant-parekh (approved):

lack:
@yuvalk and @fedepaol : is this change okay?

cnf-tests/hack/govet.sh
fedepaol: Copying the comment from the other PR, why not sharing a single govet.sh file in the root `hack` folder?

lack:
Closing in favor of #701

PR #638, opened 2021-08-02
tools: Add a ice driver building recipe using SRO

SchSeba:
This commit add all the necesary files to create a recipe and apply a special resource yaml for the SRO to build the ice OOT driver


SchSeba:
close this in favor of https://github.com/openshift-kni/cnf-features-deploy/pull/645

PR #647, opened 2021-08-09
Fix bug in cluster validation findMachineConfigPool function

SchSeba:
Two different variables shouldn't be named the same.
If they are named the same we can't compare between them.

Conflicts:
	validationsuite/cluster/validation.go


mmirecki:
LGTM

PR #649, opened 2021-08-09
ztp: Plumb through per-node 'installerArgs' and 'ignitionConfigOverride'

lack:
These allow per-node customization of the ignition and installer defined
in the SiteConfig CR, through the BareMetalHost annotations to the
AssistedService Agent.

### DIFF
--- a/ztp/ran-site-plan-crd/site-config-crd.yaml
+++ b/ztp/ran-site-plan-crd/site-config-crd.yaml
@@ -252,6 +252,14 @@ spec:
                                 should only be necessary to set this when inspection cannot
                                 automatically determine the profile.
                               type: string
+                            installerArgs:
+                              description: Additional command-line arguments to pass through to the OpenShift Installer for this node.
+                              type: array
+                              items:
+                                type: string
+                            ignitionConfigOverride:
+                              description: Json formatted string containing user overrides for the initial ignition config for this node.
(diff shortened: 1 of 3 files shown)


serngawy (commented):
I'm little confused as the ignitionConfigOverride already  exist in the InfraEnv CR so does the ignitionConfigOverride annotation in BMH override the InfraEnv OR append to it ? 
Also for the InstallerArgs , is it similar/different to the UserData attribute in the BMH ?
Would you add the reference doc or wiki for those annotation ?

lack:
> I'm little confused as the ignitionConfigOverride already exist in the InfraEnv CR so does the ignitionConfigOverride annotation in BMH override the InfraEnv OR append to it ?

Unclear.  But the underlying AI has both, so we should provide both.  I'll try to figure out what the behavior is if you set both.

> Also for the InstallerArgs , is it similar/different to the UserData attribute in the BMH ?

It is a list of arguments that get passed to the installer.  See the description here: https://github.com/openshift/assisted-service/blob/master/config/crd/bases/agent-install.openshift.io_agents.yaml#L82-L84 and docs here: https://github.com/openshift/assisted-service/blob/master/docs/user-guide/install-customization.md#installer-params

> Would you add the reference doc or wiki for those annotation ?

Good idea.  Will do.

PR #656, opened 2021-08-12
ztp: add policyGenTemplate crd and restructure the source crs 

serngawy:


vitus133:
A couple of considerations / ideas
1. This is great we are moving them under the same directory. Maybe it makes sense to keep separation to siteconfig related  and policygen related under this directory? Or there is a plan to unite the CRDs?
2. @imiller0 suggested to name this directory in such a way that it will be clear that this is our reference implementation for the DU profile. How about 'reference-profile'
For example:
├reference-profile
  ->     ├site-config
  ->     └policy-source

3. File naming. The cnf-features-deploy/deployment is using lowercase naming, while we are using CamelCase. If we are planning to reference resources from other repos in the future, does it make sense to try making the naming more uniform?

nishant-parekh:
run `make gofmt ` to fix the issue with `ci/prow/ci` run

PR #657, opened 2021-08-12
Enforce 'go test' in ZTP PolicyGenerator code

lack:
- ztp: add hack script to run 'go test' with the right environment
- infra: Enforce the ztp PolicyGenerator 'go test' passes in ci-job

### DIFF
--- a/cnf-tests/hack/govet.sh
+++ b/cnf-tests/hack/govet.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
+echo "INFO: Running govet on cnf-tests/testsuites/..."
+go vet ./testsuites/...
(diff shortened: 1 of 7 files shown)


serngawy (approved):

cnf-tests/hack/govet.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+. $(dirname "$0")/common.sh
```
fedepaol, line 3: Would it make sense to have one single govet.sh file in the root `hack` folder instead of two?

lack: I'm thinking I might actually redistribute this code a bit - In order to have to not involve 'infra' every time we want to make a change within ztp, I was thinking I'd change the top-level `make job-ci` call `$(MAKE) -c ztp job-ci`, which can then be adapted to do 'go vet' and 'go test' and whatever other checks we want to add in the future.

How do you feel about that?

lack:
Closing this one in favor of #702

PR #659, opened 2021-08-16
ztp: refactor argocd container image

vitus133:
Redesign Argo CD resource hook container image and perform the
associated resource hook software adaptations as follows:
1. Remove dependency on Kustomize
2. Base the container on ubi8-minimal image
3. Use Golang 1.16 as a builder and apply module vendoring
4. Split the container image build into two phases:
  a. Build ZTP base image from ubi8-minimal with additions of
     the latest stable oc and kubectl, add utilities needed
     for resource hook operation (bash, jq, tar etc.),
     add python and python modules required for operation.
  b. On top of that, apply the pre-built policy generator and
     resource hook scripts

/cc @serngawy @imiller0 @yrobla 
/assign @nishant-parekh


ztp/resource-generator/src/watcher.py
```diff
@@ -59,27 +59,25 @@ class PolicyGenWrapper(Logger):
...
-            command = 'kustomize build --enable-alpha-plugins'
-            oneliner_file = 'policyGenerator.yaml'
+            cwd = os.path.join(
+                '/tmp/ztp/ztp-policy-generator',
+                'kustomize/plugin/policyGenerator/v1/policygenerator/')
+            command = [
+                './PolicyGenerator',
+                'dummy_arg',
+                paths[0],
+                '../../../../../../source-cluster-crs',
```
imiller0, line 73: Since you know the path is rooted at /tmp/ztp it might be preferable to simply use /tmp/ztp/source-cluster-crs

imiller0 (approved):

ztp/resource-generator/src/watcher.py
```diff
@@ -59,27 +59,25 @@ class PolicyGenWrapper(Logger):
...
+            dest = '/tmp/ztp'
+            shutil.rmtree(dest, ignore_errors=True)
             shutil.copytree(src, dest)
-            cwd = '/tmp/cnf-features-deploy/ztp/ztp-policy-generator'
-            command = 'kustomize build --enable-alpha-plugins'
-            oneliner_file = 'policyGenerator.yaml'
+            cwd = os.path.join(
+                '/tmp/ztp/ztp-policy-generator',
+                'kustomize/plugin/policyGenerator/v1/policygenerator/')
+            command = [
```
nishant-parekh, line 69: Just curious! How does this run without having kustomize on the container?

nishant-parekh (approved):

ztp/resource-generator/src/watcher.py
vitus133: So PolicyGenerator is just an executable that we happened to invoke as kustomize plugin before. But you can also run it standalone if you provide some dummy argument (see line 71) as argv[1], shifting by that all other arguments to their correct places

ztp/resource-generator/src/watcher.py
vitus133: Yes, thanks for this note. Will do on my next PR. Don't want to mess up with the lgtm label :)

PR #660, opened 2021-08-17
ztp: rename crd directory

vitus133:
The crd directory under the ztp will now contain ran policy crd in addition to ran site crd.
Hence the name must suit both APIs
/assign @serngawy 
/cc @imiller0 @nishant-parekh


vitus133:
/close
Obsoleted by https://github.com/openshift-kni/cnf-features-deploy/pull/663

PR #661, opened 2021-08-17
wip: Argo d2

vitus133:
This is a proposed integration branch for argo cd day 2 operations.
It includes several PRs that we have in flight and will need a rebase after they are merged
/hold
cc @serngawy @nishant-parekh @imiller0


ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/common-ranGen.yaml
```diff
@@ -0,0 +1,55 @@
+---
```
imiller0, line 1: This is a duplicate of ztp/ztp-policy-generator/ranPolicyGenTempExamples/common-ranGen.yaml. We should have only one reference implementation and pull from there.

ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
```
imiller0, line 1: I'm not sure it makes sense to have a kustomization file here. These files are all committed to git where the argo pipeline is subscribed to them. We don't need a kustomization there.

ztp/resource-generator/src/watcher.py
```diff
@@ -56,30 +56,34 @@ def watch_sites(self, rv):
...
+            cwd = os.path.join(
+                '/tmp/ztp/ztp-policy-generator',
+                'kustomize/plugin/policyGenerator/v1/policygenerator/')
+            command = [
+                './PolicyGenerator',
+                'dummy_arg',
+                paths[0],
+                template_dir,
+                paths[1],
+                'true', 'false', 'true']
```
imiller0, line 81: These booleans change between siteconfigs and policygentemplates. For policygentemplates it should be "true false false" and for siteconfigs it should be  true false true

ztp/ztp-policy-generator/ranPolicyGenTempExamples/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+resources:
```
imiller0, line 1: Is this kustomization file needed?

nishant-parekh:
We need to add the `ztp:` prefix to the commit messages to make sure the CI does not fail on that.

vitus133:
/close
Replaced by https://github.com/openshift-kni/cnf-features-deploy/pull/663

PR #666, opened 2021-08-26
ztp: reduce sno monitoring footprint

vitus133:
Adds day-0 extra manifests disabling Grafana and alert manager. Required for SNO CPU footprint.
/assign @imiller0 
/cc @serngawy
/hold until tested

### DIFF
--- a/ztp/source-crs/ReduceMonitoringFootprint.yaml
+++ b/ztp/source-crs/ReduceMonitoringFootprint.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-monitoring-config
+  namespace: openshift-monitoring
+data:
+  config.yaml: |
+    grafana:
+      enabled: false
+    alertmanagerMain:
+      enabled: false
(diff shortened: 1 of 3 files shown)


vitus133:
> I think it would be preferable to make these changes as part of the source-crs and the common PolicyGenTemplate rather than adding them as day-1 install manifests.

Done

imiller0:
The changes look good (pending testing), but can you add this CR to the common or group under ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/. Until we get that moved up to a higher level in the hierarchy, it represents the "source of truth" reference implementation.

ztp/source-crs/ReduceMonitoringFootprint.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-monitoring-config
+  namespace: openshift-monitoring
+data:
+  config.yaml: |
+    grafana:
+      enabled: false
+---
```
imiller0, line 10: PolicyGen requires there to be a single CR per source file do make updates. We should split this into two files to allow the user to override if/as desired.

vitus133:
/unhold
(tested)

vitus133:
> can you add this CR to the common or group under ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/. Until we get that moved up to a higher level in the hierarchy, it represents the "source of truth" reference implementation.

Done

imiller0 (approved):
/lgtm

yuvalk:
Not the same, but related, is the need to create a retention policy for Prometheus data, otherwise it'll keep accumulating, till out of disk space...

PR #669, opened 2021-08-29
ztp: add PTP options for Collumbiaville NICs

vitus133:
Columbiaville NICs require two changes in ptp4l parameters.
This commit adds relevant comments in two places:
1. In source-crs, for the case when CVL NICs are deployed everywhere
2. In group policy template, if CVL NICs are deployed on a group of
   nodes
/assign @imiller0 
/cc @serngawy


ztp/ztp-policy-generator/testPolicyGenTemplate/group-du-sno-ranGen.yaml
```diff
@@ -49,7 +49,7 @@ spec:
           machineconfiguration.openshift.io/role: master
     - fileName: MachineConfigDisableChronyd.yaml
       policyName: "mc-chronyd-policy"
-    - fileName: PtpConfigSlave.yaml
+    - fileName: PtpConfigSlave.yaml  # Change to PtpConfigSlaveCvl.yaml for ColumbiaVille NIC
```
lack, line 52: This change needs to happen in the "real" DU profile :)

lack (changes requested):

ztp/ztp-policy-generator/testPolicyGenTemplate/group-du-sno-ranGen.yaml
vitus133: You mean in this one below?
https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/group-du-sno-ranGen.yaml

lack: Exactly.

vitus133: Done

PR #673, opened 2021-08-31
ztp: tests update for policyGenerator after day2 changes

nishant-parekh:
- reorganized test files splitting constants, getters, assertions & helpers into their own files
- updated test templates per changes made to example templates
- modify policyGen test assertions as per changes made to template
- preserve individual files for generated CRs that can be applied directly to the cluster
- nbde siteconfig test fixes
- alternative implementation for os.Executable for tests

Signed-off-by: nishant-parekh <nparekh@redhat.com>


ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGen/policyBuilder.go
```diff
@@ -41,8 +41,15 @@ func (pbuilder *PolicyBuilder) Build(policyGenTemp utils.PolicyGenTemplate) (map
 			path := policyGenTemp.Metadata.Name + "/" + policyGenTemp.Metadata.Name + "-" + sFile.PolicyName
 			if sFile.PolicyName == "" {
 				// Generate CR without append to policy
-				path := utils.CustomResource + "/" + path
-				policies[path] = resources
+				for _, resource := range resources {
```
serngawy, line 44: I don't think there is a need for loop. if it is multi yaml structure just use the first one. wt do u think ?

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go
```diff
@@ -15,19 +15,33 @@ var sourcePath string
 var tempPath string
 var outPath string
 var stdout bool
+var isTestRun bool
```
serngawy, line 18: I believe we shouldn't let the test flag come from end user. The test mode should be independent

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/utils/filesHandler.go
```diff
@@ -11,10 +11,15 @@ type FilesHandler struct {
 	sourceDir string
 	tempDir   string
 	outDir    string
+	isTestRun bool `default:"false"`
```
serngawy, line 14: same, fileHandler tests should be outside of fileHandler

serngawy (changes requested):
Better to consider policygen lib  as black box you don't give it a flag to tell I will make Test. You give input and check the expected output.

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGen/policyBuilder.go
nishant-parekh: The reason for this loop was to create a single file per resource instead of one file for all the resources within the template, this section deals with the directly applied CRs, and as discussed we wanted to keep them separate.

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go
nishant-parekh: I agree I dont want to add this flag the only issues is the `os.Executable()` call in fileHandler.go that I need to change to a os.getwd() for the test runs. The rest of the changes are just a way to enable that if you have a different suggestion I would be happy to change this approach. 

I considered mocking the fileHandler.go completely for the tests but the policyGenerator doesnt reference a fileHandler as a class variable rather it is encapsulated inside policyBuilder.

serngawy: may consider failing safe, if reading from os.Excutable() fail then read from os.getWd()

nishant-parekh: ohh that sounds reasonable and will work with the actual run as well! I can make that change! thanks for the suggestion! :)

nishant-parekh: updated

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/utils/filesHandler.go
nishant-parekh: I have removed this per our discussion

nishant-parekh:
> Better to consider policygen lib as black box you don't give it a flag to tell I will make Test. You give input and check the expected output.

I have changed the approach and added a failsafe within the `readResourceFile` method within `fileHandler.go`

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/utils/filesHandler.go
```diff
@@ -16,7 +16,6 @@ type FilesHandler struct {
 func NewFilesHandler(sourceDir string, tempDir string, outDir string) *FilesHandler {
 	return &FilesHandler{sourceDir: sourceDir, tempDir: tempDir, outDir: outDir}
 }
-
 func (fHandler *FilesHandler) WriteFile(filePath string, content []byte) error {
```
serngawy, line 19: add line between methods

serngawy (commented):
small comments

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGen/policyBuilder.go
nishant-parekh: I think we need to separate this CRs into individual files as they will be applied independently on the cluster using `oc apply <fileName>.yaml` . Based on our discussion in the working session we want to keep this implementation so as to work through the deployment on a file by file basis? @imiller0 what do you think?

nishant-parekh: I have removed the loop from the logic replaced with a size check and processing it as an array of size 1

serngawy (approved):

PR #681, opened 2021-09-05
Bug 2000726: ztp: fix rv data name and path

vitus133:
Fixes an issue discovered in scalability tests, where the
post-sync resource hook was failing due to the resource version being
expired.
The reason - the resource version number post-sync was comparing to
was initialized from config map resource version instead of its data.
/assign @imiller0 
/cc @serngawy


imiller0 (approved):
/lgtm

PR #685, opened 2021-09-08
Add bmh role annotation

yrobla:


serngawy:
this PR is missing the Nodes.Role definition of in siteConfig CRD

PR #687, opened 2021-09-08
ztp: Add SNO accelerated startup services

lack:
These systemd services and the associated script take a set of  system
services (which would normally be restricted by a PAO performance
profile isolated cores definition), and remove any CPU restrictions at
boot and at shutdown for 10 minutes. These are windows of time when any
potential CPU starvation of customer workloads would not be an issue,
and at which time the unpinned system services allow faster reboot and
recovery times.


ztp/source-crs/MachineConfigAcceleratedStartup.yaml
```diff
@@ -0,0 +1,88 @@
...
+          [Service]
+          Type=simple
+          ExecStart=/usr/local/bin/accelerated-container-startup.sh
+
+          [Install]
+          WantedBy=shutdown.target reboot.target halt.target
+        enabled: true
+        name: accelerated-container-shutdown.service
+---
+apiVersion: machineconfiguration.openshift.io/v1
```
serngawy, line 46: There is no need to duplicate the MC under source-crs for Worker & Master the l`abel: machineconfiguration.openshift.io/role: worker|master  ` code be set in the PGT

lack: Is there an example of this somewhere?

serngawy: similar to the sctp MC here
https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/ztp-policy-generator/testPolicyGenTemplate/group-du-sno-ranGen.yaml#L45

lack: Awesome.  I think I've done a much better job this time :)

serngawy (approved):

PR #688, opened 2021-09-09
Nummasters

yrobla:


ztp/ran-crd/site-config-crd.yaml
```diff
@@ -84,6 +84,13 @@ spec:
                         enum:
                           - standard
                           - sno
+                      numMasters:
```
imiller0, line 87: Do we need to expose this as an input in the CRD, or can PolicyGen set the value automatically based on either clusterType or the length of the nodes array?

serngawy: AI lets u define number of masters 3, 5, ...etc I guess we should follow the same

serngawy (approved):

PR #689, opened 2021-09-09
ztp: Split ClusterLogOperGroup into separate files

lack:
The policy generator doesn't support creating un-wrapped CRs if any of
the source CRs contain multiple YAML documents per file, so we must keep
a 1:1 document-to-file relationship throughout our source tree.

### DIFF
--- a/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGen/policyBuilder.go
+++ b/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGen/policyBuilder.go
@@ -3,11 +3,13 @@ package policyGen
 import (
 	"bytes"
 	"errors"
-	utils "github.com/openshift-kni/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/utils"
-	yaml "gopkg.in/yaml.v3"
 	"io"
+	"path/filepath"
 	"reflect"
 	"strings"
+
+	utils "github.com/openshift-kni/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/utils"
+	yaml "gopkg.in/yaml.v3"
 )
 
 type PolicyBuilder struct {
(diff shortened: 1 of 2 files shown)


ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/common-ranGen.yaml
```diff
@@ -40,6 +40,10 @@ spec:
       policyName: "pao-sub-policy"
       spec:
         image: quay.io/openshift-kni/performance-addon-operator-index:4.8-snapshot
+    - fileName: ClusterLoggingNamespace.yaml
+      policyName: "log-sub-policy"
+    - fileName: ClusterLoggingNamespace.yaml
+      policyName: "log-sub-policy"
```
serngawy, line 46: why this duplication ?

serngawy (commented):
It is not clear the reason for separating them into 2 files. I don't think there is a need to change the Namespace or OperatorGroup

ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/common-ranGen.yaml
lack: Oops, that's a typo :)

lack:
> It is not clear the reason for separating them into 2 files. I don't think there is a need to change the Namespace or OperatorGroup

Without this change, trying to generate the non-wrapped CRs fails as follows:

```
$ sed -i 's/policyName: ".*"/policyName: ""/' testPolicyGenTemplate/*.yaml
$ XDG_CONFIG_HOME=./ kustomize build --enable-alpha-plugins
panic: Update spec/data of multiple yamls structure in same file ClusterLogOperGroup.yaml not allowed. Instead separate them in multiple files

goroutine 1 [running]:
main.InitiatePolicyGen(0x7ffee4024146, 0x17, 0x7ffee402415e, 0xd, 0x7ffee402416c, 0x5, 0xc00002a001)
        /home/jramsay/src/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go:72 +0x1167
main.main()
        /home/jramsay/src/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go:25 +0x154
Error: failure in plugin configured via /tmp/kust-plugin-config-854908569; exit status 2: exit status 2
```

serngawy:
> > It is not clear the reason for separating them into 2 files. I don't think there is a need to change the Namespace or OperatorGroup
> 
> Without this change, trying to generate the non-wrapped CRs fails as follows:
> 
> ```
> $ sed -i 's/policyName: ".*"/policyName: ""/' testPolicyGenTemplate/*.yaml
> $ XDG_CONFIG_HOME=./ kustomize build --enable-alpha-plugins
> panic: Update spec/data of multiple yamls structure in same file ClusterLogOperGroup.yaml not allowed. Instead separate them in multiple files
> 
> goroutine 1 [running]:
> main.InitiatePolicyGen(0x7ffee4024146, 0x17, 0x7ffee402415e, 0xd, 0x7ffee402416c, 0x5, 0xc00002a001)
>         /home/jramsay/src/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go:72 +0x1167
> main.main()
>         /home/jramsay/src/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go:25 +0x154
> Error: failure in plugin configured via /tmp/kust-plugin-config-854908569; exit status 2: exit status 2
> ```

That is normal, its not an error. Do we have a case that we need to change NS or OperatorGroup of Cluster log ?

lack:
> > > It is not clear the reason for separating them into 2 files. I don't think there is a need to change the Namespace or OperatorGroup
> > 
> > 
> > Without this change, trying to generate the non-wrapped CRs fails as follows:
> > ```
> > $ sed -i 's/policyName: ".*"/policyName: ""/' testPolicyGenTemplate/*.yaml
> > $ XDG_CONFIG_HOME=./ kustomize build --enable-alpha-plugins
> > panic: Update spec/data of multiple yamls structure in same file ClusterLogOperGroup.yaml not allowed. Instead separate them in multiple files
> > 
> > goroutine 1 [running]:
> > main.InitiatePolicyGen(0x7ffee4024146, 0x17, 0x7ffee402415e, 0xd, 0x7ffee402416c, 0x5, 0xc00002a001)
> >         /home/jramsay/src/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go:72 +0x1167
> > main.main()
> >         /home/jramsay/src/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go:25 +0x154
> > Error: failure in plugin configured via /tmp/kust-plugin-config-854908569; exit status 2: exit status 2
> > ```
> 
> That is normal, its not an error. Do we have a case that we need to change NS or OperatorGroup of Cluster log ?

But the message says both "panic" and "error", and it doesn't actually generate all the CRs.  It stops part-way through the 'common' directory, and doesn't generate everything.

PR #690, opened 2021-09-10
ztp: Fix raw CR generation for multi-entry sources

lack:
This change allows multi-entry yaml files to pass through when creating
unwrapped raw CRs.

Signed-off-by: Jim Ramsay <jramsay@redhat.com>

PR #692, opened 2021-09-10
ztp/error flow fixup

lack:
- ztp: Fix raw CR generation for multi-entry sources
- ztp: Skip hidden files when scanning for incoming CRs
- ztp: Rework top-level error handling and log output


lack:
Note: This also contains #690 so the diff will make more sense once that's merged.

PR #694, opened 2021-09-13
ztp: Fix mc-accelerated-startup-policy-worker-du  policy name more than 63 chars

serngawy:
-  Fix increase policy name more than 63 chars     


lack:
I'm not usually a fan of squashing unrelated changes into a single commit - Would you mind splitting them up (it's fine if they're in the same PR)

lack:
Still /lgtm

PR #696, opened 2021-09-14
ztp: support networkType parametrization

mvazquezc:
Users cannot choose which SDN plugin gets deployed via the SiteConfigs, they will always get OpenShiftSDN for IPv4 clusters and OVNKubernetes for IPv6 clusters.

This PR adds a new parameter inside the Cluster definitions named `networkType` that can be set to either "OVNKubernetes" or "OpenShiftSDN". If it's not set, by default OVNKubernetes will be used.

Example:

~~~yaml
<omitted>
  clusters:
  - clusterName: "mycluster1"
    clusterType: "sno"
    clusterProfile: "du"
    clusterLabels:
      group-du-sno: ""
      common: true
      sites : "westford-lab"
    networkType: "OVNKubernetes"
<omitted>
~~~

I couldn't do something like this in the ACI yaml, it seems not possible to add a spec value and a string in the same line, that's why this modifies the CR inside the `getClusterCRs` function.

~~~
apiVersion: extensions.hive.openshift.io/v1beta1
kind: AgentClusterInstall
metadata:
  name: siteconfig.Spec.Clusters.ClusterName
  namespace: siteconfig.Spec.Clusters.ClusterName
  annotations: 
    agent-install.openshift.io/install-config-overrides: '{"networking":{"networkType":"siteconfig.Spec.Clusters.NetworkType"}}'
spec:
<omitted>
~~~

### DIFF
--- a/ztp/ran-crd/site-config-crd.yaml
+++ b/ztp/ran-crd/site-config-crd.yaml
@@ -102,6 +102,13 @@ spec:
                           - du
                           - cu
                           - cu-du
+                      networkType:
+                        description: NetworkType is the SDN plugin that will be deployed on the Cluster.
+                        type: string
+                        default: OVNKubernetes
+                        enum:
+                          - OVNKubernetes
+                          - OpenShiftSDN
                       clusterLabels:
                         description: |
                           clusterLabels specify the the day2 configuration policies that will get applied on the cluster. The following labels example should be included in order to match the policy generatroe policies rules.
--- a/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
+++ b/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
@@ -53,6 +53,14 @@ func (scbuilder *SiteConfigBuilder) Build(siteConfigTemp SiteConfig) (map[string
 		if cluster.ClusterName == "" {
 			return clustersCRs, errors.New("Error: Missing cluster name at site " + siteConfigTemp.Metadata.Name)
 		}
+		if cluster.NetworkType == "" {
+			cluster.NetworkType = "OVNKubernetes"
+			siteConfigTemp.Spec.Clusters[id].NetworkType = "OVNKubernetes"
+		}
+		if cluster.NetworkType != "OpenShiftSDN" && cluster.NetworkType != "OVNKubernetes" {
+			return clustersCRs, errors.New("Error: networkType must be either OpenShiftSDN or OVNKubernetes " + siteConfigTemp.Metadata.Name)
+		}
+		siteConfigTemp.Spec.Clusters[id].NetworkType = "{\"networking\":{\"networkType\":\"" + cluster.NetworkType + "\"}}"
 		clusterValue, err := scbuilder.getClusterCRs(id, siteConfigTemp)
 		if err != nil {
 			return clustersCRs, err
--- a/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/resources/cluster-crs.yaml
+++ b/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/resources/cluster-crs.yaml
@@ -10,6 +10,8 @@ kind: AgentClusterInstall
 metadata:
   name: siteconfig.Spec.Clusters.ClusterName
   namespace: siteconfig.Spec.Clusters.ClusterName
+  annotations:
+    agent-install.openshift.io/install-config-overrides: siteconfig.Spec.Clusters.NetworkType
 spec:
   clusterDeploymentRef:
     name: siteconfig.Spec.Clusters.ClusterName
--- a/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfig.go
+++ b/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfig.go
@@ -99,6 +99,7 @@ type Clusters struct {
 	NumMasters             uint8             `yaml:"numMasters"`
 	ClusterProfile         string            `yaml:"clusterProfile"`
 	ClusterLabels          map[string]string `yaml:"clusterLabels"`
+	NetworkType            string            `yaml:"networkType"`
 	ClusterNetwork         []ClusterNetwork  `yaml:"clusterNetwork"`
 	IgnitionConfigOverride string            `yaml:"ignitionConfigOverride"`
 	DiskEncryption         DiskEncryption    `yaml:"diskEncryption"`


ztp/ran-crd/site-config-crd.yaml
```diff
@@ -102,6 +102,9 @@ spec:
                           - du
                           - cu
                           - cu-du
+                      networkType:
```
serngawy, line 105: it will better to have this as enum values OpenShiftSDN & OVNKubernetes with default value.

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
```diff
@@ -69,6 +76,17 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
...
 		mapSourceCR := cr.(map[string]interface{})
 
+		if mapSourceCR["kind"] == "AgentClusterInstall" {
+			networkType := siteConfigTemp.Spec.Clusters[clusterId].NetworkType
+			crMetadata := mapSourceCR["metadata"].(map[string]interface{})
+			networkTypeOverride := "{\"networking\":{\"networkType\":\"" + networkType + "\"}}"
+			annotations := make(map[string]interface{})
+			// This overrides existing annotations, at this point the ACI definition
+			// in the cluster-crs.yaml file does not have any
+			annotations["agent-install.openshift.io/install-config-overrides"] = networkTypeOverride
```
serngawy, line 86: why not adding the annotation with path to the networkType in cluster-crs.yaml-> AgentClusterInstall.
should be something like 
`  annotation:
    agent-install.openshift.io/install-config-overrides: siteconfig.Spec.Clusters.NetworkType
 `

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
```diff
@@ -53,6 +53,13 @@ func (scbuilder *SiteConfigBuilder) Build(siteConfigTemp SiteConfig) (map[string
 		if cluster.ClusterName == "" {
 			return clustersCRs, errors.New("Error: Missing cluster name at site " + siteConfigTemp.Metadata.Name)
 		}
+		if cluster.NetworkType == "" {
```
serngawy, line 56: Having the CRD definition as Enum with default value will avoid those checks.

serngawy (changes requested):
Thank Mario for the work. Couple of comments AND this work is for ztp so the commit message must start with **ztp:** . You can run in the root dir `$ make verify-commits` to be sure the ci jobs will not fail the changes before you push the code.

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
mvazquezc: I couldn't do something like that in the ACI yaml, it seems not possible to add a spec value and a string in the same line, that's why this modifies the CR inside the getClusterCRs function.

ztp/ran-crd/site-config-crd.yaml
mvazquezc: Changed the CRD with enum and default vlaues.

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
mvazquezc: Removed checks since those are now at CRD level.

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
serngawy: What I mean is ; We set the `siteConfigTemp.Spec.Clusters[clusterId].NetworkType = "{\"networking\":{\"networkType\":\"" + networkType + "\"}}" ` then as the annotation in ACI yaml have been defined as above it will just set the NetworkType value. 
AND please squash the commits into 1 commit and set the commit message start with **ztp:** the ci/proc/ci Job will continue to fail if u didn't

lack:
I don't think the CRD alone can do this.  I tried a manual policy generation with my CRD omitting the new `networkType` field entirely, and I got this:
```
apiVersion: extensions.hive.openshift.io/v1beta1
kind: AgentClusterInstall
metadata:
    annotations:
        agent-install.openshift.io/install-config-overrides: '{"networking":{"networkType":""}}'
```
I was expecting it to default to `OVNKubernetes`, but it didn't...

lack:
I also noticed there are no unit tests defined for this change - We should probably start making some degree of unit testing manditory?

serngawy:
@mvazquezc any chance to update this PR soon

mvazquezc:
> I don't think the CRD alone can enforce the default value. I tried a manual policy generation with my CRD omitting the new `networkType` field entirely, and I got this:
> 
> ```
> apiVersion: extensions.hive.openshift.io/v1beta1
> kind: AgentClusterInstall
> metadata:
>     annotations:
>         agent-install.openshift.io/install-config-overrides: '{"networking":{"networkType":""}}'
> ```
> 
> I was expecting it to default to `OVNKubernetes`, but it didn't...

Hey @lack,

That's expected. The default value will be added when the object gets processed by the API Server. If we want to support default values/value validation when using the CLI tool then I'd need to add the validation/defaulting code back. Is that required?

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
mvazquezc: That did not work, this is what I got in the generated file:

~~~yaml
kind: AgentClusterInstall
metadata:
    annotations:
        agent-install.openshift.io/install-config-overrides: siteConfigTemp.Spec.Clusters[clusterId].NetworkType = "{\"networking\":{\"networkType\":\"" + siteConfigTemp.Spec.Clusters[clusterId].NetworkType + "\"}}"

~~~

imiller0:
@mvazquezc Can you update the commit message for both commits (and consider squashing them) to be prefixed with "ztp: ". The CI lanes check for all commit messages containing an indication of the component (eg ztp:).

mvazquezc:
> @mvazquezc Can you please update the commit message for both commits (and consider squashing them) to be prefixed with "ztp: ". The CI lanes check for all commit messages containing an indication of the component (eg ztp:).

Done!

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
serngawy: The cluser-crs.yaml->AgentClusterInstall should as below
```
kind: AgentClusterInstall
  metadata:
    annotations:
      agent-install.openshift.io/install-config-overrides: siteConfig.Spec.Clusters.NetworkType 
```
 In the go implementation siteConfigBuilder.go Line56  as u did the changes instead of the checks and create the annotation map we can set the networkType as below
 
 `cluster.NetworkType = "{\"networking\":{\"networkType\":\"" + cluster.NetworkType + "\"}}"`
 
 this way when we append a new value to the annotation Map we will not need to re-create the annotation will just append to the cluster-crs.yaml and the crd definition.

Sorry for being not clear in the comment AND the CI job complain about go code style format. To avoid that in the repo root dir u can check the code format by run `$ make gofmt` to make sure that there is no missing space or style.

mvazquezc: Hey @serngawy,

It's not working, I believe mixing text (agent-install.openshift.io/install-config-overrides:) and var values (siteConfig.Spec.Clusters.NetworkType) in the same line on the template doesn't quite work.

This is what I get:

kind: AgentClusterInstall
metadata:
    annotations:
        agent-install.openshift.io/install-config-overrides: siteConfig.Spec.Clusters.NetworkType

mvazquezc:
Changed code to use @serngawy approach for adding the annotation.

PR #699, opened 2021-09-15
ztp: Detect steady-state for accelerated startup

lack:
This mechanism is configurable, and can detect "steady-state" based on:
- STEADY_STATE_MINIMUM: A minimum pod count
- STEADY_STATE_THRESHOLD: A threshold, either in absolute pod count or
  percentage-change compared to the last poll interval (10s)
- STEADY_STATE_WINDOW: A window of time

If the number of running pods according to 'crictl ps' is higher than
STEADY_STATE_MINIMUM and stays within STEADY_STATE_THRESHOLD for the duration
of STEADY_STATE_WINDOW, we will consider the system to have reached
"steady-state", and exit the 10min timer early.

By default, startup will set up a steady-state check to exit either after 10
minutes (the default) or when the pod count is >40 and stays within 2% for 120s


ztp/source-crs/MachineConfigAcceleratedStartup.yaml
```diff
@@ -37,6 +59,22 @@ spec:
...
+          #  4  - absolute pod count (+/-)
+          #  4% - percent change (+/-)
+          #  -1 - disable the steady-state check
+          Environment=STEADY_STATE_THRESHOLD=-1
+
+          # Steady-state window = 60s
+          # If the running pod count stays within the given threshold for this time
+          # period, return CPU utilization to normal before the maximum wait time has
+          # expires
+          Environment=STEADY_STATE_WINDOW=60
```
imiller0, line 76: What does the steady state window do if the threshold is set to -1?

lack: It's in the comment right above [Line 69](https://github.com/openshift-kni/cnf-features-deploy/pull/699/files/ce1a4ae9d1a944de88dfe39d0a476a0cbe619033#diff-1abecdd36031dcbb754c51b436a04ea54096047e5992dfc007a903ec74f70b43R69)
```
          #  -1 - disable the steady-state check
```

lack: So the whole check is disabled, and this is ignorend.  It's just in the file here for boiler-plate if you ever set the threshold to >=0

imiller0: Ah, so just there for reference.

serngawy (approved):

PR #700, opened 2021-09-16
ztp/make ci job

lack:
- ztp: Add ztp-specific 'make ci-job' placeholder
- infra: Call ztp's 'make ci-job' as part of top-level 'make ci-job'


lack:
/cc @fedepaol @yuvalk 

Here's an alternative approach to #637 and #657: with this basis in place, the rest can be done entirely within the ZTP area.

PR #701, opened 2021-09-16
ztp/add govet

lack:
- ztp: fix all 'go vet' complaints
- ztp: Call 'go vet' as part of 'make ci-job'

/hold and obly choose this on or #637, not both


lack:
/nohold ready to go in now!  #637 is closed in favor of this mechanism.

nishant-parekh (approved):

PR #705, opened 2021-09-16
ztp: Increase the gitOps controller pod memory

serngawy:
Ref to the issue is here https://issues.redhat.com/browse/GITOPS-1255
AND https://github.com/argoproj/argo-cd/issues/5505


imiller0 (approved):
/lgtm
Confirmed with @serngawy that the resource values are all default values other than the ones specifically marked with comments.

PR #706, opened 2021-09-16
Update multi yaml file

eggfoobar:
Policy generator looks for single document `yaml` files when parsing `PolicyGenTemplate`, this pr splits the namespace creation in the `ClusterLogOperGroup.yaml` to it's own file `ClusterLogOperGroupNS.yaml` to fix errors that arise when policy gen renders the output

### DIFF
--- a/ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/common-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/common-ranGen.yaml
@@ -30,7 +30,7 @@ spec:
     - fileName: PaoSubscription.yaml
       policyName: "pao-sub-policy"
       spec:
-      # Changing the channel value will upgrade/downgrade the operator installed version.
+        # Changing the channel value will upgrade/downgrade the operator installed version.
         channel: "4.8"
     - fileName: PaoSubscriptionNS.yaml
       policyName: "pao-sub-policy"
@@ -73,9 +73,9 @@ spec:
       policyName: "disc-icsp-policy"
       spec:
         repositoryDigestMirrors:
-        - mirrors:
-          - registry.redhat.com:5000
-          source: registry.redhat.io
+          - mirrors:
+              - registry.redhat.com:5000
+            source: registry.redhat.io
     - fileName: OperatorHub.yaml
       policyName: "oper-hub-policy"
     - fileName: ReduceMonitoringFootprint.yaml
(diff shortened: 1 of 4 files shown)


lack:
@eggfoobar Thanks for the PR!

Can you give an example of the error you're seeing because of the multi-document yaml?

We recently had an issue similar to this in where this same multi-document yaml broke non-acm-wrapped policy generation, but this was fixed in #690.

eggfoobar:
Hey @lack, absolutely,

I am using the PolicyGen tool from this container image `quay.io/redhat_emp1/ztp-site-generator:latest` , if I include the `ClusterLogOperGroup.yaml` in the policy gen template `cr`, 

```yaml
apiVersion: ran.openshift.io/v1
kind: PolicyGenTemplate
metadata:
  name: "common"
  namespace: "common"
spec:
  bindingRules:
    common: "true"
  sourceFiles:
....
    - fileName: ClusterLogOperGroup.yaml
      policyName: "log-sub-policy"
....
```

I see this error crop up, that seems to be triggered by this validation check [L118](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGen/policyBuilder.go#L118)
```sh
Update spec/data of multiple yamls structure in same file ClusterLogOperGroup.yaml not allowed. Instead separate them in multiple files
panic: Update spec/data of multiple yamls structure in same file ClusterLogOperGroup.yaml not allowed. Instead separate them in multiple files

goroutine 1 [running]:
main.InitiatePolicyGen(0x7fffc2bfee9c, 0xf, 0x7fffc2bfeeac, 0x1d, 0x7fffc2bfeeca, 0x9, 0xc000094000)
        /go/src/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go:72 +0x1167
main.main()
        /go/src/cnf-features-deploy/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/policyGenerator.go:25 +0x154
```

lack:
There were two validation checks with identical error messages.

One would occur if you're doing raw CR generation (not wrapped in an ACM policy), any time there was a compound yaml document. This was fixed in #690, but depending on when your container was built, it may not yet contain the fix.  I'll see if I can figure out which git commit corresponds to that image you posted.

The other, on L118 as you saw, should only hit if you're doing a policy-wrapped generation, and then only if you're trying to overlay the content to alter one of the compound yaml files.

Are you doing the raw CR generation, or the policy-wrapped generation?

eggfoobar:
Thanks @lack !

I was running raw CR generation and as you pointed out it was already patched, I was using the image directly which does not have the latest changes. I built it locally and tried it again and the error was no longer present.

Closing the PR

lack:
Glad I could help!

PR #707, opened 2021-09-20
ztp: Perform test policy generation for 'make ci-job'

lack:
This ensures that the example policy xan generate both policy-wrapped
CRs and non-wrapped CRs.

Does not do any chexk for correctness; just success/fail

### DIFF
--- a/ztp/ztp-policy-generator/Makefile
+++ b/ztp/ztp-policy-generator/Makefile
@@ -1,0 +1,11 @@
+.PHONY: build generate-policy generate-crs
+build:
+	./hack/build.sh
+
+generate-policy:
+	./hack/generate-policy.sh
+
+generate-crs:
+	./hack/generate-crs.sh
+
+generate-all: generate-policy generate-crs
--- a/ztp/ztp-policy-generator/hack/common.sh
+++ b/ztp/ztp-policy-generator/hack/common.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+set -e
+
+pushd "$(dirname "$0")/.." >&2
+
+function finish {
+    popd >&2
+}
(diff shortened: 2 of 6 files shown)


ztp/ztp-policy-generator/Makefile
```diff
@@ -0,0 +1,11 @@
+.PHONY: build generate-policy generate-crs
+build:
+	./hack/build.sh
+
+generate-policy:
```
serngawy, line 5: I don't think using the command to generate policies or crs is the right way, the PGT deiced if it will generate policy or just CR

ztp/ztp-policy-generator/hack/common.sh
```diff
@@ -0,0 +1,39 @@
...
+
+GOPATH="${GOPATH:-~/go}"
+export GOFLAGS="${GOFLAGS:-"-mod=vendor"}"
+
+export PATH=$PATH:$GOPATH/bin
+
+PLUGINPATH=kustomize/plugin/policyGenerator/v1/policygenerator/
+PLUGINBIN=PolicyGenerator
+
+getKustomize() {
```
serngawy, line 20: I don't think we have to force install Kustomize, we just need to state kustomize is requirement if policyGen will be used as kustomize plugin. In fact we currently use the policyGen in the container image without kustomize.

serngawy (changes requested):

ztp/ztp-policy-generator/hack/common.sh
lack: We need to force-install it because the Github container that the test runs in doesn't have it.

lack:
After some offline discuss, I'm closing this in favor of a different approach

PR #708, opened 2021-09-20
cnf-tests: DRAFT (Add vhost-net tests)

mmirecki:


cnf-tests/testsuites/e2esuite/dpdk/dpdk.go
```diff
@@ -173,6 +173,81 @@ var _ = Describe("dpdk", func() {
 		}
 	})
 
+	Context("DRAFT - vhost-net", func() {
```
david-marchand, line 176: From the commands below, you are testing the tap driver.

vhost-net is not involved.
If you want to test virtio-user (which uses the vhost-net backend with a tap), you'll have to call:
--vdev virtio_user0,iface=tap23,path=/dev/vhost-net

PR #709, opened 2021-09-20
Check ztp/source-cr MachineConfig files for consistency in ci-job

lack:
- vendor: Add github.com/lack/mcmaker to check ZTP source-cr content
- infra: Add github.com/lack/mcmaker to the CI image
- ztp: Enable source-crs-builder checks for ztp ci-job
- ztp: Add the MachineConfigAcceleratedStartup constituent parts to source control
- ztp: Add more robust logic to accelerated-startup CPUset detection


ztp/source-crs/MachineConfigAcceleratedStartup.yaml
```diff
@@ -11,7 +11,7 @@ spec:
...
...ZHlTdGF0ZVRpbWUgLWd0IDAgXV07IHRoZW4KICAgICAgICBlY2hvICJSZXNldHRpbmcgc3RlYWR5LXN0YXRlIHRpbWVyIgogICAgICAgIHN0ZWFkeVN0YXRlVGltZT0wCiAgICAgIGZpCiAgICBmaQogICAgbGFzdENjb3VudD0kY2NvdW50CiAgZG9uZQogIGxvZ2dlciAiUmVjb3Zlcnk6IFJlY292ZXJ5IENvbXBsZXRlIFRpbWVvdXQiCn0KCm1haW4oKSB7CiAgaWYgISB1bnJlc3RyaWN0ZWRDcHVzZXQgPiYvZGV2L251bGw7IHRoZW4KICAgIGxvZ2dlciAiUmVjb3Zlcnk6IE5vIHVucmVzdHJpY3RlZCBDcHVzZXQgY291bGQgYmUgZGV0ZWN0ZWQiCiAgICByZXR1cm4gMQogIGZpCgogIGlmICEgcmVzdHJpY3RlZENwdXNldCA+Ji9kZXYvbnVsbDsgdGhlbgogICAgbG9nZ2VyICJSZWNvdmVyeTogTm8gcmVzdHJpY3RlZCBDcHVzZXQgaGFzIGJlZW4gY29uZmlndXJlZC4gIFdlIGFyZSBhbHJlYWR5IHJ1bm5pbmcgdW5yZXN0cmljdGVkLiIKICAgIHJldHVybiAwCiAgZmkKCiAgIyBFbnN1cmUgd2UgcmVzZXQgdGhlIENQVSBhZmZpbml0eSB3aGVuIHdlIGV4aXQgdGhpcyBzY3JpcHQgZm9yIGFueSByZWFzb24KICAjIFRoaXMgd2F5IGVpdGhlciBhZnRlciB0aGUgdGltZXIgZXhwaXJlcyBvciBhZnRlciB0aGUgcHJvY2VzcyBpcyBpbnRlcnJ1cHRlZAogICMgdmlhIF5DIG9yIFNJR1RFUk0sIHdlIHJldHVybiB0aGluZ3MgYmFjayB0byB0aGUgd2F5IHRoZXkgc2hvdWxkIGJlLgogIHRyYXAgc2V0UmVzdHJpY3RlZCBFWElUCgogIGxvZ2dlciAiUmVjb3Zlcnk6IFJlY292ZXJ5IE1vZGUgU3RhcnRpbmciCiAgc2V0VW5yZXN0cmljdGVkCiAgd2FpdEZvclJlYWR5Cn0KCmlmIFtbICIke0JBU0hfU09VUkNFWzBdfSIgPSAiJHswfSIgXV07IHRoZW4KICBtYWluICIke0B9IgogIGV4aXQgJD8KZmkK
```
imiller0, line 14: With the builder approach above is this file auto-generated now?

ztp/source-crs-builder/MachineConfigAcceleratedStartup/accelerated-container-startup.sh
```diff
@@ -0,0 +1,198 @@
...
+STEADY_STATE_MINIMUM=${STEADY_STATE_MINIMUM:-0}
+
+#######################################################
+
+unrestrictedCpuset() {
+  if [[ ! -e /var/lib/kubelet/cpu_manager_state ]]; then
+    # use all the cpus if kubelet is not configured yet
+    cat /sys/fs/cgroup/cpuset/cpuset.cpus
+  else
+    jq -r '.defaultCpuSet' </var/lib/kubelet/cpu_manager_state
```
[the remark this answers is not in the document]
lack: Good idea!  Will do.

ztp/source-crs/MachineConfigAcceleratedStartup.yaml
lack: It is!  This corresponds exactly to the contents of `accelerated-container-startup.sh`, and this is generated by running 'make' in the `source-crs-builder` directory.

What's more, it is also re-checked as part of the GitHub `ci-job` makefile target, so if you update the script and forget to run 'make' to keep this rendered file in sync, the GitHub `/ci/prow/ci` lane will fail!

I'll demo this when I make the change you suggest [above](https://github.com/openshift-kni/cnf-features-deploy/pull/709/files#r712966685)

ztp/source-crs-builder/MachineConfigAcceleratedStartup/accelerated-container-startup.service
```diff
@@ -0,0 +1,31 @@
...
+
+# Maximum wait time is 600s = 10m:
+Environment=MAXIMUM_WAIT_TIME=600
+
+# Steady-state threshold = 2%
+# Allowed values:
+#  4  - absolute pod count (+/-)
+#  4% - percent change (+/-)
+#  -1 - disable the steady-state check
+Environment=STEADY_STATE_THRESHOLD=2%%
```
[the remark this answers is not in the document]
lack: Yes, in systemd units you need to escape `%` because it's used for "specifiers"

See "table 4" in https://www.freedesktop.org/software/systemd/man/systemd.unit.html

ztp/source-crs/MachineConfigAcceleratedStartup.yaml
lack: From the failed CI logs, when I changed the shell script but "forgot" to update the resulting yaml:

```
Checking in-source ../source-crs/MachineConfigAcceleratedStartup.yaml for consistency ... Failed
```

ztp/source-crs-builder/MachineConfigAcceleratedStartup/accelerated-container-startup.service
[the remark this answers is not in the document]
lack: Added a comment to the service file to make this more obvious.

lack:
I think this is ready to go!  The commit in this sequence that builds the `mcmaker` tool into the openshift-ci container is optional, but I think it's a good idea so we don't have to worry about the CI job having to do it on the fly.

The other commit I have held (#711) is identical to this one but without that commit in the series.

ztp/Makefile
```diff
@@ -4,6 +4,8 @@ export GOFLAGS := $(if $(GOFLAGS),$(GOFLAGS),-mod=vendor)
 
 .PHONY: ci-job govet gotest
 ci-job: govet gotest
+.PHONY: ci-job govet gotest checkSourceCrs
+ci-job: govet gotest checkSourceCrs
```
imiller0, line 8: The diff makes it look like the .PHONY and ci-job lines are duplicated rather than changed. Can you verify this?

ztp/source-crs/MachineConfigAcceleratedStartup.yaml
```diff
@@ -11,7 +11,7 @@ spec:
...
```
imiller0, line 14: This file looks to contain static content that is equivalent to what mcmaker generates above. If the future goal is to have mcmaker generate this content, maybe separate this file out to a different commit/PR so that the current PR is focused only on the CI lane side?

lack: This PR has the CI lane change, and a commit with content that matches the exact `MachineConfigAcceleratedStartup.yaml`, but then it goes one step farther, with an extra commit as an example of a commit that changes the shell script.  I originally pushed that commit without rebuilding the yaml file, to demonstrate the CI lane failing (https://github.com/openshift-kni/cnf-features-deploy/pull/709#discussion_r713171397), but then fixed that and re-pushed it as a demonstration of the CI lane succeeding.

I agree this could be split out as a separate PR, and normally I'd agree; the change itself is reasonable but not strictly required - it just addresses one of your earlier comments about the shell script and adds some robustness around the reserved CPU detection logic.  But it's properly segregated in a separate commit in the history. I'm happy with it being a separate commit, still within the same PR.

I've also got a branch staged that I'll throw in next that does the same thing for `MachineConfigContainerMountNS.yaml`

imiller0: Makes complete sense. Thanks!

fedepaol:
Approving based on other reviews
/approve

PR #715, opened 2021-09-23
ztp: Reduce kubelet CPU usage

imiller0:
Increase the housekeeping and eviction monitoring intervals. The CPU use
of kubelet increases (approximately) linearly with increasing numbers of
pods driven largely by these tasks. Increasing the intervals brings CPU
usage back closer to baseline.


feature-configs/deploy/container-mount-namespace/container-private-mounts/mc-container-mount-namespace.yaml
```diff
@@ -1,7 +1,9 @@
 apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfig
 metadata:
-  name: container-mount-namespace
+  name: container-mount-namespace-and-kubelet-conf
```
lack, line 4: I'm actually surprised that the name change at this location didn't break the CI lanes.  We could just leave this file unchanged, but it's okay as long as it didn't break cnf-tests.

ztp/source-crs/extra-manifest/01-container-mount-ns-and-kubelet-conf.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfig
 metadata:
-  name: container-mount-namespace
+  name: container-mount-namespace-and-kubelet-conf
```
lack, line 4: Please remove this file and replace it with a symlink to ../MachineConfigContainerMountNS.yaml

lack (changes requested):

feature-configs/deploy/container-mount-namespace/container-private-mounts/mc-container-mount-namespace.yaml
imiller0: I thought the same thing. The CI lane looks for a machine config with the container-mount-namespace unit in it. The specific name of the MC isn't checked.

ztp/source-crs/extra-manifest/01-container-mount-ns-and-kubelet-conf.yaml
imiller0: Done

PR #722, opened 2021-09-28
Bug 2004510: ztp: Add authentication retries to ztp container

imiller0:
Add retry logic to cluster API authentication and fast-fail if
authentication fails after 25 retries.

### DIFF
--- a/ztp/resource-generator/src/common.sh
+++ b/ztp/resource-generator/src/common.sh
@@ -1,16 +1,36 @@
 #!/bin/bash
 
-export APISERVER=https://kubernetes.default.svc:443
-export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
-export CACERT=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-export NAMESPACE=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
+# Setup environment, validate access to API server
+# init <resourceName>
+init() {
+    export CHECK_RC=1
+    for idx in {25..1} ; do
+        export APISERVER=https://kubernetes.default.svc:443
+        export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+        export CACERT=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+        export NAMESPACE=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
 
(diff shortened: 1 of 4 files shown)


ztp/resource-generator/src/common.sh
```diff
@@ -1,11 +1,28 @@
...
-export CACERT=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-export NAMESPACE=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
+export CHECK_RC=1
+for idx in {25..1} ; do
+    export APISERVER=https://kubernetes.default.svc:443
+    export TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+    export CACERT=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+    export NAMESPACE=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
 
-oc login $APISERVER --token=$TOKEN --certificate-authority=$CACERT &> /dev/null
```
lack, line 8: Nice job getting the token off the cmdline!  Security approves :)

ztp/resource-generator/src/get-pre-sync-rv.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Log in and setup environment
+. $(dirname "$0")/common.sh $1
```
lack, line 4: If this is the way it's used, perhaps it's better to put the code in a function, as opposed to the work being done as a side-effect of sourcing it?

```
. $(dirname "$0")/common.sh

# Login and setup environment
waitForLogin $1
```

ztp/resource-generator/src/post-sync-entrypoint.sh
```diff
@@ -2,4 +2,4 @@
 
 . $(dirname "$0")/common.sh $1
```
lack, line 3: What is `$1`?  Can we give it a name?

ztp/resource-generator/src/get-pre-sync-rv.sh
lack: This approach would also lend itself to eventually adding a unit test around the `waitForLogin` function or whatever you want to call it.

imiller0: Agreed. Reworked to be a function. Side-effects from loading aren't nice :)

lack (approved):

PR #723, opened 2021-09-28
features: Revert "Reduce kubelet CPU usage"

lack:
This reverts commit 468efae8bbd195761f7c2e7758abedb3055287e9.

We want this specific feature to only be the container mount namespace
hiding, no additional kubelet changes (which we want in the DU profile
version of this MC object only)

### DIFF
--- a/feature-configs/deploy/container-mount-namespace/container-private-mounts/mc-container-mount-namespace.yaml
+++ b/feature-configs/deploy/container-mount-namespace/container-private-mounts/mc-container-mount-namespace.yaml
@@ -1,9 +1,7 @@
 apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfig
 metadata:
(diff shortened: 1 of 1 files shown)


imiller0 (commented):
/lgtm

yuvalk:
/approve
yeah, it's 100% clearer that way.
and seems that this specific droplet can easily live in it's own MC

lack:
> /approve
> yeah, it's 100% clearer that way.
> and seems that this specific droplet can easily live in it's own MC

Thanks!

The underlying kubelet change suffers from the same issues as the node relabeling idea, in that it needs to add cmdline args to kubelet but my wrapper script here isn't flexible enough to allow additional wrappers to ExecStart.

I think we can fix this in a more generic way, but for now this particular MC needs to be isolated.

PR #735, opened 2021-10-03
ztp: comment on disabling optional sriov pods

vitus133:
This commit adds a comment to the SR-IOV operator configuration manifest
on how to disable optional SR-IOV operator pods.
/assign @browsell 
/cc @lack @imiller0 @serngawy


vitus133:
@browsell what do you think, these optional pods should be enabled or disabled by default in our profile?
I left them enabled because it might be error-prone to bring up a workload without them. But if our purpose is to use the default profile for resource measurement without additional tuning, I guess we'll want them disabled by default.

ztp/source-crs/SriovOperatorConfig.yaml
```diff
@@ -4,8 +4,21 @@ metadata:
...
 spec:
-  enableInjector: true
-  enableOperatorWebhook: true
   configDaemonNodeSelector:
     "node-role.kubernetes.io/$mcp": ""
   disableDrain: false
+  # Injector and OperatorWebhook pods can be disabled (set to "false") below
+  #   to reduce the number of management pods. It is recommended to first verify the 
+  #   correctness of user manifests by working with webhook and injector pods, and only then
+  #   disable them.
```
lack, line 13:
```suggestion
  # Injector and OperatorWebhook pods can be disabled (set to "false") below
  # to reduce the number of management pods. It is recommended to start with the 
  # webhook and injector pods enabled, and only disable them after verifying the
  # correctness of user manifests.
```

lack (approved):
Looks good!  If you squash the commits now to fix the CI job complaint that they don't both start with `ztp:` this will probably be good to go!

PR #737, opened 2021-10-04
WIP do not merge

SchSeba:


hack/setup-test-cluster.sh
```diff
@@ -54,3 +54,126 @@ EOF
 # https://docs.openshift.com/container-platform/4.3/networking/multiple-networks/configuring-sr-iov.html#nw-multus-ipam-object_configuring-sr-iov
 oc patch networks.operator.openshift.io cluster --type='merge' \
       -p='{"spec":{"additionalNetworks":[{"name":"dummy-dhcp-network","simpleMacvlanConfig":{"ipamConfig":{"type":"dhcp"},"master":"eth0","mode":"bridge","mtu":1500},"type":"SimpleMacvlan"}]}}'
+
+#Note: build the index image for all the images we need
+dockercgf=`oc -n openshift-marketplace get sa builder -oyaml | grep imagePullSecrets -A 1 | grep -o "builder-.*"`
```
fedepaol, line 59: I wonder if we want to make it feature-selective.
At some point, we'll have the official images out and we may want to consume those instead of the baked ones

PR #747, opened 2021-10-13
infra: create a new entry in the makefile for the index build

SchSeba:
This commit make a new entry in the makefile as other projects are using
the `setup-test-cluster` entry.

I also create an environment variable for the operator versions.
Is not possible to move it forward to 4.10 as the some operators
still don't have the 4.10 bundels in place. I will do it in a follow up PR in the future


Makefile
```diff
@@ -43,7 +43,9 @@ wait-and-validate:
 	@echo "Validating"
 	SKIP_TESTS="$(SKIP_TESTS)" DONT_FOCUS=true TEST_SUITES="validationsuite" hack/run-functests.sh
 
-functests-on-ci: setup-test-cluster feature-deploy feature-wait functests
+functests-on-ci: setup-test-cluster setup-build-index-image feature-deploy feature-wait functests
+
+functests-on-internal-ci: setup-test-cluster feature-deploy feature-wait functests
```
fedepaol, line 48: mind changing the name to something that reflects the behaviour and not where this is supposed to run?
like functests-on-ci-no-index or something like that

PR #750, opened 2021-10-13
ztp: add conditional chronyd on day-0

vitus133:
Configures chronyd to run only when PTP is not synchronized.
The configuration is done on Day-0, which saves a reboot.
This solution will be changed in the future to rely on PTP synchronization status provided directly by the linuxptp-daemon

/assign @lack 
/cc @browsell @imiller0


ztp/source-crs-builder/MachineConfigChronydDay0/ptp-sync-check
```diff
@@ -0,0 +1,104 @@
...
+  CLOCK_STATE="good"
+fi
+
+# Take decisions based on the clock state and sync freshness
+if [[ "$CLOCK_STATE" = "good" ]] && [[ "$SYNC_FRESHNESS" = "fresh" ]]; then
+  log_debug "ensure $SYNC_MARKER_PATH exists"
+  touch $SYNC_MARKER_PATH
+fi
+
+if [[ "$CLOCK_STATE" = "bad" ]] || [[ "$SYNC_FRESHNESS" = "stale" ]]; then
```
lack, line 100: This last bit should not be in an `if`clause, but just the fall through, in case $CLOCK_STATE or $SYNC_FRESHNESS are somehow set to unexpected values

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
```diff
@@ -0,0 +1,157 @@
...
+# GLOBAL:
+#   none
+# OUTPUT:
+#   LAST_STATUS - Last line of the log containing offset and timestamp
+#     Something like "2021-10-18T16:47:06.022983665+00:00 stdout F ptp4l[8180.150]: [ens4f0] 
+#     master offset 3 s2 freq -16756169 path delay 1481"
+# RETURN:
+#   0 if all sanity checks passed, 1 otherwise
+get_last_ptp_status() {
+  LOG_SUBDIR=$(ls -tr /var/log/pods/ |grep openshift-ptp_linuxptp-daemon |tail -1)
```
lack, line 55: So this gives us the newest directory.

Would it be better to query `crictl pods` to look for a matching pod so we only find the one that's currently running? Or is the newest log dir good enough?

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
```diff
@@ -0,0 +1,157 @@
...
+  if ! var_defined $LOG_SUBDIR; then
+    return 1
+  fi
+  LOG_DIR="/var/log/pods/$LOG_SUBDIR/linuxptp-daemon-container"
+  if ! var_defined $LOG_DIR; then
```
lack, line 60: This condition is always true.  Either check `[[ -d $LOG_DIR ]]` or drop this check entirely.

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
```diff
@@ -0,0 +1,157 @@
...
+    return 1
+  fi
+  LAST_LOG_STR=$(ls -ltr $LOG_DIR/ |tail -1)
+  if ! var_defined $LAST_LOG_STR; then
+    return 1
+  fi
+  LAST_LOG_ARRAY=($LAST_LOG_STR)
+  LOG_NAME=${LAST_LOG_ARRAY[-1]}
```
lack, line 68: If you remove the `-l` from the ls command above, do you still need this array to find the last word?

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
```diff
@@ -0,0 +1,157 @@
...
+  if ! var_defined $LOG_NAME; then
+    return 1
+  fi
+  LAST_STATUS=$(tail -2000 $LOG_DIR/$LOG_NAME |grep "] master offset" |tail -1)
```
lack, line 72: It might be a better calling convention to echo this string instead of setting a variable.  Then elsewhere you could do:

```
local LOG_STATUS=$(get_last_ptp_status)
```

And not have to rely on a side-effect global variable set.

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
```diff
@@ -0,0 +1,157 @@
...
+  if get_last_ptp_status && get_offset_and_timestamp && ptp_synced; then
+    log_debug "ensure $SYNC_MARKER_PATH exists"
+    touch $SYNC_MARKER_PATH
+  else
+    log_debug "ensure $SYNC_MARKER_PATH does not exist"
+    rm $SYNC_MARKER_PATH &> /dev/nul
+  fi
+}
+
+if [[ "${BASH_SOURCE[0]}" = "${0}" ]]; then
```
lack, line 154: Modulino++!!

Now... Can I encourage you to add some unit tests for some of the more critical functions?

ztp/source-crs-builder/MachineConfigChronyDynamicMasterParams/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+GOPATH=${GOPATH:-${HOME}/go}
+GOBIN=${GOBIN:-${GOPATH}/bin}
+MCMAKER=${MCMAKER:-${GOBIN}/mcmaker}
+
+${MCMAKER} -stdout -name 05-chronyd-dynamic-params -mcp master \
+	dropin -source 20-environment-vars.conf -for chronyd-restart.service
```
lack, line 8: I had a question about this.  If eventually you're going to want the drop-in in the same MC object as the service, why have a drop-in at all?  Just put these environment variables right in the service.  Admins could craft custom drop-ins to change the values, but we don't need to do that since we own the original .service file.

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
```diff
@@ -0,0 +1,157 @@
...
+# ARGUMENTS:
+#   none
+# GLOBAL:
+#   LAST_STATUS - A log line containing offset and timestamp
+#     Example: "2021-10-18T16:47:06.022983665+00:00 stdout F ptp4l[8180.150]: [ens4f0] 
+#     master offset 3 s2 freq -16756169 path delay 1481"
+#   SYNC_AGE - Maximum allowed age in seconds of a log line to be considered valid
+#     for sync evaluation
+# OUTPUT:
+#   OFFSET_VALUE - in ns
```
lack, line 86: As suggested earlier, relying on globals and side-effect isn't great... Why not just pass in args as $@ and return values via echo?

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
vitus133: `crictl pods` gives me the pod, but not the directory it's logging to:
```bash
[root@cnfdt16 core]# ls /var/log/pods/ | grep linuxptp
openshift-ptp_linuxptp-daemon-lthj2_22cce38c-a99d-4818-ab63-3dbe6558e03c
```
versus
```bash
[root@cnfdt16 core]# crictl pods |grep ptp
3231b0544dec3       18 hours ago        Ready               ptp-operator-8c65cd8f6-p7wjx                                 openshift-ptp                                      0                   (default)
946589188a99b       18 hours ago        Ready               linuxptp-daemon-lthj2                                        openshift-ptp                                      0                   (default)
```

ztp/source-crs-builder/MachineConfigChronyDynamicMasterParams/build.sh
vitus133: Agree, removing the dropin

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
vitus133: Done

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
vitus133: Can I opt for doing it in a separate PR? This one is hanging too long...

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
vitus133: We can use crictl  to get the log path!
1. Get the linuxptp daemon container id:
```bash
# crictl ps --name linuxptp-daemon-container --latest --state Running --quiet
6947569dd2f1d247ca8e6018eae908bb8087fe77cea01643ae6763880fa34054
```
2. Get the container log path:
```bash
 # crictl inspect -o json 6947569dd2f1d247ca8e6018eae908bb8087fe77cea01643ae6763880fa34054 |jq '.status.logPath'
"/var/log/pods/openshift-ptp_linuxptp-daemon-lthj2_22cce38c-a99d-4818-ab63-3dbe6558e03c/linuxptp-daemon-container/16.log"
```
I was also trying to get away by using `crictl logs` command, but I had an issue with piping the output to `tail`. I noticed that the authors added these to the `crictl log` command options, but I wanted to pipe to tail twice: first to take a reasonable size buffer, then filter it with grep, and then take the latest line. With built-in --tail it doesn't work... Do you have an idea how can I work around this?

vitus133: We can also get away exclusively with crictl:
```bash
# crictl logs --since 1m --timestamps 6947569dd2f1d247ca8e6018eae908bb8087fe77cea01643ae6763880fa34054 |grep -m 1 "master offset"
2021-10-19T15:30:41.031729724Z ptp4l[5353.967]: [ens4f0] master offset        -16 s2 freq   +9455 path delay      1460

```
I'm still checking this

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
```diff
@@ -0,0 +1,162 @@
...
+#   none
+# GLOBAL:
+#   none
+# OUTPUT:
+#   Last line of the log containing offset and timestamp
+#     Something like "2021-10-18T16:47:06.022983665+00:00 stdout F ptp4l[8180.150]: [ens4f0] 
+#     master offset 3 s2 freq -16756169 path delay 1481"
+# RETURN:
+#   0 if all sanity checks passed, 1 otherwise
+get_last_ptp_status() {
```
lack, line 53: With the switch to crictl, I think this is a lot clearer and a lot less fragile!  Nice!

ztp/source-crs-builder/MachineConfigChronyDynamicMaster/ptp-sync-check
lack: > Can I opt for doing it in a separate PR? This one is hanging too long...

I'm of the opinion that unit tests should always be in the same commit as the related code, or at least in the same PR, so our CI can prove the new code is okay and does what it intends.

That said, with the changeover to crictl, a lot of the complexity has been removed, and with a lot of our scripts mocking complex cmdline utilities is more trouble that it's worth. I think if you focused your unit testing on get_ptp_offset I'd be happy enough.

PR #768, opened 2021-10-26
ztp: Add amq interconnect operator for ptp operator fast event requirement

aneeshkp:
Signed-off-by: Aneesh Puttur <aneeshputtur@gmail.com>
AMQ interconnect operator is a messaging router, required for PTP when the fast event publishing option is enabled in the ptp config.
This operator runs under its own namespace and allows event publisher's/consumer's to connect to transport hosts and produce/consume messages.

### DIFF
--- a/ztp/source-crs/AmqSubscription.yaml
+++ b/ztp/source-crs/AmqSubscription.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: amq7-interconnect-subscription
+  namespace: amq-router
+spec:
+  channel: 1.10.x
+  name:  amq7-interconnect-operator
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
--- a/ztp/source-crs/AmqSubscriptionNS.yaml
+++ b/ztp/source-crs/AmqSubscriptionNS.yaml
@@ -1,0 +1,7 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: amq-router
+  annotations:
+    workload.openshift.io/allowed: management
--- a/ztp/source-crs/AmqInstance.yaml
+++ b/ztp/source-crs/AmqInstance.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: interconnectedcloud.github.io/v1alpha1
+kind: Interconnect
+metadata:
+  name: amq-router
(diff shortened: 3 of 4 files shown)


ztp/source-crs/AmqSubscriptionNS.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: amq-router
+  annotations:
+    workload.openshift.io/allowed: management
```
imiller0, line 7: Note that the amq pod (update needed in its codebase) will need the corresponding annotation in order to be allocated to the management cores:
` target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'`

imiller0 (commented):
How much CPU load does the AMQ operator and router add to the system under normal load for this use case?

aneeshkp:
sum (rate (pod:container_cpu_usage:sum{namespace="amq-router"}[24h])) by (pod)
amq Operator 0.02m and router 0.0025m
and for openshift-ptp namespace
sum (rate (pod:container_cpu_usage:sum{namespace="openshift-ptp"}[24h])) by (pod)
Here the ptp daemon is around 0.7m
there will be temporary spikes, short burst .

ztp/source-crs/AmqSubscriptionNS.yaml
aneeshkp: Create a Jira for AMQ team to track this requirement 
https://issues.redhat.com/browse/ENTMQIC-3265

aneeshkp:
@imiller0  Can this be merged?, the performance issue is with the PTP daemon. I have a different story to investigate those.
Also AMQ team is working on adding the annotation

ztp/source-crs/AmqSubscription.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: amq7-interconnect-subscription
+  namespace: amq-router
+spec:
+  channel: 1.10.x
+  name:  amq7-interconnect-operator
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
```
imiller0, line 11: Can you add a newline to keep github from flagging this.

aneeshkp: Done

PR #769, opened 2021-10-26
ztp: add default catsrc to source crs

vitus133:
Add default catalog source for
consolidating all the disconnected catalog sources
/assign @imiller0 
/cc @serngawy


imiller0 (approved):
/lgtm

PR #770, opened 2021-10-27
ztp: remove machineconfig disabling chronyd

vitus133:
With the introduction of dynamic chronyd service deployed on day0,
there is no need in this machineconfig.
/hold until https://github.com/openshift-kni/cnf-features-deploy/pull/750 is merged
/assign @lack 
/cc @imiller0 @serngawy


ztp/source-crs/MachineConfigDisableChronyd.yaml
```diff
@@ -1,33 +0,0 @@
-apiVersion: machineconfiguration.openshift.io/v1
```
vitus133, line 1: Or should I leave this in source crs for backward compatibility?

lack (approved):

PR #778, opened 2021-11-01
oot-driver: use DockerImage insted of ImageStreamTag for the driverContainer

sudomakeinstall2:
This commit allows the flexibility to use an on-premise private docker registry as well as the internal openshift registry.

### DIFF
--- a/hack/setup-build-index-image.sh
+++ b/hack/setup-build-index-image.sh
@@ -61,8 +61,6 @@ spec:
 
           git clone --single-branch --branch OPERATOR_RELEASES https://github.com/openshift/metallb-operator.git
           cd metallb-operator
-          # TODO: temp sed till "quay.io/openshift/origin-metallb-frr" is public.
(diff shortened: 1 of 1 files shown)


SchSeba:
Hi @sudomakeinstall2 can you please introduce the change under an environment variable?

I reason I wanted to use the imageStream is that we can add a trigger to that this mean if someone update the source container (for a new driver version for example) you will get a new build started automatically

sudomakeinstall2:
@SchSeba I added environment variable USE_DOCKER_IMAGE. Setting this environment variable replaces ImageStreamTag with DockerImage. I couldn't come up with a better name for the variable, I am open to suggestions.

Also I changed the tag of the `oot-source-driver ` and removed the nested namespace. This allows us to use quay which doesn't support nested namespaces.

SchSeba:
Hi @sudomakeinstall2 please add `oot-driver: ` to the commit title to make the CI happy

PR #783, opened 2021-11-03
ztp: Create a fixed ACM inform policy

Missxiaoguo:
At the initial stage, this fixed inform policy is built to measure
some conditions that cannot be captured via the configuration policies
which are applied in the ztp pipeline. It should be used along with
configuration policies to determine if ztp of a cluster is completed
or not.

Jira: https://issues.redhat.com/browse/CNF-2959

### DIFF
--- a/ztp/policygenerator/resources/acm-policy-du-platform-check-template.yaml
+++ b/ztp/policygenerator/resources/acm-policy-du-platform-check-template.yaml
@@ -1,0 +1,185 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: du-platform-check
+  labels:
+    name: du-platform-check
+---
+# This inform policy verifies the following conditions:
+#     MCP has required MCs applied
+#     Sriov network operator is available and ready
+#     SriovNetworkNodeState is synced
+#     Ptp operator is available and ready
+#     Pao operator is available and ready
+#     kube-apiserver pod is running
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  name: du-platform-check-policy
+  namespace: du-platform-check
+  annotations:
+    policy.open-cluster-management.io/categories: CM Configuration Management
+    policy.open-cluster-management.io/controls: CM-2 Baseline Configuration
+    policy.open-cluster-management.io/standards: NIST SP 800-53
+spec:
+  remediationAction: inform
+  disabled: false
+  policy-templates:
+    - objectDefinition:
+        apiVersion: policy.open-cluster-management.io/v1
+        kind: ConfigurationPolicy
+        metadata:
+          name: du-platform-check-policy-check
+        spec:
+          remediationAction: inform
+          severity: low
+          namespaceselector:
+            exclude:
+                - kube-*
+            include:
+                - '*'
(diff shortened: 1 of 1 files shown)


lack:
Should this be in `ztp/source-crs` instead of `ztp/policygenerator/resources` so it can be added to the cluster by PGT?

Missxiaoguo:
> Should this be in `ztp/source-crs` instead of `ztp/policygenerator/resources` so it can be added to the cluster by PGT?

I was thinking the same question, but I feel it's more make sense to add in policygenerator resource for now as this policy is not the resource we intend to add in the PGT. The usage of this policy requires PolicyGen tool update to create a instance directly from this yaml.  When starting CNF-2960/CNF-2961, we can split the policy into pieces of CRs and put them in ztp/source-crs?

My two cents, I am not strongly against your thought!

ztp/policygenerator/resources/acm-policy-du-platform-check-template.yaml
```diff
@@ -0,0 +1,185 @@
...
+                        name: 02-master-workload-partitioning
+                      - apiVersion: machineconfiguration.openshift.io/v1
+                        kind: MachineConfig
+                        name: 04-accelerated-container-startup-master
+                      - apiVersion: machineconfiguration.openshift.io/v1
+                        kind: MachineConfig
+                        name: 05-chronyd-dynamic-master
+                      - apiVersion: machineconfiguration.openshift.io/v1
+                        kind: MachineConfig
+                        name: 50-performance-openshift-node-performance-profile
```
imiller0, line 82: Just a note that follow-on work will dynamically populate these values with the specific names of MC objects that are included in the deployment.

ztp/policygenerator/resources/acm-policy-du-platform-check-template.yaml
```diff
@@ -0,0 +1,185 @@
...
+                  conditions:
+                    - type: Available
+                      status: "True"
+                      reason: MinimumReplicasAvailable
+                  availableReplicas: 1
+                  readyReplicas: 1
+            - complianceType: musthave
+              objectDefinition:
+                apiVersion: apps/v1
+                kind: Deployment
```
imiller0, line 122: We should also verify that the PerformanceProfile status includes Available=true and Degraded=false. This can be done in a subsequent improvement.

imiller0 (approved):

imiller0:
> > Should this be in `ztp/source-crs` instead of `ztp/policygenerator/resources` so it can be added to the cluster by PGT?
> 
> I was thinking the same question, but I feel it's more make sense to add in policygenerator resource for now as this policy is not the resource we intend to add in the PGT. The usage of this policy requires PolicyGen tool update to create a instance directly from this yaml. When starting CNF-2960/CNF-2961, we can split the policy into pieces of CRs and put them in ztp/source-crs?
> 
> My two cents, I am not strongly against your thought!

I can see going either way. For now I think resources is likely the best location because a follow-on improvement will be to dynamically generate portions of this policy based on what the user actually includes in their PGT. It would be somewhat circular to include this within source-crs when it is partly generated based on the source-crs.

PR #784, opened 2021-11-04
infra: Ansible playbooks to deploy SRO on a HUB

daosman:
First set of Ansible playbooks to help manage out of tree drivers using the "Special Resource Operator" on HUB/Spoke OCP deployments.

US: https://issues.redhat.com/browse/CNF-3220
EPIC:  https://issues.redhat.com/browse/CNF-2375


tools/oot-driver/playbooks/resources/sro_subscription.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: special-resource-operator
+  namespace: openshift-special-resource-operator
+spec:
+  channel: alpha
+  installPlanApproval: Automatic
+  name: special-resource-operator
+  source: community-operators
```
SchSeba, line 10: we need to deploy the sro from the redhat-operators not the community one

tools/oot-driver/playbooks/roles/deploy_sro/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+    api_version: operators.coreos.com/v1
+    kind: OperatorGroup
+    name: openshift-special-resource-operators
+    namespace: "{{ sro_namespace }}"
+    kubeconfig: "{{ kubeconfig }}"
+    state: present
+
+- name: Create the SRO subscription
+  kubernetes.core.k8s:
+    definition: "{{ lookup('file', '../../../resources/sro_subscription.yml') | from_yaml }}"
```
SchSeba, line 20: can't the subscription be also in place here? there is a reason to have it different from the OperatorGroup?

tools/oot-driver/playbooks/roles/get_managed_clusters/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
...
+  set_fact:
+    ocp_versions: "{{ ocp_versions + 
+                    [{'ClusterName': item.metadata.name,
+                      'ClusterVersion': item.metadata.labels.openshiftVersion}] }}"
+  loop: "{{ mc_response.resources }}"
+  when: "item.metadata.labels.openshiftVersion is defined"
+  loop_control:
+    label: "{{ item.metadata.name }}"
+
+- name: Display Managed clusters
```
SchSeba, line 28: do you need this debug here or it was just for your testing?

tools/oot-driver/playbooks/deploy_sro.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Play to deploy SRO on a HUB cluster
```
SchSeba, line 2: you also need the nfd to be deploy.

please check the feature-configs/deploy/sro/ folder

SchSeba:
in general I think this should be in the ZTP folder not the tools folder.

tools/oot-driver/playbooks/deploy_sro.yml
daosman: 1. NFD is automatically deployed as a pre-requisite to SRO.
2. We do not actually need NFD on the HUB cluster since we don't want it to label the HUB cluster itself as we install SRO on the HUB.  For now, we let SRO deploy NFD as a pre-requisite but we do not create an instance of the CR because we don't want the HUB labeled.

tools/oot-driver/playbooks/roles/deploy_sro/tasks/main.yml
daosman: The reason why I put it as a file instead of using the kubernetes.core.k8s module is I was not sure how to pass the metadata section of the YAML file through that module.

tools/oot-driver/playbooks/roles/get_managed_clusters/tasks/main.yml
daosman: It was for my own testing.  Will remove

tools/oot-driver/playbooks/deploy_sro.yml
daosman: Seems that SRO deploys NFD automatically when I add the SRO Subscription.

pacevedom: You need to deploy the NFD special resource or else SRO wont stay running. This [file](https://github.com/openshift-kni/cnf-features-deploy/blob/master/feature-configs/deploy/sro/nfd-config.yaml)

PR #794, opened 2021-11-09
ztp:Add Ptp operator config to enable fast event

aneeshkp:
Signed-off-by: Aneesh Puttur <aneeshputtur@gmail.com>
Adding PTP Operator Config policy to enable or disable fast events.

### DIFF
--- a/ztp/source-crs/PtpOperatorConfigForEvent.yaml
+++ b/ztp/source-crs/PtpOperatorConfigForEvent.yaml
@@ -1,0 +1,10 @@
+apiVersion: ptp.openshift.io/v1
+kind: PtpOperatorConfig
+metadata:
+  name: default
+  namespace: openshift-ptp
+spec:
+  daemonNodeSelector: {}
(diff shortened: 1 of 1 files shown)


ztp/source-crs/PtpOperatorConfigForEvent.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: ptp.openshift.io/v1
+kind: PtpOperatorConfig
+metadata:
+  name: default
```
imiller0, line 4: PtpOperatorConfig is a Namespace scoped CR. Can you add
```suggestion
  name: default
  namespace: openshift-ptp
```

imiller0 (changes requested):

ztp/source-crs/PtpOperatorConfigForEvent.yaml
aneeshkp: done

PR #797, opened 2021-11-11
features: Fix gatekeeper deployment

SchSeba:
There is no need anymore to request a specific image.

Also the gatekeeper operator add a PDB and requesting 1 replica stuck the node from being drained because there is no other instance of the webhook deployment.

### DIFF
--- a/feature-configs/deploy/gatekeeper/config.yaml
+++ b/feature-configs/deploy/gatekeeper/config.yaml
@@ -3,13 +3,5 @@ kind: Gatekeeper
 metadata:
   name: gatekeeper
 spec:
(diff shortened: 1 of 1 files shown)


cynepco3hahue:
Both failures are expected
```
[Fail] [rfe_id:27363][performance] CPU Management when pod runs with the CPU load balancing runtime class [It] [test_id:32646] should disable CPU load balancing for CPU's used by the pod 
/go/src/github.com/openshift-kni/cnf-features-deploy/vendor/github.com/openshift-kni/performance-addon-operators/functests/1_performance/cpu_management.go:323
```

PR #798, opened 2021-11-11
Update PAO manifests to use 4.10 images

cynepco3hahue:


feature-configs/deploy/performance/operator_catalogsource.yaml
```diff
@@ -10,6 +10,6 @@ spec:
   icon:
     base64data: ""
     mediatype: ""
-  image: quay.io/openshift-kni/performance-addon-operator-index:4.9-snapshot
+  image: quay.io/openshift-kni/performance-addon-operator-index:4.10-snapshot
```
bartwensley, line 13: Why is a separate catalog required for the PAO? It seems the PAO is present in the redhat-operators catalog and the OCP documentation [1] points there as well.

[1] https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes#install-operator-cli_cnf-master)

cynepco3hahue: The cnf-features-deploy should point to the 4.10 PAO image that is still not released and does not exist under the redhat-operators.

bartwensley: OK - thanks for the explanation.

SchSeba (commented):
/lgtm
/approve

PR #800, opened 2021-11-12
ztp: Add platform update source CR for upgrades operator interoperability 

nishant-parekh:
ztp: Added ClusterVersion source CR and updated common PGT to include reference

### DIFF
--- a/ztp/source-crs/ClusterVersion.yaml
+++ b/ztp/source-crs/ClusterVersion.yaml
@@ -1,0 +1,12 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterVersion
+metadata:
+  name: PlatformUpgrade
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "100"
+spec:
+  channel: $channel
+  desiredUpdate:
+    force: false
+    version: $version
+    image: $image


ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
```diff
@@ -9,6 +9,15 @@ spec:
...
     # Create inform policy to validate configuration CRs that will be applied to all clusters
+    - fileName: ClusterVersion.yaml
+      policyName: "cluster-version-policy"
+      remediationAction: inform
+      spec:
+        channel: "4.10"
+        desiredUpdate:
+          force: false
+          version: "4.10"
+          upstream: "upstream"
```
imiller0, line 20: I don't think we should include the ClusterVersion CR in the original ZTP set of CRs. The cluster installation version will be controlled by the choice of ClusterImageSet in the SiteConfig. If the user wants to do an upgrade they should create a Policy with the ClusterVersion CR (can be via gitops ang PGT, just not part of the ZTP set of PGTs) and then manually create an upgrade CR to roll that out.

ztp/policygenerator-kustomize-plugin/testPolicyGenTemplate/common-ranGen.yaml
```diff
@@ -6,16 +6,28 @@ metadata:
...
     # Create operators policies that will be installed in all clusters
+    - fileName: ClusterVersion.yaml
+      policyName: "cluster-version-policy"
+      remediationAction: inform
+      spec:
+        channel: "4.10"
+        desiredUpdate:
+          force: false
+          version: "4.10"
+          upstream: "upstream"
```
imiller0, line 20: We should make test data "generic" and not copy production code in. Having the same content here and in the reference PGT leads to confusion about which set of data is the active code for users.

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
nishant-parekh: ok so want to make sure I understand it correctly:
1. should it be its own set of crs that we only need for upgrades like for instance `upgradesCRs` is that what we are looking for? 
2. should it be part of the `sourceCRs` and have its own PGT that we could integrate in the platform upgrade flow?

ztp/policygenerator-kustomize-plugin/testPolicyGenTemplate/common-ranGen.yaml
nishant-parekh: I will update this to generic values

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
imiller0: I don't have a problem with including the CR in source-crs. My request is that it not be a part of the reference PolicyGenTemplates that are used for ZTP (any of the PGTs under ztp/gitops-subscriptions/argocd/examples/policygentemplates). 
If at "day-N" a user wants to use a PGT to create this ClusterVersion CR they can do so because we've included it in source-crs, we're just not creating it as part of ZTP (because it isn't needed then),

imiller0 (commented):
/lgtm

PR #801, opened 2021-11-12
ztp: WIP separate siteconfig from PGT

serngawy:
- separate siteconfig from PGT
- add siteconfig kustomize plugin
- add resource-generator
- allow enduser to add extra-manifest to siteconfig


ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -90,31 +95,29 @@ type Clusters struct {
...
+	ClusterType            string            `yaml:"clusterType"`
 	AdditionalNTPSources   []string          `yaml:"additionalNTPSources"`
 	Nodes                  []Nodes           `yaml:"nodes"`
 	MachineNetwork         []MachineNetwork  `yaml:"machineNetwork"`
 	ServiceNetwork         []string          `yaml:"serviceNetwork"`
-	ManifestsConfig        ManifestsConfig   `yaml:"manifestsConfig"`
-	ClusterType            string            `yaml:"clusterType"`
 	NumMasters             uint8             `yaml:"numMasters"`
 	NumWorkers             uint8             `yaml:"numWorkers"`
-	ClusterProfile         string            `yaml:"clusterProfile"`
```
lack, line 101: If you're removing it here, should you remove it from the CRD as well?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -78,20 +73,23 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
 			cluster := siteConfigTemp.Spec.Clusters[clusterId]
 			dataMap, err := scbuilder.getExtraManifest(dataMap, cluster)
 			if err != nil {
-				return clusterCRs, err
+				log.Printf("warning could not read extra-manifest %s.%s %s\n", cluster.ClusterName, cluster.ExtraManifestPath, err)
 			}
 
-			// FIXME: Assuming 1 node for SNO deployment needs to be changed for RWN deployment
-			if len(siteConfigTemp.Spec.Clusters[clusterId].Nodes) > 0 {
+			// Adding workload partitions MC only for SNO clusters.
```
lack, line 79: Nice!

lack (changes requested):

ztp/gitops-subscriptions/argocd/README.md
```diff
@@ -6,101 +6,131 @@
...
 1. Create a directory structure with separate paths for SiteConfig and PolicyGenTemplate CRs
-1. Add pre-sync.yaml and post-sync.yaml from resource-hook-example/*policygentemplates*/ to the path for PolicyGenTemplate CRs
-1. Add pre-sync.yaml and post-sync.yaml from resource-hook-example/*siteconfig*/ to the path for SiteConfg CRs
-    > If your hub cluster operates in a disconnected environment you must update the `image` all four pre/post sync hook CRs.
+2. Export the argocd directroy from the ztp-site-generator container image by executing the following commands
+```
+    $ mkdir -p ./out
+    $ podman create -ti --name ztp-site-gen ztp-site-generator:latest bash
+    $ podman cp ztp-site-gen:/home/ztp ./out
+    $ podman rm -f ztp-site-gen
```
lack, line 14: Not necessary in this PR, but should we consider an easier way?  Maybe something like:
```
podman run -v out:/out ztp-site-generator:$tag extract deployment /out
```

Where `extract` is a script we provide inside the container that copies the relevant data to the given directory?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -78,20 +73,23 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
 			cluster := siteConfigTemp.Spec.Clusters[clusterId]
 			dataMap, err := scbuilder.getExtraManifest(dataMap, cluster)
 			if err != nil {
-				return clusterCRs, err
+				log.Printf("warning could not read extra-manifest %s.%s %s\n", cluster.ClusterName, cluster.ExtraManifestPath, err)
```
lack, line 76: If `getExtraManifest` returns an error, why are we continuing?  We don't have any guarantee that `dataMap` is still in a valid state for use below.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -6,16 +6,21 @@ import (
 	"strings"
 )
 
-const clusterCRsFileName = "cluster-crs.yaml"
-const extraManifestPath = "extra-manifest"
-const workloadPath = extraManifestPath + "/workload"
+//const clusterCRsFileName = "cluster-crs.yaml"
```
lack, line 9: Why not remove this?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -180,7 +178,7 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
 			continue
 		}
 
-		filePath := extraManifestPath + "/" + file.Name()
+		filePath := dir + "/" + localExtraManifestPath + "/" + file.Name()
```
lack, line 181: Using `localExtraManifestPath` here seems fragile; should `GetResourceFiles()` return that, since it knows what directory it was operating on when it created the `files` list?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -193,21 +191,41 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
...
 	}
+
+	// Adding End User Extra-manifest
+	files, err = GetFiles(clusterSpec.ExtraManifestPath)
+	if err != nil {
+		return dataMap, err
+	}
+	for _, file := range files {
+		if file.IsDir() {
+			continue
```
lack, line 210: Should we recurse into subdirectories?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -193,21 +191,41 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
...
+		}
+
+		filePath := clusterSpec.ExtraManifestPath + "/" + file.Name()
```
lack, line 213: Because we're adding content provided by the end-user, we should probably log this - Print out at least the filenames that we're adding.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -193,21 +191,41 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
...
+		manifestFile, err := ReadFile(filePath)
```
lack, line 214: For the future (not for this PR): Should we do the same `.tmpl` template processing for user input files?

ztp/gitops-subscriptions/argocd/deployment/deployment-openshift-repo-server-patch.json
```diff
@@ -0,0 +1,49 @@
...
+            ],
+            "imagePullPolicy": "Always",
+            "volumeMounts": [
+              {
+                "name": "kustomize",
+                "mountPath": "/.config"
+              }
+            ],
+            "terminationMessagePolicy": "File",
+            "image": "quay.io/redhat_emp1/ztp-site-generator:latest",
```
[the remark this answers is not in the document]
serngawy: so that mean the version should 4.10 for master branch ?

ztp/gitops-subscriptions/argocd/deployment/deployment-openshift-repo-server-patch.yaml
```diff
@@ -0,0 +1,27 @@
...
+            - sh
+            - '-c'
+          imagePullPolicy: Always
+          volumeMounts:
+            - name: kustomize
+              mountPath: /.config
+          terminationMessagePolicy: File
+          image: 'quay.io/redhat_emp1/ztp-site-generator:latest'
+          args:
+            - 'cp -r /kustomize /.config'
```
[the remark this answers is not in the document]
serngawy: kind of duplicated, will remove it then.

ztp/gitops-subscriptions/argocd/example/siteconfig/test-sno-secrets.yaml
```diff
@@ -0,0 +1,20 @@
+# test-sno-bmh-secret & assisted-deployment-pull-secret need to be created under same namespace test-sno as siteconfig
+apiVersion: v1
+data:
+  password: cHdkX2Jhc2U2NA== # pwd_based64
```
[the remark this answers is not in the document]
serngawy: not recommending that just example how the secrets name & namsespace should looks like.

ztp/siteconfig-generator/siteConfig/siteConfig.go
serngawy: sure

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: cause the extra-manifest configMap if it has missing data doesn't prevent the cluster provisioning it will just raise warning. We don't validate the extra-manifest as we open it to end user

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: `GetResourceFiles() ` return the binary execution dir and our extra-manifest dir in relative path to it. Is that what u mean ?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: I will document this it needs to be flat dir for now then create enhancement as we discuss overlay & template for extra-manifest.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: agree, lets keep for future work

ztp/gitops-subscriptions/argocd/example/siteconfig/test-sno-secrets.yaml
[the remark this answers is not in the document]
lack: I think having this in the same directory as the examples we want users to copy and paste is not a great choice.  Maybe in a README.md somewhere instead?

ztp/gitops-subscriptions/argocd/deployment/deployment-openshift-repo-server-patch.json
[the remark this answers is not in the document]
lack: @imiller0 What do you think?  Release branches should point at the real release image, but what should our master point at?  How and when do we keep our redhat_emp1 copy up-to-date?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
lack: This first level of extra-manifests is the one inside our container, which we consider critical to a successful deployment.  I agree for that the user-supplied additional directory that should be ignored.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
lack: I guess I'm confused - If GetResourceFiles returns the directory and the files, why do we need to do anything more than `dir + file.Name()` here?  Why do we need to know that they're in `localExtraManifestPath`?

serngawy: cause our extra-manifest in sub directory not in the same directory.

ztp/siteconfig-generator/siteConfig/filesHandler.go
```diff
@@ -0,0 +1,67 @@
...
+	dir := filepath.Dir(ex)
+	files, err := GetFiles(dir + "/" + localExtraManifestPath)
+	if err != nil {
+		dir, err = os.Getwd()
+		if err != nil {
+			return nil, "", err
+		}
+		files, err = GetFiles(dir + "/" + localExtraManifestPath)
+	}
+	return files, dir, err
```
lack, line 66: Here - We're returning files that are relative to `dir + "/" + localExtraManifestPath` so why not just
```
    return files, dir + "/" + localExtraManifestPath, err
```
Then the caller doesn't have to know where it found them.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
lack: See my [related comment](https://github.com/openshift-kni/cnf-features-deploy/pull/801/files#r752551696)

ztp/siteconfig-generator/siteConfig/filesHandler.go
serngawy: okay, got it will do. just get lost while cleaning the code

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: so if there is an error we should stop not continue or just log the error ? AND there is almost no chance that our pre-defined extra-manifest get errors when we add them. like we know the content and the files path (its hidden from end user) and we do test to make sure they r included correctly.

ztp/gitops-subscriptions/argocd/example/siteconfig/test-sno-secrets.yaml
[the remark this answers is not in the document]
serngawy: okay will remove the file.

ztp/siteconfig-generator-kustomize-plugin/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+```
+The extra-manifest directory contain the MCs (Machine Configs) that will be associated with clusters
+
+- Run the following command to execute kustomization.yaml
+```
+    $ make test
+```
+
+- Run the following command to dump the kusomization output to files under `out/` directory
```
imiller0, line 47: What is the input to kustomize that this is dumping the output for?

ztp/siteconfig-generator-kustomize-plugin/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
imiller0, line 1: Is this test infrastructure? If so, maybe put this kustomization.yaml under testSiteConfig and have the test run kustomize on that directory directly?

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/testUserExtraManifest/01-endUser-ex-machine-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+    storage:
+      files:
+        - contents:
+            source: data:,
+            verification: {}
+          filesystem: root
+          mode: 420
+          path: /etc/modprobe.d/endusr-mc.conf
+        - contents:
+            source: data:text/plain;charset=utf-8,sctp
```
imiller0, line 20: Change sctp to "test" or "endusr"

ztp/siteconfig-generator/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+In order to provision an OpenShift cluster using ACM, end user need to define the following CRs (Custom Resources);
+  - AgentClusterInstall
+  - ClusterDeployment
+  - NMStateConfig
+  - KlusterletAddonConfig
+  - ManagedCluster
+  - InfraEnv
+  - BareMetalHost
+  - ConfigMap for extra-manifest configurations
```
imiller0, line 11: Saying that the end user needs to create these makes it sound like that is what we are proposing they need to do. Maybe reword to say:

> The siteconfig-generator library makes cluster deployment easier by generating the following CRs for a user based on a SiteConfig CR instance

ztp/siteconfig-generator/main.go
```diff
@@ -0,0 +1,55 @@
...
+	// Collect and parse siteconfig files paths
+	siteConfigFiles := pflag.Args()
+	var outputBuffer bytes.Buffer
+	scBuilder, _ := siteConfigs.NewSiteConfigBuilder()
+
+	for _, siteConfigFile := range siteConfigFiles {
+		fileData, err := siteConfigs.ReadFile(siteConfigFile)
+		if err != nil {
+			log.Printf("Error: could not read file %s: %s\n", siteConfigFile, err)
+			continue
```
imiller0, line 26: If any of the files cause an error, should a variable be set so that the return code at the end can be non-zero to indicate that error? I like that it continues and generates all the files but returning success at the end seems wrong if a file failed to generate. Same for other errors below.

ztp/siteconfig-generator/main.go
```diff
@@ -0,0 +1,55 @@
...
+		clusters, err := scBuilder.Build(siteConfig)
+		if err != nil {
+			log.Printf("Error: could not build the entire SiteConfig defined by %s: %s", siteConfigFile, err)
+			continue
+		}
+		for _, crs := range clusters {
+			for _, crIntf := range crs {
+				cr, err := yaml.Marshal(crIntf)
+				if err != nil {
+					panic(err)
```
imiller0, line 45: Why panic here when other errors are logged and continued? Maybe clear out the outputBuffer and break out of this loop so that no CRs for this cluster are generated but other clusters can continue.

ztp/siteconfig-generator/siteConfig/filesHandler.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func GetExtraManifestResourceFiles() ([]os.FileInfo, string, error) {
+	ex, err := os.Executable()
+	if err != nil {
+		return nil, "", err
+	}
+
+	dir := filepath.Dir(ex)
+	files, err := GetFiles(dir + "/" + localExtraManifestPath)
```
imiller0, line 58: The code to create the localExtraManifestPath duplicates the same logic above. Can you factor this out into a common function? That makes it easier in the future to add more complex path searching or override via env var, etc

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
imiller0: Agreed. Really nice feature to have but would prefer separate PRs for different features. It makes tracking, backporting, and  reverting a feature (when necessary) easier.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -78,20 +73,23 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
...
-			if len(siteConfigTemp.Spec.Clusters[clusterId].Nodes) > 0 {
+			// Adding workload partitions MC only for SNO clusters.
+			if siteConfigTemp.Spec.Clusters[clusterId].ClusterType == SNO &&
+				len(siteConfigTemp.Spec.Clusters[clusterId].Nodes) > 0 {
 				cpuSet := siteConfigTemp.Spec.Clusters[clusterId].Nodes[0].Cpuset
 				if cpuSet != "" {
 					k, v, err := scbuilder.getWorkloadManifest(cpuSet)
 					if err != nil {
-						return clusterCRs, err
+						log.Printf("warning could not read WorkloadManifest %s %s\n", cluster.ClusterName, err)
```
imiller0, line 86: If this fails we will end up deploying a cluster without workload partitioning which is not good. We should consider it a failure if any of our baseline/required CRs aren't created for the cluster.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -193,21 +191,44 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
...
+		}
+	}
+
+	// Adding End User Extra-manifest
+	if clusterSpec.ExtraManifestPath != "" {
+		files, err = GetFiles(clusterSpec.ExtraManifestPath)
+		if err != nil {
+			return dataMap, err
+		}
+		for _, file := range files {
```
imiller0, line 209: The logic for handling user files is the same as handling for our standard extra manifests. This should be a common function used for both. This would also enable template handling for user manifests. If that is not desired a flag can be passed to the common function to suppress it.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -170,8 +168,8 @@ func (scbuilder *SiteConfigBuilder) getWorkloadManifest(cpuSet string) (string,
 }
 
 func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interface{}, clusterSpec Clusters) (map[string]interface{}, error) {
```
imiller0, line 170: Rather than passing dataMap in as a parameter and then returning the updated map (all resulting in copies of the map) you can pass dataMap in as a pointer and modify in place. Just be sure that you handle the error case in the caller if you are relying on returning "nil" to clear it out.

imiller0 (commented):
Can you add unit tests which cover the new "user provided" extra-manifests? Since the content is user provided having unit tests which cover failure scenarios is especially important. Invalid yaml, templating, filename collision with our standard set, etc.
Thanks!

ztp/gitops-subscriptions/argocd/example/siteconfig/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+generators:
+- test-sno.yaml
```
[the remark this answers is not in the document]
lack: We had talked about the ability to use wildcards here - We should make sure that's valid and put that in our example if it is.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
lack: Proper error-checking here also helps us make sure we haven't made a mistake, since we run this code in our CI lanes.

I'd say we should log and stop.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -193,21 +191,44 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
...
+			}
+
+			filePath := clusterSpec.ExtraManifestPath + "/" + file.Name()
+			manifestFile, err := ReadFile(filePath)
 			if err != nil {
 				return dataMap, err
 			}
+
 			manifestFileStr := string(manifestFile)
 			dataMap[file.Name()] = manifestFileStr
```
lack, line 221: This will allow a user to completely override any of the files we provide if they put one with the same filename in their override directory.

Not saying this is a bad thing, but I want to make sure we introduce this facility deliberately, not accidentally.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: okay, so it will follow the same logic and return empty configMap then it will break in the outer caller return.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: not really fan of using pointers especially as there is a recursive call function to get the field value.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: okay, I believe this will get better shape as we discuss it in future enhancement.

ztp/gitops-subscriptions/argocd/deployment/argocd-openshift-gitops-patch.json
```diff
@@ -0,0 +1,17 @@
+{
```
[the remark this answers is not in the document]
lack: Why do we need the json copy?

ztp/gitops-subscriptions/argocd/deployment/clusters-app.yaml
```diff
@@ -12,11 +12,15 @@ spec:
...
     server: https://kubernetes.default.svc
     namespace: clusters-sub
-  project: default
+  project: ztp-app-project
   source:
     path: ztp/gitops-subscriptions/argocd/resource-hook-example/siteconfig
     repoURL: https://github.com/openshift-kni/cnf-features-deploy
     targetRevision: master
+    # uncomment the below plugin if you will be adding the plugin binaries in the same repo->dir where
+    # the sitconfig.yaml exist AND uncomment the configManagementPlugins in the argocd-openshift-gitops.yaml
```
lack, line 21: Will users want to do this?  Who is this for?

ztp/gitops-subscriptions/argocd/example/siteconfig/kustomization.yaml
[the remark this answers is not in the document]
lack: Looking at kustomize in general, it doesn't look like they like the idea of wildcards at all.

See this discussion for example: https://github.com/kubernetes-sigs/kustomize/issues/217

ztp/gitops-subscriptions/hack/patch-argocd-dev.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# For dev use; if you will add the kustomize plugin binaries in the same repo->directory where the sitconfig.yaml exist
+# run this script to add configManagementPlugins to the ../argocd/deployment/argocd-openshift-gitops-patch.json patch
+# then re-patch the ArgoCD->openshift-gitops instance with the changes. Finally uncomment the ../argocd/deployment/clusters-app.yaml
+# plugin field and apply changes.
```
lack, line 6: Maybe this could patch both?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
lack: There's no reason to *not* allow templates in the user-provided dir, though, is there?

ztp/gitops-subscriptions/argocd/deployment/argocd-openshift-gitops-patch.json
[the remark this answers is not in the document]
serngawy: removed and added a patch script under gitops-subscriptions/hack to cover the dev case.

ztp/gitops-subscriptions/hack/patch-argocd-dev.sh
serngawy: I think no need to make patch for the App. Usually If some one will use this hack, he will need to have his own clusters-app.yaml (with another name) not using the one shared with others.

ztp/siteconfig-generator/main.go
imiller0: What about catching the error here, setting a flag and issuing the panic at the end of main()? In that way everything which can be generated successfully is done and then the panic informs kustomize that there was an issue? I don't think that argo should show "success" in the application if there are failures.

ztp/siteconfig-generator/siteConfig/filesHandler.go
imiller0: Yes there are 2 different returns here but this same logic is above in the ReadExtraManifestResourceFile too. Seems like both methods are solving the same problem of finding a resource file in the same location.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
imiller0: Did this question get resolved? I think that any failure to fully generate what we expect + what the user asks for should result in an error. Otherwise the user is left to debug why their cluster is only partially functional or doesn't perform correctly. Issuing the error early and making it apparent to the user will shorten their iteration/test cycle.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
imiller0: That's a "feature" that would be really hard to take away. What if we prevent the user from overwriting our extra-manifest and then relax that as the future enhancement? Or do you see the need to allow overwriting?

ztp/gitops-subscriptions/argocd/README.md
```diff
@@ -6,101 +6,131 @@
...
-1. Did the post hook job run?  
-    `oc describe job -n clusters-sub siteconfig-post`
-    1. Should indicate `succeeded: 1`
-    1. Note that if the job fails Argo will retry it. In some cases the first pass will fail and the second pass will do nothing but show pass
-
-1. Are there errors in the post hook job?  
-    `oc get pod -n clusters-sub` Note the name of the siteconfig-post-xxxxx pod  
-    `oc logs -n clusters-sub siteconfig-post-xxxxx`  
-    Do the logs indicate errors in the generation of CRs?
+1. Did the SiteConfig->ManagedCluster get generated to the hub cluster ?
```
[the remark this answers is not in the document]
lack: Even if it does, we're oddly silent in a lot of cases.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: so I will change the message to error and return , then it will fail to generate the CRs and break in the main call

ztp/gitops-subscriptions/hack/patch-argocd-dev.sh
serngawy: done

ztp/siteconfig-generator/main.go
imiller0: Can you add a note here (and below for the unmarshall) to document that kustomize/argocd will handle this class of errors. This will help as we are maintaining this code to know the error case is handled.

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/testUserExtraManifest/01-endUser-ex-machine-config.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: load-endusr-mc
```
lack, line 6: I see this file being added, and the directory referenced from the SiteConfig, but nothing checks that it was actually added to the final result.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -193,21 +194,51 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
...
+				return dataMap, err
+			}
+
+			manifestFileStr := string(manifestFile)
+			dataMap[file.Name()] = manifestFileStr
+		}
+	}
+
+	// Adding End User Extra-manifest
+	if clusterSpec.ExtraManifestPath != "" {
```
lack, line 208: This whole block needs a unit test.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	scBuilder, _ := NewSiteConfigBuilder()
+
+	// Expect to fail as the localExtraManifest path is in its default value
+	_, err = scBuilder.Build(sc)
+	assert.NotEqual(t, err, nil)
+	assert.Equal(t, strings.Contains(err.Error(), "no such file or directory"), true)
+
+	// Set the local extra manifest path to cnf-feature-deploy/ztp/source-crs/extra-manifest
+	scBuilder.setLocalExtraManifestPath("../../source-crs/extra-manifest")
```
lack, line 136: This dependency on an external past is not great; we should preferably refer only to things in a local `testdata` directory.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -0,0 +1,160 @@
...
+	// Setting the networkType to its default value
+	sc.Spec.Clusters[0].NetworkType = "OVNKubernetes"
```
lack, line 138: Why is this required?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -0,0 +1,160 @@
...
+	clustersCRs, err := scBuilder.Build(sc)
+	assert.Equal(t, err, nil)
+	assert.NotEqual(t, clustersCRs["test-site/cluster1"], nil)
+	// expect all the installation CRs generated
+	assert.Equal(t, len(clustersCRs["test-site/cluster1"]), len(scBuilder.SourceClusterCRs))
+
+	// Setting invalid user extra manifest path
+	sc.Spec.Clusters[0].ExtraManifestPath = "invalid-path/extra-manifest"
+	sc.Spec.Clusters[0].NetworkType = "OVNKubernetes"
```
lack, line 147: We already set this above - Why does it need to be re-set?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -0,0 +1,160 @@
...
+	assert.Equal(t, strings.Contains(err.Error(), "no such file or directory"), true)
+
+	// Setting valid user extra manifest path
+	sc.Spec.Clusters[0].ExtraManifestPath = "testdata/user-extra-manifest"
+	sc.Spec.Clusters[0].NetworkType = "OVNKubernetes"
+	clustersCRs, err = scBuilder.Build(sc)
+	assert.Equal(t, err, nil)
+	assert.NotEqual(t, clustersCRs["test-site/cluster1"], nil)
+	// expect all the installation CRs generated
+	assert.Equal(t, len(clustersCRs["test-site/cluster1"]), len(scBuilder.SourceClusterCRs))
```
lack, line 159: This is a good start but doesn't test the most important part of the extramanifest logic that changed:
- An additional file from the ExtraManifestPath with a unique name is added to the ConfigMap
- An additional file from the ExtraManifestPath with a name collision with a default manifest is not allowed

lack: Note: you could have a separate unit test around `getExtraManifest` specifically, it doesn't need to use `Build` as the entrypoint for all unit tests.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
serngawy: well, that is need because "../../source-crs/extra-manifest" become part of the lib validation. We raise error if the extra-`manifest/workload/` MC not exist ( This one reason why I was against that ;) ). otherwise will need to copy the `source-crs/extra-manifest"` under `testdata/`

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
serngawy: cause we override the NetworkType to include the annotation [L63](https://github.com/serngawy/cnf-features-deploy/blob/newSiteconfig/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go#L63)

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
serngawy: ops, forget to add the check for preventing override. will add

lack (changes requested):

ztp/siteconfig-generator/main.go
```diff
@@ -0,0 +1,55 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+
+	siteConfigs "github.com/openshift-kni/cnf-features-deploy/ztp/siteconfig-generator/siteConfig"
+	"github.com/spf13/pflag"
```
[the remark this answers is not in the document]
lack: Though honestly for all we're doing with it, don't bother with pflag, and just `import "flag"`, the standard go module: https://pkg.go.dev/flag

I think if you change the import to "flag" and then just globally `s/pflag/flag/` it will work as-is!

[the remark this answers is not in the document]
lack: I checked; it does :)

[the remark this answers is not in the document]
serngawy: thanks :) , done

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
lack: I didn't realize that!  That seems like a very bad thing to do, but I don't think we should change it today :)

lack (approved):

PR #802, opened 2021-11-12
Bug 2019181: ztp: Update PAO subscription catalog source

bartwensley:
Updating source CR for PAO subscription to use redhat-operators
catalogsource where the operator is now located.

Leaving the PaoSubscriptionCatalogSource.yaml in place to avoid breaking
any deployments that are currently using it.

Signed-off-by: bartwensley <bwensley@redhat.com>

### DIFF
--- a/ztp/policygenerator-kustomize-plugin/testPolicyGenTemplate/common-ranGen.yaml
+++ b/ztp/policygenerator-kustomize-plugin/testPolicyGenTemplate/common-ranGen.yaml
@@ -22,14 +22,9 @@ spec:
       policyName: "ptp-sub-policy"
     - fileName: PaoSubscription.yaml
       policyName: "pao-sub-policy"
-      spec:
-      # Changing the channel value will upgrade/downgrade the operator installed version.
-        channel: "4.8"
     - fileName: PaoSubscriptionNS.yaml
       policyName: "pao-sub-policy"
     - fileName: PaoSubscriptionOperGroup.yaml
-      policyName: "pao-sub-policy"
-    - fileName: PaoSubscriptionCatalogSource.yaml
       policyName: "pao-sub-policy"
     - fileName: ClusterLogOperGroup.yaml
       policyName: "log-sub-policy"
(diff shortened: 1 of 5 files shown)


ztp/source-crs/PaoSubscriptionCatalogSource.yaml
```diff
@@ -1,3 +1,5 @@
+# This CatalogSource is no longer required. The PAO operator is now available in the
```
yliu127, line 1: I would remove this comment - just like CatalogSource yamls for other operators, this file provides flexibility to override the catalog source when operators are from different sources.

ztp/policygenerator-kustomize-plugin/testPolicyGenTemplate/common-ranGen.yaml
```diff
@@ -22,15 +22,10 @@ spec:
       policyName: "ptp-sub-policy"
     - fileName: PaoSubscription.yaml
       policyName: "pao-sub-policy"
-      spec:
```
yliu127, line 25: Personally I think this section is better to be commented out instead of remove. As this provides an example on how we can override the default operator version.

ztp/source-crs/PaoSubscriptionCatalogSource.yaml
bartwensley: That makes sense - I will remove the comment.

ztp/policygenerator-kustomize-plugin/testPolicyGenTemplate/common-ranGen.yaml
bartwensley: This file is just for unit tests - the ztp/ran-crd/policy-gen-template-ex.yaml file still has the channel override as an example.

ztp/source-crs/PaoSubscriptionCatalogSource.yaml
bartwensley: Comment has been removed.

imiller0 (commented):
/approve
/lgtm

PR #805, opened 2021-11-16
ztp: Update PolicyGen to create static inform policy

Missxiaoguo:
Move the CRs need to be validated into a source YAML file
and update example common PGT to create a inform policy
to validate configuration CRs to help indicate when ztp
of a cluster is done

Jira: https://issues.redhat.com/browse/CNF-2960


ztp/policygenerator/policyGen/policyBuilder.go
```diff
@@ -85,6 +87,29 @@ func (pbuilder *PolicyBuilder) Build(policyGenTemp utils.PolicyGenTemplate) (map
 				policies[output] = acmPolicy
 			}
 		}
+
+		for key := range policyGenTemp.Spec.BindingRules {
+			// Bind static ACM validator inform policy to 'common' PGT
+			match, err := regexp.MatchString("[c|C]ommon|COMMON", key)
```
serngawy, line 93: are we grantee that "common" key word will not get change ?

serngawy (changes requested):
its better to put this on hold till. complainceType & remedationAction PRs get merged, you will need this work.

Missxiaoguo:
Commit is updated. As we discussed, leverage PGT to create validator inform policy.
/cc @serngawy @imiller0

ztp/policygenerator-kustomize-plugin/testPolicyGenTemplate/common-ranGen.yaml
```diff
@@ -7,6 +7,11 @@ spec:
   bindingRules:
     common: "true"
   sourceFiles:
+    # Create inform policy to validate configuration CRs that will be applied to all clusters
+    - fileName: validatorCRs/informDuValidator.yaml
+      complianceType: musthave
+      remediationAction: inform
+      policyName: "du-validator-policy"
```
imiller0, line 14: I don't think we should be copying all the published content into the test data directory. This adds confusion as to where the active code really is. If the added YAML presents a new type of input then creating generic test data here makes sense. In this case it looks like policy with remediationType of "inform" is new and it makes sense to add in tests to cover that.

Missxiaoguo: Updated.

serngawy (approved):

PR #806, opened 2021-11-17
ztp: Cleanup multinode siteconfig schema

lack:
- ztp: Compute SiteConfig NumMasters and NumWorkers from node roles
- ztp: deprecate SiteConfig clusters.clusterType

### DIFF
--- a/ztp/ran-crd/site-config-crd.yaml
+++ b/ztp/ran-crd/site-config-crd.yaml
@@ -84,16 +84,6 @@ spec:
                         enum:
                           - standard
                           - sno
-                      numMasters:
-                        description: Number of masters for the control plane.
-                        type: integer
-                        default: 1
-                        enum:
-                          - 1
-                          - 3
-                      numWorkers:
-                        description: Number of workers for the control plane.
-                        type: integer
                       clusterProfile:
                         description: |
                           clusterProfile specify the day 0 configuration that will get applied on the cluster. Must be one
--- a/ztp/ran-crd/site-config-crd.yaml
+++ b/ztp/ran-crd/site-config-crd.yaml
@@ -77,13 +77,6 @@ spec:
                       baseDomain:
                         description: BaseDomain is the base domain to which the cluster should belong.
                         type: string
-                      clusterType:
-                        description: clusterType either standard or sno
-                        type: string
-                        default: standard
-                        enum:
-                          - standard
-                          - sno
                       clusterProfile:
                         description: |
                           clusterProfile specify the day 0 configuration that will get applied on the cluster. Must be one
--- a/ztp/ran-crd/site-config-crd.yaml
+++ b/ztp/ran-crd/site-config-crd.yaml
@@ -77,17 +77,6 @@ spec:
                       baseDomain:
                         description: BaseDomain is the base domain to which the cluster should belong.
                         type: string
-                      clusterProfile:
-                        description: |
-                          clusterProfile specify the day 0 configuration that will get applied on the cluster. Must be one
-                           of the following none, du, cu, cu-du.
-                        type: string
-                        default: none
-                        enum:
-                          - none
-                          - du
-                          - cu
-                          - cu-du
                       networkType:
                         description: NetworkType is the SDN plugin that will be deployed on the Cluster.
                         type: string
--- a/ztp/siteconfig-generator/siteConfig/siteConfig.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfig.go
@@ -14,6 +14,7 @@ const workloadKubeletFile = "kubelet.conf"
 const cpuset = "$cpuset"
 const SNO = "sno"
 const Standard = "standard"
+const Master = "master"
 
 var Separator = []byte("---\n")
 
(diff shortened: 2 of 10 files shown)


ztp/ran-crd/site-config-crd.yaml
```diff
@@ -78,21 +78,16 @@ spec:
                         description: BaseDomain is the base domain to which the cluster should belong.
                         type: string
                       clusterType:
-                        description: clusterType either standard or sno
+                        deprecated: true
+                        description: clusterType either standard or sno (Ignored: Computed based on number of master nodes defined)
```
lack, line 82: Open question to @serngawy and @imiller0 : Should I keep these fields in as "deprecated", or should I just remove them from the CRD altogether?

ztp/siteconfig-generator/siteConfig/siteConfig_test.go
```diff
@@ -88,21 +75,79 @@ spec:
...
-	// Validate ClusterType
+	// Validate Disk Encryption type
 	assert.Equal(t, siteConfig.Spec.Clusters[0].DiskEncryption.Type, "nbde")
 	assert.Equal(t, siteConfig.Spec.Clusters[1].DiskEncryption.Type, "none")
 	assert.Equal(t, siteConfig.Spec.Clusters[2].DiskEncryption.Type, "none")
+}
+
+func TestRoleCounters(t *testing.T) {
+	input := `
+apiVersion: ran.openshift.io/v1
```
serngawy, line 97: There is a test case where the node role not set, it should be master by default

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -91,34 +91,49 @@ type Clusters struct {
...
+	rv.NumWorkers = 0
+	for _, node := range rv.Nodes {
+		if node.Role == "master" {
+			rv.NumMasters += 1
+		} else {
+			rv.NumWorkers += 1
+		}
+	}
+	// Autodetect ClusterType based on the node counts
+	if rv.NumMasters == 1 && rv.NumWorkers == 0 {
```
serngawy, line 132: I believe we should raise error if the num of nodes not match requirements similar to AI. 
- standard cluster num of masters must be 3 
- SNO cluster num of master must be 1

ztp/siteconfig-generator/siteConfig/siteConfig_test.go
```diff
@@ -88,21 +75,79 @@ spec:
...
+kind: SiteConfig
+spec:
+  clusters:
+  - clusterName: "ignore-user-supplied-values"
+    numMasters: 5
```
serngawy, line 102: if the user set num of masters and workers that mean he know what he is doing, we should check the nodes he set and raise error if the num of masters/workers didn't match.

serngawy (changes requested):

ztp/siteconfig-generator/siteConfig/siteConfig_test.go
lack: No, the user should never set numMasters or numWorkers.  The point of this change is that there is no reason to do so and we should remove that field from the CRD entirely.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -91,34 +91,49 @@ type Clusters struct {
...
 	}
 
 	out := defaults
 	err := unmarshal(&out)
 	*rv = Clusters(out)
+	// Tally master and worker counts based on node roles
+	rv.NumMasters = 0
+	rv.NumWorkers = 0
+	for _, node := range rv.Nodes {
+		if node.Role == "master" {
```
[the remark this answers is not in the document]
lack: This assumes the nodes have already been initialized properly, but I take your point.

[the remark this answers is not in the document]
lack: Hmmm... In order to be future-proof (where there may be more non-master roles like "worker-du" or variants like that), I think I have a solution that works - If Role is "master" or "", assume master.  Any other non-zero string is a worker of some kind.

ztp/ran-crd/site-config-crd.yaml
lack: Removed now.

ztp/ran-crd/site-config-crd.yaml
```diff
@@ -77,23 +77,6 @@ spec:
...
-                        description: Number of masters for the control plane.
-                        type: integer
-                        default: 1
-                        enum:
-                          - 1
-                          - 3
-                      numWorkers:
-                        description: Number of workers for the control plane.
-                        type: integer
                       clusterProfile:
```
serngawy, line 80: would u clean the cluster Profile as well, no use for it.

serngawy (commented):
sounds good, if u can remove the clusterProfile as well. if not we can do it in another PR

ztp/ran-crd/site-config-crd.yaml
lack: Sure!

serngawy (approved):

PR #811, opened 2021-11-23
ztp: Add example configs for standard and 3-node clusters

lack:


ztp/gitops-subscriptions/argocd/resource-hook-example/siteconfig/kustomization.yaml
```diff
@@ -1,4 +1,6 @@
 resources:
 - post-sync.yaml
 - pre-sync.yaml
-- test-sno.yaml  
+- test-sno.yaml
+- test-3node.yaml
+- test-standard.yaml
```
imiller0, line 6: In the current hook methodology there's no need for the kustomization file. No harm in adding these here but for now its a noop.

ztp/gitops-subscriptions/argocd/resource-hook-example/siteconfig/test-3node.yaml
```diff
@@ -0,0 +1,155 @@
...
+    numMasters: 3
+    clusterLabels:
+      group-du-sno: ""
+      common: true
+      sites : "test-3node"
+    clusterNetwork:
+      - cidr: 1001:db9::/48
+        hostPrefix: 64
+    machineNetwork:
+      - cidr: 2620:52:0:10e7::/64
```
imiller0, line 33: Maybe use clearly fictitious prefixes here to indicate we expect them to be changed for real clusters? 1111:2222:3333:4444::/64? Same below for bmc and node IPs.

ztp/gitops-subscriptions/argocd/resource-hook-example/siteconfig/test-3node.yaml
```diff
@@ -0,0 +1,155 @@
...
+                server:
+                - 2620:52:0:1310::1
+            routes:
+              config:
+              - destination: ::/0
+                next-hop-interface: eno1
+                next-hop-address: 2620:52:0:10e7::fc
+                table-id: 254
+      - hostName: "test-node2.example.com"
+        root: "master"
```
imiller0, line 79: Is this a field we can control? Prefer other terms rather than "master".

ztp/gitops-subscriptions/argocd/resource-hook-example/siteconfig/test-3node.yaml
```diff
@@ -0,0 +1,155 @@
...
+    name: "assisted-deployment-pull-secret"
+  clusterImageSetNameRef: "openshift-4.10"
+  sshPublicKey: "ssh-rsa AAAA..."
+  clusters:
+  - clusterName: "test-3node"
+    clusterType: "standard"
+    clusterProfile: "du"
+    numMasters: 3
+    clusterLabels:
+      group-du-sno: ""
```
imiller0, line 26: This points out that our original naming of the group wasn't quite right. We use the group to combine features that we expect are more regionally specific (eg remote logging server). Using "du-sno" as part of the group name implies it is specific to that type of cluster, so when we re-use it here it looks "wrong". We should consider revisiting the group naming in subsequent work.

ztp/gitops-subscriptions/argocd/resource-hook-example/siteconfig/test-3node.yaml
lack: Good catch - This was actually supposed to be `role: "master"` but because the default role is actually master, this typo was completely ignored by the tooling and the output was still correct.  Fixing.

ztp/gitops-subscriptions/argocd/resource-hook-example/siteconfig/test-3node.yaml
lack: Fair point.  Maybe for now I'll fix this just as "group-du" so it implies how you ought to have a different group policy for non-SNO clusters.

lack:
Latest update also makes the SNO example much more generic.

PR #812, opened 2021-11-23
ztp: Remove extra CR separator from StorageLV

imiller0:
The extra separator "---" was causing PolicyGen to create an empty CR in
the corresponding policy. This empty CR kept the policy from going
compliant (ACM complains it is missing a "kind" attribute).


lack (approved):
Controversial! ;)

PR #815, opened 2021-11-24
Bug 2009233: ztp: Support complianceType for full PGT

imiller0:
Give the user control over the complianceType for all CRs in a
PolicyGenTemplate and change the default to "musthave". This ensures
that, by default, we realize the CPU savings on managed clusters,
but provide a path for users to choose "mustonlyhave" policies
when needed. The difference in CPU use for managing "mustonlyhave"
policies for the current feature set is ~0.8 cpu as measured by
the prometheus query:
```
sum(pod:container_cpu_usage:sum{namespace=~"openshift-logging|openshift-authentication-operator|openshift-kube-controller-manager|openshift-kube-apiserver|openshift-apiserver|open-cluster-management-agent|open-cluster-management-agent-addon"})
```

### DIFF
--- a/ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/common-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/common-ranGen.yaml
@@ -19,14 +19,12 @@ spec:
       policyName: "sriov-sub-policy"
     - fileName: SriovSubscriptionNS.yaml
       policyName: "sriov-sub-policy"
-      complianceType: musthave
     - fileName: SriovSubscriptionOperGroup.yaml
       policyName: "sriov-sub-policy"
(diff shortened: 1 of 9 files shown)


ztp/gitops-subscriptions/argocd/resource-hook-example/policygentemplates/common-ranGen.yaml
```diff
@@ -13,20 +13,19 @@ metadata:
 spec:
   bindingRules:
     common: "true"
+  complianceType: musthave
```
Missxiaoguo, line 16: I am thinking probably more clear if we add this under something like "policyDefaults"? so a default remediationType can be added here as well.
spec:
  policyDefaults:
    complianceType: musthave

serngawy (changes requested):

serngawy (approved):

PR #817, opened 2021-11-25
infra: decrease feature-wait timeout

yuvalk:
in the past each cycle was really fast, so we needed 200 attempts just
to be on the safe side.
now that we're using the validation suite, each cycle takes 1-10 minutes
so 20 attempts are more than enough

### DIFF
--- a/hack/feature-wait.sh
+++ b/hack/feature-wait.sh
@@ -13,7 +13,7 @@ if [ "$FEATURES" == "" ]; then
 fi
 
 ATTEMPTS=0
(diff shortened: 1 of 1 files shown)


yuvalk:
/retest
this looks like a flake, but asked on slack so we can understand better

SchSeba:
closing this PR we have already made changes to the timeout https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/feature-wait.sh#L19

/close

PR #818, opened 2021-11-26
Add ZTP MachineConfig which sets iommu kernelArgs

mcornea:
To avoid issues caused by SR-IOV operator rebooting SNO cluster to apply iommu kernelArgs create a MachineConfig that sets them up at deployment time.

More details of the issue can be found in:
https://bugzilla.redhat.com/show_bug.cgi?id=2021151


browsell:
This needs to be addressed in the sriov-operator. It needs to add the necessary machine-config

PR #819, opened 2021-11-28
cnf-tests: Improve ovs_qos debuggability

oribon:
Printing some values as part of the tests will help
us debug failures, e.g knowing what was the rate limit
before the MachineConfig was deleted.
In addition, changing the names of the pod will make it
clear who is the sender/receiver.

Signed-off-by: Ori Braunshtein <obraunsh@redhat.com>

### DIFF
--- a/cnf-tests/testsuites/e2esuite/ovs_qos/ovs_qos.go
+++ b/cnf-tests/testsuites/e2esuite/ovs_qos/ovs_qos.go
@@ -103,6 +103,7 @@ var _ = Describe("ovs_qos", func() {
 			gSkipReason = "At least two nodes are required for ovs_qos tests."
 			return
 		}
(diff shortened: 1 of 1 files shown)


cnf-tests/testsuites/e2esuite/ovs_qos/ovs_qos.go
```diff
@@ -103,6 +103,7 @@ var _ = Describe("ovs_qos", func() {
 			gSkipReason = "At least two nodes are required for ovs_qos tests."
 			return
 		}
+		fmt.Printf("iperf3BitrateOverride: %s\n", iperf3BitrateOverride)
```
fedepaol, line 106: I think there was something like ginkgo.Writer to print things

oribon: changed thanks

PR #829, opened 2021-12-02
ztp: Added support for defining OCP release per cluster

mvazquezc:
This PR adds support for configuring specific OCP versions for specific clusters within the site. If cluster does not define specific OCP release, the OCP release specified at site level will be used. This is useful when you want to deploy clusters with different OCP releases in the same site, i.e: OCP 4.8.X and OCP 4.9.X. As of today, without this feature you're required to have two different SiteConfigs even if they're the same "physical" site.

Signed-off-by: Mario Vazquez <mavazque@redhat.com>

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
@@ -61,6 +61,15 @@ func (scbuilder *SiteConfigBuilder) Build(siteConfigTemp SiteConfig) (map[string
 			return clustersCRs, errors.New("Error: Repeated Cluster Name " + siteConfigTemp.Metadata.Name + "/" + cluster.ClusterName)
 		}
 		siteConfigTemp.Spec.Clusters[id].NetworkType = "{\"networking\":{\"networkType\":\"" + cluster.NetworkType + "\"}}"
+
+		if siteConfigTemp.Spec.ClusterImageSetNameRef == "" && siteConfigTemp.Spec.Clusters[id].ClusterImageSetNameRef == "" {
+			return clustersCRs, errors.New("Error: Site and cluster clusterImageSetNameRef cannot be empty " + siteConfigTemp.Metadata.Name + "/" + cluster.ClusterName)
+		}
+		// If cluster has not set a clusterImageSetNameRef we use the site one
+		if siteConfigTemp.Spec.Clusters[id].ClusterImageSetNameRef == "" {
+			siteConfigTemp.Spec.Clusters[id].ClusterImageSetNameRef = siteConfigTemp.Spec.ClusterImageSetNameRef
+		}
+
 		clusterCRs, err := scbuilder.getClusterCRs(id, siteConfigTemp)
 		if err != nil {
 			return clustersCRs, err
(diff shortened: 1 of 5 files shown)


serngawy:
thnx Mario, would u re-base this. Let me know if the re-structure is not clear

mvazquezc:
@serngawy I just sent the patches. I couldn't test locally since now there are two different binaries and I haven't had the time to test it on a live cluster. I'd say not merge until I can verify the patch continues to work as expected. Thanks.

ztp/siteconfig-generator/siteConfig/siteConfig_test.go
```diff
@@ -74,6 +74,32 @@ spec:
 	assert.Equal(t, siteConfig.Spec.Clusters[0].Nodes[3].Role, "master")
 }
 
+// Test cases for default/specific values on SiteConfig.Clusters[].ClusterImageSetNameRef
+
+func TestClusterImageSetNameRef(t *testing.T) {
```
serngawy, line 79: Would you add unit test in siteConfigBuilder_test.go making sure the AgentClusterInstall for both clusters getting different versions.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -77,6 +77,11 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
 	for _, cr := range scbuilder.SourceClusterCRs {
 		mapSourceCR := cr.(map[string]interface{})
 
+		// If cluster has not set a clusterImageSetNameRef we use the site one
+		if siteConfigTemp.Spec.Clusters[clusterId].ClusterImageSetNameRef == "" {
```
serngawy, line 81: Would u move this to the Build func (similar as networkType) AND add validation if both spec->clusterImageSetNameRef AND spec->clusters[i]->clusterImageSetNameRef are empty

ztp/policygenerator/resources/cluster-crs.yaml
```diff
@@ -16,7 +16,7 @@ spec:
   clusterDeploymentRef:
     name: siteconfig.Spec.Clusters.ClusterName
   imageSetRef:
-    name: siteconfig.Spec.ClusterImageSetNameRef
```
serngawy, line 19: This file is absolute the right place to change that is her https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/siteconfig-generator/siteConfig/clusterCRs.go#L25

ztp/siteconfig-generator/siteConfig/siteConfig_test.go
mvazquezc: Done.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
mvazquezc: Done.

ztp/policygenerator/resources/cluster-crs.yaml
mvazquezc: Done.

serngawy (approved):

PR #832, opened 2021-12-03
ztp: Update PerfProfile api to v2 and add globallyDisableIrqLoadBalancing parameter to the template

mvazquezc:
This PR adds two things to the PerformanceProfile template.

1. Updates the object apiVersion to v2
2. Adds the globallyDisableIrqLoadBalancing so user can set it to true/false. Default will be true.

Signed-off-by: Mario Vazquez <mavazque@redhat.com>


imiller0:
Just for reference I added a fix under PR 833 which allows the globalDisable to be set in the user's PGT without having to explicitly put it in the source-cr (or any other valid fields) and separated out the api version change under PR 835. Thank you for finding these!

mvazquezc:
Thanks Ian, closing this one.

PR #834, opened 2021-12-03
ztp/multinode extramanifests

lack:
- ztp: Generate all static MachineConfig objects for both workers and masters
- ztp: Rename source-crs-builder to extra-manifests-builder


lack:
General overview of the changes:
- Renamed source-crs-builder to extra-manifests-builder
- Changed the ztp/extra-manifests-builder/Makefile to make master and worker profiles
- Made source-crs/extra-manifests/* the default target where extra-manifests-builder builds its things
- Maybe symlinks from source-crs/XX to the real files in source-crs/extra-manifests/XX for those master MachineConfigs that we had previosuly put in source-crs

ztp/extra-manifests-builder/03-sctp-machine-config/build.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# The SCTP machine config just asks for the sctp kernel module to be loaded
+
+echo "\
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: ${MCPROLE}
+  name: load-sctp-module-${MCPROLE}
```
imiller0, line 13: This name change means that any users who have an existing PolicyGenTemplate with the MachineConfig included will experience an update to the MCP. One option would be to name it "load-sctp-module" if mcp==master and load-sctp-module-<mcp> for everything else.

ztp/extra-manifests-builder/README.md
```diff
@@ -8,13 +8,13 @@ subdirectory follows the following conventions:
...
-  input files remain the same (it must not include any date stamps or
-  git commit hashes).  This is used both to generate the rendered
-  `../source-crs/*.yaml` and to do an integrity check as part of the ci-job
-  target which ensures the rendered yaml stays in-sync with the source
-  content.
+- The directory must contain a `build.sh` which is run with $MCPROLE set to the
+  expected MCP role to produce.  This script produces to stdout the rendered
+  yaml identically every time it is called, provided that the input files
+  remain the same (it must not include any date stamps or git commit hashes).
+  This is used both to generate the rendered `../source-crs/*.yaml` and to do
```
imiller0, line 15: Is the output path now ../source-crs/extra-manifests?

lack: Good catch!

serngawy (commented):
By adding all manifest under ztp/source-crs/extra-manifest that means even for SNO cluster  the extra-manifest ConfigMap will have the MC for both master and worker. Did u test deploying SNO with new  extra-manifest ConfigMap ?

ztp/extra-manifests-builder/03-sctp-machine-config/build.sh
lack: I wonder if the contents is identical, if the resulting rendered config will be the same and no extra reboot would be needed...  Going to do some testing.

lack: Conclusion: If a 2nd MachineConfig object comes in with identical content and a different name, no extra reboot is incurred.  Both objects show up in the list, and the later one will be ignored.

imiller0 (approved):

lack:
> By adding all manifest under ztp/source-crs/extra-manifest that means even for SNO cluster the extra-manifest ConfigMap will have the MC for both master and worker. Did u test deploying SNO with new extra-manifest ConfigMap ?

I did.  The worker MCs don't affect the master (or SNO) in any way.

PR #836, opened 2021-12-03
ztp: CNF-3661 Reduce number of policies

pixelsoccupied:
The current reference PolicyGenTemplates (common, group, and test-sno) result in ~20 policies being created for the cluster. Managing this number of policies per cluster results in increased CPU use on the spoke cluster and scaling issues in the hub.

The work of this story is to reduce the number of policies created by using a common policy name for all policies in each of the common-ranGen.yaml, group-du-sno-ranGen.yaml, and test-sno.yaml files. 

Naming convention --> 
```
subscriptions-policy -- Containing all the Namespace, OperatorGroup and Subscription CRs
config-policy -- Containing the rest (OperatorHub, ImageContentSourcePolicy, ReduceMonitoring, etc)
```

-----------------
Before -->
```
[root@jumphost1 ~]# kubectl get policy -n common
NAME                                REMEDIATION ACTION   COMPLIANCE STATE   AGE
common-log-sub-policy       enforce              NonCompliant       2d22h
common-mon-offload-policy   enforce              Compliant          2d22h
common-pao-sub-policy       enforce              NonCompliant       2d22h
common-ptp-sub-policy       enforce              Compliant          2d22h
common-sriov-sub-policy     enforce              Compliant          2d22h
```

After --> 
```
[root@jumphost1 ~]# kubectl get policy -n common
NAME                                REMEDIATION ACTION   COMPLIANCE STATE   AGE
common-config-policy       enforce              NonCompliant       2d22h
```

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
@@ -15,47 +15,47 @@ spec:
       policyName: "du-validator-policy"
     # Create operators policies that will be installed in all clusters
     - fileName: SriovSubscription.yaml
-      policyName: "sriov-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: SriovSubscriptionNS.yaml
-      policyName: "sriov-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: SriovSubscriptionOperGroup.yaml
-      policyName: "sriov-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: PtpSubscription.yaml
-      policyName: "ptp-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: PtpSubscriptionNS.yaml
-      policyName: "ptp-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: PtpSubscriptionOperGroup.yaml
-      policyName: "ptp-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: PaoSubscription.yaml
-      policyName: "pao-sub-policy"
+      policyName: "subscriptions-policy"
       spec:
       # Changing the channel value will upgrade/downgrade the operator installed version.
         channel: "4.9"
     - fileName: PaoSubscriptionNS.yaml
-      policyName: "pao-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: PaoSubscriptionOperGroup.yaml
-      policyName: "pao-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: ClusterLogNS.yaml
-      policyName: "log-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: ClusterLogOperGroup.yaml
-      policyName: "log-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: ClusterLogSubscription.yaml
-      policyName: "log-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: StorageNS.yaml
-      policyName: "storage-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: StorageOperGroup.yaml
-      policyName: "storage-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: StorageSubscription.yaml
-      policyName: "storage-sub-policy"
+      policyName: "subscriptions-policy"
     - fileName: ReduceMonitoringFootprint.yaml
-      policyName: "mon-offload-policy"
+      policyName: "config-policy"
     #
     # These CRs are in support of installation from a disconnected registry
     #
     - fileName: OperatorHub.yaml
-      policyName: "registry-policy"
+      policyName: "config-policy"
     - fileName: DefaultCatsrc.yaml
-      policyName: "registry-policy"
+      policyName: "config-policy"
       # The Subscriptions all point to redhat-operators. The OperatorHub CR
       # disables the defaults and this CR replaces redhat-operators with a
       # CatalogSource pointing to the disconnected registry. Including both of
@@ -66,7 +66,7 @@ spec:
         displayName: disconnected-redhat-operators
         image: registry.example.com:5000/disconnected-redhat-operators/disconnected-redhat-operator-index:v4.9
     - fileName: DisconnectedICSP.yaml
-      policyName: "registry-policy"
+      policyName: "config-policy"
       spec:
         repositoryDigestMirrors:
         - mirrors:
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
@@ -11,10 +11,10 @@ spec:
   mcp: "master"
   sourceFiles:
     - fileName: ConsoleOperatorDisable.yaml
-      policyName: "console-policy"
+      policyName: "config-policy"
     # Set ClusterLogForwarder & ClusterLogging as example might be better to create another policyTemp-Group
     - fileName: ClusterLogForwarder.yaml
-      policyName: "log-forwarder-policy"
+      policyName: "config-policy"
       spec:
         outputs:
           - type: "kafka"
@@ -33,7 +33,7 @@ spec:
             outputRefs:
              - kafka-open
     - fileName: ClusterLogging.yaml
-      policyName: "log-policy"
+      policyName: "config-policy"
       spec:
         curation:
           curator:
@@ -43,7 +43,7 @@ spec:
             type: "fluentd"
             fluentd: {}
     - fileName: PtpConfigSlave.yaml   # Change to PtpConfigSlaveCvl.yaml for ColumbiaVille NIC
-      policyName: "ptp-config-policy"
+      policyName: "config-policy"
       metadata:
         name: "du-ptp-slave"
       spec:
@@ -54,11 +54,11 @@ spec:
           ptp4lOpts: "-2 -s --summary_interval -4"
           phc2sysOpts: "-a -r -n 24"
     - fileName: SriovOperatorConfig.yaml
-      policyName: "sriov-conf-policy"
+      policyName: "config-policy"
       spec:
         disableDrain: true
     - fileName: StorageLV.yaml
-      policyName: "local-disks-policy"
+      policyName: "config-policy"
       spec:
         storageClassDevices:
           - storageClassName: "example-storage-class-1"
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/example-sno-site.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/example-sno-site.yaml
@@ -10,14 +10,14 @@ spec:
   mcp: "master"
   sourceFiles:
     - fileName: SriovNetwork.yaml
-      policyName: "sriov-nw-fh-policy"
+      policyName: "config-policy"
       metadata:
         name: "sriov-nw-du-fh"
       spec:
         resourceName: du_fh
         vlan: 140
     - fileName: SriovNetworkNodePolicy.yaml
-      policyName: "sriov-nnp-fh-policy"
+      policyName: "config-policy"
       metadata:
         name: "sriov-nnp-du-fh"
       spec:
@@ -29,14 +29,14 @@ spec:
         priority: 10
         resourceName: du_fh
     - fileName: SriovNetwork.yaml
-      policyName: "sriov-nw-mh-policy"
+      policyName: "config-policy"
       metadata:
         name: "sriov-nw-du-mh"
       spec:
         resourceName: du_mh
         vlan: 150
     - fileName: SriovNetworkNodePolicy.yaml
-      policyName: "sriov-nnp-mh-policy"
+      policyName: "config-policy"
       metadata:
         name: "sriov-nnp-du-mh"
       spec:
@@ -48,7 +48,7 @@ spec:
         priority: 10
         resourceName: du_mh
     - fileName: PerformanceProfile.yaml
-      policyName: "perfprofile-policy"
+      policyName: "config-policy"
       metadata:
         name: openshift-node-performance-profile
       spec:
@@ -61,7 +61,7 @@ spec:
             - size: 1G
               count: 32
     - fileName: TunedPerformancePatch.yaml
-      policyName: "tuned-perf-patch-policy"
+      policyName: "config-policy"
       spec:
         profile:
           - name: performance-patch


ztp/gitops-subscriptions/argocd/example/policygentemplates/test-sno.yaml
```diff
@@ -16,14 +16,14 @@ spec:
   mcp: "master"
   sourceFiles:
     - fileName: SriovNetwork.yaml
-      policyName: "sriov-nw-fh-policy"
+      policyName: "config-policy"
```
imiller0, line 19: Each file (common, group, test-sno) will need a uniqe policy name, otherwise they will overwrite one another.

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
```diff
@@ -18,50 +18,50 @@ spec:
     - fileName: validatorCRs/informDuValidator.yaml
       complianceType: musthave
       remediationAction: inform
-      policyName: "du-validator-policy"
+      policyName: "config-policy"
```
imiller0, line 21: @Missxiaoguo is there any reason this policy needs to be separate from the config policies or can it be combined with them?

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
```diff
@@ -72,7 +72,7 @@ spec:
         displayName: disconnected-redhat-operators
         image: registry.example.com:5000/disconnected-redhat-operators/disconnected-redhat-operator-index:v4.9
     - fileName: DisconnectedICSP.yaml
-      policyName: "registry-policy"
+      policyName: "config-policy"
```
imiller0, line 75: If the user chooses to adapt an existing deployment to this new convention I believe it will result in duplicate/parallel policies. We don't have a good way to automatically "move" already existing policies to the new names. At a minimum we need to document for the user how to clean up the old policies, but we should think about ways to make this easier if we can.

ztp/gitops-subscriptions/argocd/example/policygentemplates/test-sno.yaml
pixelsoccupied: ah wouldn't the policy names gets prepended by `common` or `test-sno`? 

Do have any suggestion for the name? `config-policy-common`, `config-policy-group` and `config-policy-sno` ?

imiller0: Yes, apologies. The generated policy name will be `<pgtName>-<policyName>`. So `common-config-policy`, `group-du-sno-config-policy` and `test-sno-config-policy`.

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
pixelsoccupied: So basically we need to help user migrate and since this feature is actually benefiting everyone without a loss in functionality it could be made default? 

To help migrate: we can introduce a special variable `migrateToMakeItBetter` (maybe added using patch), which basically deletes the old ones and recreates policies with the same name? 

To help make it default: on the docs say `policyName` is "something you shouldn't be explicit about" and then we programatically add `config-policy` to it when it's missing/empty. 

Migration might be overkill (used by small number of users and for a limited amount of time?) but default behaviour should be implanted I think. Seeing `policyName: "config-policy"` in so many places is bad UX IMO

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
Missxiaoguo: We shouldn't combine this policy with other config policies as this is an inform policy but others are enforce and policyGen doesn't support have different types of remedicationAction in one policy.

Even we will eventually have the default to inform, we should leave it separated for LO to know this policy shouldn't be copied for enforce.

browsell (commented):
I think we need to reconsider these groupings. 

We are grouping s/w installation and operator config together. For upgrades for instance we want a single policy that encapsulates all the olm subscriptions but I don't think that should include other config, so many something like: 

common-subscriptions
common-config

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
browsell: +1

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -49,12 +49,12 @@ spec:
             type: "fluentd"
             fluentd: {}
     - fileName: MachineConfigSctp.yaml
-      policyName: "mc-sctp-policy"
+      policyName: "config-policy"
```
browsell, line 52: Why do we have a policy for this. It should be applied as part of the day 1 extra manifests, delete.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -17,10 +17,10 @@ spec:
   mcp: "master"
   sourceFiles:
     - fileName: ConsoleOperatorDisable.yaml
-      policyName: "console-policy"
+      policyName: "config-policy"
```
browsell, line 20: Why is this a group vs common policy ?

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -64,14 +64,14 @@ spec:
           ptp4lOpts: "-2 -s --summary_interval -4"
           phc2sysOpts: "-a -r -n 24"
     - fileName: SriovOperatorConfig.yaml
-      policyName: "sriov-conf-policy"
+      policyName: "config-policy"
```
browsell, line 67: Why is this group ? Mandatory for all DU deployments.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -64,14 +64,14 @@ spec:
...
       spec:
         disableDrain: true
     - fileName: MachineConfigAcceleratedStartup.yaml
-      policyName: "mc-accelerated-startup-policy"
```
browsell, line 71: Why is this not an extra manifest vs a day2 policy ?

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
browsell: When the new image is instantiated, couldn't we just have a one shot cleanup that deletes the redundant policies ?

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
imiller0: Agreed. Removing them from the templates is being done under PR #834.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
imiller0: We talked about making it common but the point was raised that in a deployment with a mix of SNO and standard clusters they may want the console for standard(?). In that case a separate group could/would be made for group-du-std which doesn't contain the Console disable.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
imiller0: I believe it is in the group because it contains the mapping to a machine config pool and the disableDrain flag. The thinking was that theyse might be different between SNO and standard clusters.

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
imiller0: Yes. Automating the move/cleanup/rename/etc is something we could (and should) do. I suspect that we will need (or want) traceability between source PolicyGenTemplate and created artifacts to make this easier.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
browsell: Fair point

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
browsell: Right, although we do not expose the selector in our current reference example,

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -58,14 +58,14 @@ spec:
           ptp4lOpts: "-2 -s --summary_interval -4"
           phc2sysOpts: "-a -r -n 24"
```
lack, line 59: I don't know if this change is in scope here, but the fact that this is a "group" template but the PTP config just above has a specific interface name seems dangerous to me.

imiller0 (commented):
Overall I think this looks good. When policies are built the name of the policy is the combined PolicyGenTemplate name and policyName field. In this PR each PGT uses "config-policy" for the policyName of all included CRs, so this will result in 3 policies created:
common-config-policy (one shared by all nodes)
group-du-sno-config-policy (one shared by all nodes in this group)
example-sno-config-policy (one per cluster, example-sno replaced by site name)

The common policy will contain all the CRs necessary to install operators along with configuration unrelated to the operators (disconnected registry access and Monitoring footprint). 

I'll wait for others to comment but otherwise will approve tomorrow.

serngawy:
Policies are same, I'm good with that. Would you just confirm that it has been tested. Like SNO is provisioned and then those  3  policies get applied and everything went fine.

pixelsoccupied:
@serngawy Here's the output with and without the consolidation from the hub. 

```
[root@jumphost1 ~]# kubectl get policy -A
I1216 09:24:38.808993 2923311 request.go:668] Waited for 1.182937751s due to client-side throttling, not priority and fairness, request: GET:https://api.mycluster.hub.ran.dfwt5g.lab:6443/apis/oauth.openshift.io/v1?timeout=32s
NAMESPACE          NAME                                                        REMEDIATION ACTION   COMPLIANCE STATE   AGE
cnfdf01-policies   cnfdf01-perfprofile-policy                                  enforce                                 6d18h
cnfdf01-policies   cnfdf01-sriov-nnp-fh-policy                                 enforce                                 6d18h
cnfdf01-policies   cnfdf01-sriov-nnp-mh-policy                                 enforce                                 6d18h
cnfdf01-policies   cnfdf01-sriov-nw-fh-policy                                  enforce                                 6d18h
cnfdf01-policies   cnfdf01-tuned-perf-patch-policy                             enforce                                 6d18h
cnfdf01            cnfdf01-policies.cnfdf01-perfprofile-policy                 enforce                                 3s
cnfdf01            cnfdf01-policies.cnfdf01-sriov-nnp-fh-policy                enforce                                 3s
cnfdf01            cnfdf01-policies.cnfdf01-sriov-nnp-mh-policy                enforce                                 3s
cnfdf01            cnfdf01-policies.cnfdf01-sriov-nw-fh-policy                 enforce                                 3s
cnfdf01            cnfdf01-policies.cnfdf01-tuned-perf-patch-policy            enforce                                 3s
cnfdf01            common-cnfdf01.common-cnfdf01-log-sub-policy                enforce                                 3s
cnfdf01            common-cnfdf01.common-cnfdf01-mon-offload-policy            enforce                                 3s
cnfdf01            common-cnfdf01.common-cnfdf01-pao-sub-policy                enforce                                 3s
cnfdf01            common-cnfdf01.common-cnfdf01-ptp-sub-policy                enforce                                 3s
cnfdf01            common-cnfdf01.common-cnfdf01-sriov-sub-policy              enforce                                 3s
cnfdf01            group-cnfdf01.group-cnfdf01-console-policy                  enforce                                 3s
cnfdf01            group-cnfdf01.group-cnfdf01-log-forwarder-policy            enforce                                 3s
cnfdf01            group-cnfdf01.group-cnfdf01-log-policy                      enforce                                 3s
cnfdf01            group-cnfdf01.group-cnfdf01-mc-accelerated-startup-policy   enforce                                 3s
cnfdf01            group-cnfdf01.group-cnfdf01-mc-sctp-policy                  enforce                                 3s
cnfdf01            group-cnfdf01.group-cnfdf01-sno-network-policy              enforce                                 3s
cnfdf01            group-cnfdf01.group-cnfdf01-sriov-conf-policy               enforce                                 3s
cnfdf02-policies   cnfdf02-config-policy                                       enforce              Compliant          6d19h
cnfdf02            cnfdf02-policies.cnfdf02-config-policy                      enforce              Compliant          3s
cnfdf02            common-cnfdf02.common-cnfdf02-config-policy                 enforce              NonCompliant       3s
cnfdf02            group-cnfdf02.group-cnfdf02-config-policy                   enforce              NonCompliant       3s
common-cnfdf01     common-cnfdf01-log-sub-policy                               enforce                                 6d18h
common-cnfdf01     common-cnfdf01-mon-offload-policy                           enforce                                 6d18h
common-cnfdf01     common-cnfdf01-pao-sub-policy                               enforce                                 6d18h
common-cnfdf01     common-cnfdf01-ptp-sub-policy                               enforce                                 6d18h
common-cnfdf01     common-cnfdf01-sriov-sub-policy                             enforce                                 6d18h
common-cnfdf02     common-cnfdf02-config-policy                                enforce              NonCompliant       6d19h
group-cnfdf01      group-cnfdf01-console-policy                                enforce                                 6d18h
group-cnfdf01      group-cnfdf01-log-forwarder-policy                          enforce                                 6d18h
group-cnfdf01      group-cnfdf01-log-policy                                    enforce                                 6d18h
group-cnfdf01      group-cnfdf01-mc-accelerated-startup-policy                 enforce                                 6d18h
group-cnfdf01      group-cnfdf01-mc-sctp-policy                                enforce                                 6d18h
group-cnfdf01      group-cnfdf01-sno-network-policy                            enforce                                 6d18h
group-cnfdf01      group-cnfdf01-sriov-conf-policy                             enforce                                 6d18h
group-cnfdf02      group-cnfdf02-config-policy                                 enforce              NonCompliant       6d19h
```

Both `cnfdf02` and `cnfdf01` has the exact policySource but `cnfdf02` is using the new way of doing `policyName`.

But I'll need some help to see if they are actually working on the SNOs.

imiller0 (commented):
/lgtm

imiller0 (approved):
/lgtm
As discussed the subscription related CRs are grouped separately from the configuration (within common PGT). This results in 4 total policies being created:
common-subscriptions-policy
common-config-policy
group-du-sno-config-policy
example-sno-config-policy.

PR #840, opened 2021-12-07
ztp: fix ArgoCD deployment policies-app-project.yaml path

mcornea:
Fix ArgoCD deployment kustomization.yaml to include the correct policies-app-project.yaml path.

### DIFF
--- a/ztp/gitops-subscriptions/argocd/deployment/kustomization.yaml
+++ b/ztp/gitops-subscriptions/argocd/deployment/kustomization.yaml
@@ -3,7 +3,7 @@ kind: Kustomization
 
 resources:
   - app-project.yaml
-  - policy-app-project.yaml
+  - policies-app-project.yaml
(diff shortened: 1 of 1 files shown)


serngawy (approved):
thanks for fixing that

ztp/source-crs/AcceleratorsCatalogSource.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: operators.coreos.com/v1alpha1
```
serngawy, line 1: why adding this ? AND in which PGT (common, group, site) will be used ? There are already many CatalogSource exist under ztp/source-crs that can be overlay

mcornea: Please ignore this, it was picked by mistake. I removed it.

serngawy (approved):

PR #842, opened 2021-12-07
ztp: Fix siteConfig binary execution and  Add more unit test

serngawy:
ztp: Fix siteConfig binary execution and  Add more unit test


lack:
ERROR: gofmt reported formatting issues

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -225,3 +264,59 @@ func Test_getExtraManifestTemplatedRoles(t *testing.T) {
...
+	for clusterId, cluster := range sc.Spec.Clusters {
+		testClusterCRs := clustersCRs[sc.Metadata.Name+"/"+cluster.ClusterName]
+		assert.NotNil(t, testClusterCRs)
+		expectedCRs := make([]interface{}, 0)
+		for _, sourceCR := range scBuilder.SourceClusterCRs {
+			mapSourceCR := sourceCR.(map[string]interface{})
+			assert.NotNil(t, mapSourceCR)
+			if mapSourceCR["kind"] == "BareMetalHost" || mapSourceCR["kind"] == "NMStateConfig" {
+				for nodeId, _ := range cluster.Nodes {
+					expectedCR, err := scBuilder.getClusterCR(clusterId, sc, mapSourceCR, nodeId)
```
lack, line 298: I'm dubious about this "call scBuilder to test scBuilder" methodology - What exactly are we trying to test here?

ztp/siteconfig-generator/main.go
```diff
@@ -11,13 +11,17 @@ import (
 )
 
 func main() {
+	localExtraManifestPath := flag.String("manifestPath", "", "Directory where pre-defined extra manifest exist")
```
lack, line 14: Minor language fix-up suggestion:
```suggestion
	localExtraManifestPath := flag.String("manifestPath", "", "Directory with pre-defined extra manifests")
```

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -225,3 +264,59 @@ func Test_getExtraManifestTemplatedRoles(t *testing.T) {
 		}
 	}
 }
+
+func Test_SNOClusterSiteConfigBuild(t *testing.T) {
+	sc := SiteConfig{}
+	err := yaml.Unmarshal([]byte(siteConfigTest), &sc)
+	assert.Nil(t, err)
```
lack, line 271: It's better for errors to use `assert.NoError(t, err)` - If it fails, it's easier to see why.
```suggestion
	assert.NoError(t, err)
```

Note: There are a lot of other instances of this same thing through these unit tests, and related instances such as `assert.NotEqual(t, err, nill)` which should be `assert.Error(t, err)`

lack (changes requested):

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
serngawy: I added the expected output from those test siteconfigs in order to make sure that whatever changes happen in the future can be validated for the same given siteconfigs.

ztp/siteconfig-generator/main.go
serngawy: done

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
serngawy: done

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
lack: I see that the outer tests seem to be doing useful things, but here's what I see in this function:
1. we call `scBuilder.Build` to make all the cluster CRs
2. For each cluster we call `scBuilder.getClusterCr`, basically duplications what the `scBuilder.Build` did
3. We check that the test logic around  `scBuilder.getClusterCr` made the same thing that `scBuilder.Build` made
4. We marshal the result and that's what the caller checks against.

I think the check you mentioned on the expected SiteConfig relies on doing #1 and #4, but I don't think we need #2 or #3.  They just seem to be checking that our `scBuilder.Build` matches our local duplications of the code in this function.  Am I missing something?

serngawy: Steps 2 and 3 for testing the business logic changes in `scBuilder.Build` AND `scBuilder.getClusterCrs`. if there is a change in the func the unitTest will fail before comparing the end result output.  [L336](https://github.com/openshift-kni/cnf-features-deploy/pull/842/files#diff-debded0a9a2b5f548423222493ec145e61e597d7fc5df62c4bbd3320798351c1R336) . What do u think I can remove it and just check for the end output ?

serngawy: moved getClusterCR in another unitTest

lack (approved):

PR #844, opened 2021-12-08
ztp: BZ-2029181 Add reference of DisableSnoNetworkDiag.yaml to group-du-sno PolicyGenTemplate

pixelsoccupied:
New source-crs `DisableSnoNetworkDiag` is now added to example group-du for ZTP


imiller0 (approved):
/lgtm

PR #846, opened 2021-12-08
cnf-tests: Fix ovs_qos flake

oribon:
Recently we're seeing 'Validate MCO removed egress/ingress'
tests flaking. From inspecting the logs the following
situation can happen in a test:

We delete the ingress MC, we see that the QoS is deleted from
the node, then we try to validate that the ingress limitation is
really deleted by creating the sender-receiver pods again.
Then we check that the traffic rate that arrives to the
receiver surpasses the ingressBitLimit that was set before.
This could be misleading, because if for some reason
the sender sends in a rate below what was set as the
ingressBitLimit, the test would fail thinking that
the ingress QoS still applies in the node.

Setting the limitMultiplier lower could significantly
reduce the chances of that happening.

Maybe TODO: redo how we check that the rate set by
the QoS is enforced/not enforced.

Signed-off-by: Ori Braunshtein <obraunsh@redhat.com>

### DIFF
--- a/cnf-tests/testsuites/e2esuite/ovs_qos/ovs_qos.go
+++ b/cnf-tests/testsuites/e2esuite/ovs_qos/ovs_qos.go
@@ -52,7 +52,7 @@ const (
 	// IngressMCName contains the name of the ingress MC applied by the ovs_qos tests
 	IngressMCName = "qos-ingress"
 	// LimitMultiplier is the multiplier to get the MC limit rate without discovery
-	LimitMultiplier = 0.7
+	LimitMultiplier = 0.3 // TODO: see commit message
 )
 
 var (


yuvalk:
maybe you should test the rate before and after and compare,
so you'll know if the rate change is because qos or something else

lack:
/lgtm

For a band-aid it's okay, but I agree revisiting this with a better check is important.

oribon:
> maybe you should test the rate before and after and compare, so you'll know if the rate change is because qos or something else

Not sure I understand what you're suggesting, anyways I'm writing this for reference later: the tests currently work by determining the rate limit for the MachineConfig by sending/receiving traffic when they're not limited to get an idea what's the "max" rate and applying that rate*limitMultiplier to it. Then the tests validate that the rate is enforced under different scenarios. Finally the tests remove the MC and validate that all is back to normal, by sending/receiving again and comparing the rate sent/received to what was set in the (now deleted) MachineConfig before, which could be problematic as I stated in the commit message.



> /lgtm
> 
> For a band-aid it's okay, but I agree revisiting this with a better check is important.

Thanks, I agree this probably shouldn't be a permanent solution + we should see if it actually solves the flake.

yuvalk:
/approve

but let's keep the story open

PR #850, opened 2021-12-08
features: Move our mount namespace execstart last

lack:
There is a collision with a drop-in owned by MCO which also edits crio's
commandline by replacing ExecStart.  It turns out that the drop-in is
redundant with the environment variable that's already being set, so it
is safe for us to move our ExecStart later in the drop-in order and
simply replace the incompatible one.


lack:
The associated PR in MCO that removes the redundant ExecStart entirely is https://github.com/openshift/machine-config-operator/pull/2858

imiller0 (commented):
Same comments as in the related PR #849. 
https://github.com/openshift-kni/cnf-features-deploy/pull/849#pullrequestreview-828212184
/lgtm

PR #852, opened 2021-12-09
ztp: Clean up PtpConfig source-crs

lack:
This adds default command options.  They can still be overridden by PGT.

This change is backwards-compatible with previous PGT uses.


vitus133 (changes requested):
Hi,
I think there is a copy-paste error in PtpConfigMaster.yaml. After checking in our old reference code, I see for GM
```yaml
ptp4lOpts: "-2 --summary_interval 0"
phc2sysOpts: "-a -r -r -n 24"
```
For the slave it looks correct
Thanks!

lack:
@vitus133 : Updated!  Thanks for catching that!

PR #853, opened 2021-12-09
ztp: Clean up StorageLV source-crs and SNO example

lack:
This makes the example in source-crs for StorageLV a bit more generic,
with a group policy example that looks more like what customers would
likely be doing.

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
@@ -59,5 +59,17 @@ spec:
         disableDrain: true
     - fileName: StorageLV.yaml
       policyName: "local-disks-policy"
+      spec:
+        storageClassDevices:
+          - storageClassName: "example-storage-class-1"
+            volumeMode: Filesystem
+            fsType: xfs
+            devicePaths:
+              - /dev/sdb1
+          - storageClassName: "example-storage-class-2"
+            volumeMode: Filesystem
+            fsType: xfs
+            devicePaths:
+              - /dev/sdb2
     - fileName: DisableSnoNetworkDiag.yaml
       policyName: "config-policy"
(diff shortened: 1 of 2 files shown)


ztp/source-crs/StorageLV.yaml
```diff
@@ -7,7 +7,8 @@ spec:
   logLevel: Normal
   managementState: Managed
   storageClassDevices:
-    - storageClassName: "fs-sc"
+    # The list of storage classes and associated devicePaths need to be specified like this example:
```
serngawy, line 10: why changing the storageClassName ? does changing the group-du-sno-ranGen.yaml not enough ? 
I'm just worry if someone use the StorageLV.yaml as it is, you force him to change the name. what do you think ?

imiller0: I think we are OK here. The StorageLV was added during 4.10 and has not yet been backported to 4.9

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -59,5 +59,17 @@ spec:
...
+          - storageClassName: "example-storage-class-1"
+            volumeMode: Filesystem
+            fsType: xfs
+            devicePaths:
+              - /dev/sdb1
+          - storageClassName: "example-storage-class-2"
+            volumeMode: Filesystem
+            fsType: xfs
+            devicePaths:
+              - /dev/sdb2
```
imiller0, line 73: We should tailor the reference PGTs to the expected use case. I don't know how many classes might be typical for a RAN deployment but my guess would be 1. @browsell any insights?

lack: From what I recall there is no "typical" storage config; everyone will want something different.

ztp/source-crs/StorageLV.yaml
lack: The point of changing the source-crs is that no one should use the default value.  Storage layout is too user-specific and they must always put their config in their PGT.

serngawy: okay, if the back compatibility issue I guess its good then

serngawy (approved):

PR #855, opened 2021-12-09
ztp: move sno PerformanceProfile to group

lack:
Performance profiles will generally apply to an entire group of machines, so let's model this.

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
@@ -73,3 +73,34 @@ spec:
               - /dev/sdb2
     - fileName: DisableSnoNetworkDiag.yaml
       policyName: "config-policy"
+    - fileName: PerformanceProfile.yaml
+      policyName: "config-policy"
+      metadata:
+        name: openshift-node-performance-profile
+      spec:
+        cpu:
+          # These must be tailored for the specific hardware platform
+          isolated: "2-19,22-39"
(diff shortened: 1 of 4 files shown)


lack:
Holding until #853 merges, since this builds upon that work.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -59,5 +59,48 @@ spec:
...
+            devicePaths:
+              - /dev/sdb1
+          - storageClassName: "example-storage-class-2"
+            volumeMode: Filesystem
+            fsType: xfs
+            devicePaths:
+              - /dev/sdb2
     - fileName: DisableSnoNetworkDiag.yaml
       policyName: "config-policy"
+    - fileName: PerformanceProfile.yaml
```
imiller0, line 76: Moving this here will require the wave defined in PR 862 to be updated to match the other CRs in the group.

lack: Come to think of it, this MUST change because the multi-node group examples have the PerformanceProfile in their group PGTs.

imiller0 (approved):
/lgtm
With respect to backwards compatibility, anyone who picks up these new reference PGTs will automatically get this change and things will work (assuming they pick up both the group and site-specific references but that seems safe to assume). Anyone who doesn't use these new reference PGTs up will continue on with their PerfProfile and TunedConfig in the site-specific policies. Those older PGTs had these CRs in their own policies (not grouped) so the wave's won't matter.

PR #858, opened 2021-12-09
ztp: Add an initial example of a DU profile

lack:
Interestingly enough, the main differences between SNO and multi-node
site configurations are in the 'group-du-ranGen' vs
'group-du-sno-ranGen' level.  The actual site-specifics look very
similar (apart from the MCP role name)


lack:
Tested - The resulting CRs look pretty reasonable and work well on a standard cluster.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-standard-ranGen.yaml
```diff
@@ -0,0 +1,55 @@
+---
+apiVersion: ran.openshift.io/v1
+kind: PolicyGenTemplate
+metadata:
+  name: "group-du"
```
serngawy, line 5: duplicated name as group-du-3node-ranGen.yaml .

serngawy (commented):
policies sounds good to me, just make sure u can run policyGen kustomize plugin without raising errors  AND add those files to kustomization.yaml with comment or notes

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-standard-ranGen.yaml
lack: Good catch!  I just found I had to introduce a differentiation between the 3-node and standard clusters.  I'll rename to match.

lack: Done.

lack:
@serngawy I have updated the example kustomization.yaml and when I run the policygenerator kustomize on it, it does not error and creates all policies as expected.

serngawy (approved):

PR #862, opened 2021-12-13
ztp: Add ztp-deploy-wave annotation in source CR

Missxiaoguo:
Add 'ran.openshift.io/ztp-deploy-wave' annotation in source CR
to indicate its deploy order. It will be used to set policy's
'ran.openshift.io/ztp-deploy-wave' annotation.

A check added in Makefile to make sure our provided sourceCRs have this wave annotation.

Resources have same wave means they can be applied in one policy.
Resources have different waves means they need to be applied in order and in different policies.

The default resource waves are set according to [PR836](https://github.com/openshift-kni/cnf-features-deploy/pull/836/) which are based on common, group-du and site. We can adjust the default waves if some resources in one group(common/group-du/site) need to be applied in order.

Wave 1(common):
    Catalogsources
    DisconnectedICSP
    Disableoperatorhub
    ReduceMonitoringFootprint

Wave 2(common):
    OperatorNamespaces
    OperatorGroups
    OperatorSubscriptions
 
Wave 10(group-du):
    ClusterLogForwarder
    ClusterLogging
    consoleOperatorDisable
    ptpConfigSlave
    ptpConfigMaster
    ptpConfigSlaveCvl
    ptpOperatorConfigForEvent
    sriovOperatorConfig
    DisableSnoNetworkDiag
    StorageLV
 
 Wave 100(site specifics):
    Amqinstance
    performanceProfile
    SriovNetwork
    SriovNetworkNodePolicy
    TunedPerformancePatch

Jira: https://issues.redhat.com/browse/CNF-3635
/cc @imiller0 @serngawy @browsell

### DIFF
--- a/ztp/Makefile
+++ b/ztp/Makefile
@@ -1,6 +1,6 @@
-.PHONY: ci-job test-policygen checkExtraManifests test-policygen-kustomize test-siteconfig test-siteconfig-kustomize
+.PHONY: ci-job test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize
 
-ci-job: test-policygen checkExtraManifests test-policygen-kustomize test-siteconfig test-siteconfig-kustomize
+ci-job: test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize
 
 test-policygen:
 	@echo "ZTP: Build policy generator and run test"
@@ -8,6 +8,16 @@ test-policygen:
 
 checkExtraManifests:
 	$(MAKE) -C ./extra-manifests-builder check
+
+checkSourceCRsAnnotation:
+	for sourcefile in ./source-crs/*.yaml;do \
+		if [[ "$${sourcefile}" != *"MachineConfig"* ]];then \
+			if ! grep -qE "ran.openshift.io/ztp-deploy-wave" "$${sourcefile}";then \
+				echo "Error: missing annotation 'ran.openshift.io/ztp-deploy-wave' in $${sourcefile}"; \
+				exit 1; \
+			fi; \
+		fi; \
+	done; \
 
 test-policygen-kustomize:
 	@echo "ZTP: Build policy generator kustomize plugin and run test"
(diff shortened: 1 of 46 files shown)


ztp/source-crs/ClusterLogSubscription.yaml
```diff
@@ -3,6 +3,8 @@ kind: Subscription
 metadata:
   name: cluster-logging
   namespace: openshift-logging
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "1"
```
browsell, line 7: Need to confirm that a subscription will eventually get reconciled if the ns/operatorgroup has not. This use to be a problem need to confirm it has been resolved

browsell:
See the comment I made in 836. I think we want to seperate config from s/w installation, i.e. the subscription related cr's have their own policy

ztp/Makefile
```diff
@@ -9,6 +9,16 @@ test-policygen:
 checkExtraManifests:
 	$(MAKE) -C ./extra-manifests-builder check
 
+checkSourceCRsAnnotation:
+	for sourcefile in ./source-crs/*.yaml;do \
+		if [[ "$${sourcefile}" != *"MachineConfig"* ]];then \
```
lack, line 14: Why not check MachineConfig files?

We have symlinks in source-crs for backwards compatibility, so it's possible users might still include them within as policy.

What happens if there's no annotation on a CR in a PGT?

Missxiaoguo: Based on the discussion this morning, no need waves for MCs we want to deprecate the use of MCs after installation via policies. We need to consider how to deprecate.

To the question that "What happens if there's no annotation on a CR in a PGT"
One policy should only have sources with same waves, otherwise, it's an error. This includes a CR doesn't have wave but others have.
If policy has no wave which means its sourceCRs do not have wave, this policy will not be managed by lifecycle operator.

Missxiaoguo:
> See the comment I made in 836. I think we want to seperate config from s/w installation, i.e. the subscription related cr's have their own policy

updated

imiller0 (approved):
/lgtm
On the question on subscriptions resolving when namespace and operator group not yet in place. I believe this is referencing BZ 1960455 which was fixed and verified in 4.9.

PR #863, opened 2021-12-13
ztp: Add ztp-deploy-wave annotation in generated policy

Missxiaoguo:
In order to make ClusterGroupUpgrade operator determine the policies
deploy order, 'ran.openshift.io/ztp-deploy-wave' annotation is added
to each policy according to the sourceCRs wave annotation.
See, https://github.com/openshift-kni/cnf-features-deploy/pull/862

Resources have different waves means they need to be applied in order, so
one policy cannot have resources with different waves. PolicyGen will raise
error if it happens.
The wave annotation can be overwritten/added via PGT. 

Jira: https://issues.redhat.com/browse/CNF-3635
/cc @imiller0 @serngawy @browsell

### DIFF
--- a/ztp/policygenerator/policyGen/policyBuilder.go
+++ b/ztp/policygenerator/policyGen/policyBuilder.go
@@ -151,7 +151,6 @@ func (pbuilder *PolicyBuilder) Build(policyGenTemp utils.PolicyGenTemplate) (map
 func (pbuilder *PolicyBuilder) getCustomResources(sFile utils.SourceFile, sourceCRFile []byte, mcp string) ([]map[string]interface{}, error) {
 	yamls, err := pbuilder.splitYamls(sourceCRFile)
 	resources := make([]map[string]interface{}, 0)
-
 	if err != nil {
 		return resources, err
 	}
@@ -198,11 +197,34 @@ func (pbuilder *PolicyBuilder) getCustomResource(sourceFile utils.SourceFile, so
 		resourceMap["metadata"].(map[string]interface{})["namespace"] = sourceFile.Metadata.Namespace
 	}
 	if len(sourceFile.Metadata.Labels) != 0 {
-		resourceMap["metadata"].(map[string]interface{})["labels"] = sourceFile.Metadata.Labels
+		// convert from map[string]string to map[string]interface{}
+		sourceFileLabels := make(map[string]interface{})
+		for k, v := range sourceFile.Metadata.Labels {
+			sourceFileLabels[k] = v
+		}
+
+		if resourceMap["metadata"].(map[string]interface{})["labels"] == nil {
+			// source CR does not have labels
+			resourceMap["metadata"].(map[string]interface{})["labels"] = make(map[string]interface{})
+		}
+		resourceMap["metadata"].(map[string]interface{})["labels"] = pbuilder.setValues(
+			resourceMap["metadata"].(map[string]interface{})["labels"].(map[string]interface{}), sourceFileLabels)
 	}
 	if len(sourceFile.Metadata.Annotations) != 0 {
-		resourceMap["metadata"].(map[string]interface{})["annotations"] = sourceFile.Metadata.Annotations
+		// convert from map[string]string to map[string]interface{}
+		sourceFileAnnotations := make(map[string]interface{})
+		for k, v := range sourceFile.Metadata.Annotations {
+			sourceFileAnnotations[k] = v
+		}
+
+		if resourceMap["metadata"].(map[string]interface{})["annotations"] == nil {
+			// source CR does not have annotations
+			resourceMap["metadata"].(map[string]interface{})["annotations"] = make(map[string]interface{})
+		}
+		resourceMap["metadata"].(map[string]interface{})["annotations"] = pbuilder.setValues(
+			resourceMap["metadata"].(map[string]interface{})["annotations"].(map[string]interface{}), sourceFileAnnotations)
 	}
+
 	if resourceMap["spec"] != nil {
 		resourceMap["spec"] = pbuilder.setValues(resourceMap["spec"].(map[string]interface{}), sourceFile.Spec)
 	} else if sourceFile.Spec != nil {
@@ -317,12 +339,19 @@ func (pbuilder *PolicyBuilder) createAcmPolicy(name string, namespace string, re
 	}
 	objTempArr := make([]utils.ObjectTemplates, len(resources))
 
+	isFirstOjbTemp := true
 	for idx, resource := range resources {
 		objTemp := BuildObjectTemplate(resource)
 		objTempArr[idx] = objTemp
+
+		if idx != 0 {
+			isFirstOjbTemp = false
+		}
+		if err = SetPolicyDeployWave(acmPolicy.Metadata, resource, isFirstOjbTemp); err != nil {
+			return acmPolicy, err
+		}
 	}
 	acmPolicy.Spec.PolicyTemplates[0].ObjDef.Spec.ObjectTemplates = objTempArr
-
 	return acmPolicy, nil
 }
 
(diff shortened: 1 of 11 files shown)


serngawy (changes requested):
this is not needed in PGT , just compute the order in the upgrade operator AND no need to annotate the policy all what u need to is to set the order in the generated UOCR

Missxiaoguo:
> this is not needed in PGT , just compute the order in the upgrade operator AND no need to annotate the policy all what u need to is to set the order in the generated UOCR

We don't allow one policy to have resources with different waves. PGT should validate this when building the policy.  And also adding wave in policy is much easier for upgrade operator to query.

ztp/policygenerator/policyGen/policyBuilder.go
```diff
@@ -320,9 +331,12 @@ func (pbuilder *PolicyBuilder) createAcmPolicy(name string, namespace string, re
 	for idx, resource := range resources {
 		objTemp := BuildObjectTemplate(resource)
 		objTempArr[idx] = objTemp
+
+		if err = SetPolicyDeployWave(acmPolicy.Metadata, len(objTempArr), resource); err != nil {
```
imiller0, line 335: Maybe double-check the use of len(objTempArr) here. On line 329 the array is constructed using the len(resources). I believe this will call SetPolicyDeployWave with the same value for this second arg on each iteration. From the code below it looks like you want to call it with '1' the first iteration and any other value on subsequent iterations. Maybe pass "idx+1"?

ztp/policygenerator/policyGen/policyBuilder_test.go
```diff
@@ -479,3 +480,117 @@ spec:
...
+
+	// Run the PGT through the generator
+	pBuilder = NewPolicyBuilder(fHandler)
+	policies, err = pBuilder.Build(pgt)
+
+	// Validate the run
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "doesn't match with Policy")
+	assert.NotNil(t, policies)
+}
```
imiller0, line 596: Consider adding two more tests

1. If the user has an overlay that changes a wave so that the policy ends up with multiple values
2. If the user has an overlay that "fixes" a mismatched set of waves in the source.

ztp/policygenerator/policyGen/policyBuilder.go
```diff
@@ -201,8 +200,20 @@ func (pbuilder *PolicyBuilder) getCustomResource(sourceFile utils.SourceFile, so
 		resourceMap["metadata"].(map[string]interface{})["labels"] = sourceFile.Metadata.Labels
 	}
 	if len(sourceFile.Metadata.Annotations) != 0 {
-		resourceMap["metadata"].(map[string]interface{})["annotations"] = sourceFile.Metadata.Annotations
+		// covert from map[string]string to map[string]interface{}
```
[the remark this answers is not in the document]
Missxiaoguo: Done

ztp/policygenerator/policyGen/policyBuilder.go
```diff
@@ -201,8 +200,20 @@ func (pbuilder *PolicyBuilder) getCustomResource(sourceFile utils.SourceFile, so
...
+		for k, v := range sourceFile.Metadata.Annotations {
+			sourceFileAnnotations[k] = v
+		}
+
+		if resourceMap["metadata"].(map[string]interface{})["annotations"] == nil {
+			// source CR does not have annotations
+			resourceMap["metadata"].(map[string]interface{})["annotations"] = make(map[string]interface{})
+		}
+		resourceMap["metadata"].(map[string]interface{})["annotations"] = pbuilder.setValues(
+			resourceMap["metadata"].(map[string]interface{})["annotations"].(map[string]interface{}), sourceFileAnnotations)
```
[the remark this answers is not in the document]
Missxiaoguo: I agree. I updated the overlay for labels as well and related unittests.

ztp/policygenerator/policyGen/policyHelper.go
```diff
@@ -80,3 +81,42 @@ func BuildObjectTemplate(resource generatedCR) utils.ObjectTemplates {
...
+				return nil
+			}
+		} else if !foundCrDeployWave {
+			// no wave for policy and resource
+			return nil
+		}
+
+		// resource wave doesn't match with policy wave
+		// resource has no wave but policy has or vice versa
+		return errors.New("'ran.openshift.io/ztp-deploy-wave' annotation in Resource " +
```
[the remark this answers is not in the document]
Missxiaoguo: Done

ztp/policygenerator/policyGen/policyBuilder.go
Missxiaoguo: You are obsoultely right! Thanks for finding this out. Updated.

ztp/policygenerator/policyGen/policyBuilder_test.go
Missxiaoguo: Added.

ztp/policygenerator/policyGen/policyBuilderOverlay_test.go
```diff
@@ -74,6 +75,19 @@ spec:
...
+	assert.NotNil(t, annotations)
+	assert.Equal(t, len(annotations), 2)
+	assert.Equal(t, annotations["key1"], "value1")
+	assert.Equal(t, annotations["key2"], "value2")
+
+	labels := objDef["metadata"].(map[string]interface{})["labels"].(map[string]interface{})
+	assert.NotNil(t, labels)
+	assert.Equal(t, len(labels), 2)
+	assert.Equal(t, annotations["key1"], "value1")
+	assert.Equal(t, annotations["key2"], "value2")
```
imiller0, line 90: Did you intend to validate the labels here rather than the annotations again?

ztp/policygenerator/policyGen/policyBuilderOverlay_test.go
```diff
@@ -103,6 +117,13 @@ spec:
   sourceFiles:
     - fileName: GenericCR.yaml
       policyName: "gen-policy1"
+      metadata:
+        labels:
+          key1: newvalue
+          key3: value3
```
imiller0, line 123: Use key/values for labels that are different from annotations to catch any bugs where the code accidentally uses the wrong datastructure.

ztp/policygenerator/policyGen/policyBuilderOverlay_test.go
```diff
@@ -310,6 +346,11 @@ spec:
   sourceFiles:
     - fileName: GenericDataCR.yaml
       policyName: "gen-policy1"
+      metadata:
+        labels:
+          key1: value1
```
imiller0, line 351: Same note on unique values.

ztp/policygenerator/policyGen/testData/GenericSourceFiles/GenericCR.yaml
```diff
@@ -4,6 +4,12 @@ kind: JustForTest
 metadata:
   name: test123
   namespace: generic-ns
+  labels:
+    key1: value1
+    key2: value2
```
imiller0, line 9: See note above on unique values.

ztp/source-crs/validatorCRs/informDuValidator.yaml
```diff
@@ -27,10 +27,10 @@ status:
         name: load-sctp-module-master
       - apiVersion: machineconfiguration.openshift.io/v1
         kind: MachineConfig
-        name: 02-master-workload-partitioning-master
+        name: 02-master-workload-partitioning
```
imiller0, line 30: We should factor this out into a separate PR in case it needs to be backported.

ztp/policygenerator/policyGen/policyBuilderOverlay_test.go
Missxiaoguo: Good catch!! Thank you. Will update

ztp/policygenerator/policyGen/policyBuilderOverlay_test.go
Missxiaoguo: Updated

ztp/source-crs/validatorCRs/informDuValidator.yaml
Missxiaoguo: Sure. will create a separate PR for this. Removed from this PR.

imiller0 (approved):
/lgtm

PR #866, opened 2021-12-14
Add biosConfigMap annotation to BareMetalHost

tliu2021:
The desired BIOS configuration can be provided in a YAML file which
will be applied as a ConfigMap. The biosConfigMapRef is added to the
SiteConfig to specify the ConfigMap name. This ConfigMap is consumed
by the baremetal-operator to configure the BIOS on the baremetal host.


ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -189,6 +189,7 @@ type Nodes struct {
 	InstallerArgs          string                 `yaml:"installerArgs"`
 	IgnitionConfigOverride string                 `yaml:"ignitionConfigOverride"`
 	Role                   string                 `yaml:"role"`
+	BiosConfigMapRef       BiosConfigMapRef       `yaml:"biosConfigMapRef"`
```
serngawy, line 192: Would you also define this to the cluster level, So if the cluster has same nodes don't need to keep define it per node.
AND the node definition override the cluster definition if both defined.

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
```diff
@@ -39,6 +39,8 @@ spec:
           thumbprint: "1234567890"
     nodes:
       - hostName: "node1"
+        biosConfigMapRef:
```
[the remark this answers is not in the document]
tliu2021: That's a very good suggestion. I will make changes accordingly.

ztp/siteconfig-generator/siteConfig/siteConfig.go
tliu2021: Yes, thanks for the suggestion. I will add BiosConfigMapRef to the cluster level as well.

ztp/gitops-subscriptions/argocd/example/siteconfig/hp.profile
```diff
@@ -0,0 +1,6 @@
+BootMode: Uefi
```
browsell, line 1: Wouldn't we drop this in as extra manifests

tliu2021: Yes, that's another option.

tliu2021: I can move it to extra-manifest.

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
```diff
@@ -13,6 +13,8 @@ spec:
     name: "sshPrvKey"
   clusters:
   - clusterName: "site-sno-du-1"
+    biosConfigMapRef:
+      name: "testSiteConfig/hp.profile"
```
imiller0, line 17: For test data we should keep it generic and not reference a specific hardware vendor. Maybe testHW.profile?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -161,6 +170,31 @@ func (scbuilder *SiteConfigBuilder) getClusterCR(clusterId int, siteConfigTemp S
...
 }
 
+func (scbuilder *SiteConfigBuilder) getBiosConfigMapCR(clusterId int, siteConfigTemp SiteConfig, nodeId int) (interface{}, error) {
+
+	// if BiosConfigMapRef is not set for the host, use the one set for the cluster
+	if siteConfigTemp.Spec.Clusters[clusterId].Nodes[nodeId].BiosConfigMapRef.Name == "" {
+		siteConfigTemp.Spec.Clusters[clusterId].Nodes[nodeId].BiosConfigMapRef = siteConfigTemp.Spec.Clusters[clusterId].BiosConfigMapRef
+	}
+
+	annotations := map[string]string{"argocd.argoproj.io/sync-wave": "1"}
```
imiller0, line 180: We should define all of the waves as constants and keep those constants together in one place so that it is easy to look at all mappings of CRs to waves in one place.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -161,6 +170,31 @@ func (scbuilder *SiteConfigBuilder) getClusterCR(clusterId int, siteConfigTemp S
 	return mapIntf, nil
 }
 
+func (scbuilder *SiteConfigBuilder) getBiosConfigMapCR(clusterId int, siteConfigTemp SiteConfig, nodeId int) (interface{}, error) {
```
imiller0, line 173: Minor nit. This method should be "buildBiosConfigMapCR" (or similar). Using "get" for the name makes it sound like this is simply looking it up and returning it, but this method contructs a new ConfigMap and returns it.

imiller0 (commented):
If BIOS configuration is not part of the 4.10 release we should hold off on merging this until the branch is cut.

ztp/gitops-subscriptions/argocd/example/siteconfig/hp.profile
tliu2021: The extra-manifests are used to generate the ConfigMap that is consumed by assisted-service to populate the MachineConfig for the spoke clusters. The BIOS configuration file is used to generate the ConfigMap that is consumed by baremetal-operator to populate the Spec fields.
It is prudent to take a closer look at this option.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
tliu2021: Agree.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -161,6 +170,31 @@ func (scbuilder *SiteConfigBuilder) getClusterCR(clusterId int, siteConfigTemp S
...
+	biosFile := siteConfigTemp.Spec.Clusters[clusterId].Nodes[nodeId].BiosConfigMapRef.Name
+	fileData, err := ReadFile(biosFile)
+	if err != nil {
+		return biosCM, err
+	}
+	err = yaml.Unmarshal(fileData, biosCM.Data)
+	if err != nil {
+		return biosCM, err
+	}
+	siteConfigTemp.Spec.Clusters[clusterId].Nodes[nodeId].BiosConfigMapRef.Name = siteConfigTemp.Spec.Clusters[clusterId].Nodes[nodeId].HostName
```
tliu2021, line 194: I am thinking using a host name as the generated config map name is not ideal as it could clash with the extra manifest config map, when the host name and cluster are the same.  May we ask the user to enter a config map name or generate a unique name for each host?

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
tliu2021: Yeah.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -232,3 +243,8 @@ type Interfaces struct {
 	Name       string `yaml:"name"`
 	MacAddress string `yaml:"macAddress"`
 }
+
+// BiosConfigMapRef
+type BiosConfigMapRef struct {
+	Name string `yaml:"name"`
```
serngawy, line 249: As we discussed , we better let the user enter the configMap name and add another field for the filePath.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -118,6 +118,15 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
 			clusterCRs = append(clusterCRs, crValue)
 		} else if mapSourceCR["kind"] == "BareMetalHost" || mapSourceCR["kind"] == "NMStateConfig" {
 			for ndId := range siteConfigTemp.Spec.Clusters[clusterId].Nodes {
+				if mapSourceCR["kind"] == "BareMetalHost" &&
```
serngawy, line 121: the loop will re-create the configMap for every node even if all of them use the same configMap defined in the cluster level. I guess better to check only for the node here and then level up the cluster check out of the loop.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: yes, let the user enter it

serngawy (commented):
please add unit test as well to the PR.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
tliu2021: Defining all the waves as constants will be addressed in a followup PR.

tliu2021:
/on hold till the 4.10 branch is cut

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/testHW.profile
```diff
@@ -0,0 +1,6 @@
+BootMode: Uefi
```
lack, line 1: We do have a `bootMode` field in SiteConfig as well - What will happen if this line and that one hold different values?

ztp/siteconfig-generator/siteConfig/clusterCRs.go
```diff
@@ -105,6 +105,7 @@ metadata:
     bmac.agent-install.openshift.io/installer-args: "{{ .Node.InstallerArgs }}"
     bmac.agent-install.openshift.io/ignition-config-overrides: "{{ .Node.IgnitionConfigOverride }}"
     bmac.agent-install.openshift.io/role: "{{ .Node.Role }}"
+    bmac.agent-install.openshift.io/deployType: ztp
```
lack, line 108: Does this mean some other piece of software is hard-coding the name "ztp" somewhere inside of it, and takes special action because of it?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -121,7 +131,19 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
 				if err != nil {
 					return clusterCRs, err
 				}
-
+				if kind == "BareMetalHost" {
+					// Check for note level BIOS configuration
+					if siteConfigTemp.Spec.Clusters[clusterId].Nodes[ndId].BiosConfigMapRef.FilePath != "" {
```
lack, line 136: If you like, I think you should be able to adapt a multi-level lookup scheme I added; see `node.CrTemplateSearch` and maybe we can just use similar logic to lookup the BiosConfigMapRef too, by sharing some of the code and the algorithm?

lack (changes requested):
We should probably also add this to the examples in `ztp/gitops-subscriptions/argocd/example/siteconfig/`

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/testHW.profile
tliu2021: Yes, there is a bootMode field in SiteConfig which is transformed to BMH host.Spec.BootMode. The BMH spec setting is only used in Ironic, and it does not overwrite what is configured on the host.  The bootMode field in the BIOS setting is  configured to the host.

ztp/siteconfig-generator/siteConfig/clusterCRs.go
tliu2021: Yes, it is assumed that the image-customization-controller (ICC) needs a way to tell whether it should build an image with 2 agents. It might be better to rename this annotation to something like “bmac.agent-install.openshift.io/agent-required:  true”

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
tliu2021: Sure, I will make changes accordingly.

ztp/siteconfig-generator/siteConfig/clusterCRs.go
serngawy: its better to keep the path way to identify the value something like "{{ .Node.BiosConfigMap.Path }}"

serngawy (changes requested):
Would you also squash the commits

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/testHW.profile
lack: So should we remove this from SiteConfig if it doesn't actually affect the host's configuration?

tliu2021: Yes, it is confusing to the user.

ztp/siteconfig-generator/siteConfig/clusterCRs.go
tliu2021: I was looking the [PoC for assisted installation](https://github.com/openshift/ironic-agent-image/pull/35).  I don't think this annotation is needed and I will remove it from my update.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -276,6 +298,39 @@ func translateTemplateKey(key string) (string, error) {
 	return "", fmt.Errorf("Key %q could not be translated", key)
 }
 
+func (scbuilder *SiteConfigBuilder) buildBiosConfigMapCR(clusterId int, siteConfigTemp SiteConfig, nodeId int) (interface{}, string, error) {
```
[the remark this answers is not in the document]
tliu2021: The config map generated from the extra-manifests is consumed by the assisted-service to populate the MachineConfig for the spoke clusters.  The data is mostly passed through. It would be difficult for the assisted-service to search for BIOS configuration in the same config map.

lack (approved):

PR #870, opened 2021-12-15
ztp: Add upgrade.md

serngawy:

### DIFF
--- a/ztp/gitops-subscriptions/argocd/Upgrade.md
+++ b/ztp/gitops-subscriptions/argocd/Upgrade.md
@@ -1,0 +1,43 @@
+## Steps to Upgrade from v4.9 [argocd/example](https://github.com/openshift-kni/cnf-features-deploy/tree/release-4.9/ztp/gitops-subscriptions/argocd/resource-hook-example) to v4.10 [argocd/example](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example)
+
+Follow the steps 1, 2 and 3 to prepare the hub cluster as [Readme](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/README.md) (Preparation of Hub cluster) section mentioned.
+ 
+**Upgrading the SiteConfig repository:**
+In the siteConfig repository follow the steps below:
+ 1. Delete the post-sync.yaml and pre-sync.yaml files.
+ 1. If the siteconfig*.yaml in the repository contain definition for the Namespace remove it.
+ 1. Update the kustomization.yaml file to include all SiteConfig yaml files in the `generators` section as below
+```
+generators:
+# list of all the siteConfig*.yaml exist in the repository
+- site1.yaml
+- site2.yaml
+- site3.yaml
+``` 
+ Save and push changes to the repository. The ArgoCD application will detect the changes that have been done to the SiteConfig git repository   
+ and re-sync the cluster installation CRS to the hub cluster. Note: the installed clusters will not be re provisioned.
+ 
+**Upgrading the policyGenTemplate repository:**
+In the policyGenTemplate repository follow the steps below:
+ 1. Delete the post-sync.yaml and pre-sync.yaml files.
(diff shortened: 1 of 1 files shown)


ztp/gitops-subscriptions/argocd/Upgrade.md
```diff
@@ -0,0 +1,38 @@
...
+# List of all the PolicyGenTemplate*.yaml exist in the repository
+- common-ranGen.yaml
+- group-du-sno-ranGen.yaml
+- example-sno-site.yaml
+
+resources:
+# List of Namespace*.yaml required by the PolicyGenTemplate*.yaml
+- ns.yaml
+``` 
+ Save and push changes to the repository
```
lack, line 35: What will happen when the user does this?

ztp/gitops-subscriptions/argocd/Upgrade.md
```diff
@@ -0,0 +1,38 @@
...
+generators:
+# list of all the siteConfig*.yaml exist in the repository
+- example-sno.yaml
+``` 
+ Save and push changes to the repository
+ 
+**Upgrading the policyGenTemplate repository:**
+At the policyGenTemplate repository follow those steps:
+ 1. Delete the post-sync.yaml and pre-sync.yaml files.
+ 1. If the PolicyGenTemplate*.yaml in the repository contain definition for the Namespace remove it and define the Namespace in separate file similar to the new [example/ns.yaml](https://github.com/openshift-kni/cnf-features-deploy/blob/release-4.10/ztp/gitops-subscriptions/argocd/example/policygentemplates/ns.yaml).
```
lack, line 20:
```suggestion
 1. If any PolicyGenTemplate*.yaml in the repository contains a namespace definition, remove it and define the Namespace in separate file similar to the new [example/ns.yaml](https://github.com/openshift-kni/cnf-features-deploy/blob/release-4.10/ztp/gitops-subscriptions/argocd/example/policygentemplates/ns.yaml).
```

ztp/gitops-subscriptions/argocd/Upgrade.md
```diff
@@ -0,0 +1,38 @@
...
+ 1. Adjust the Namespace name to start with prefix  `ztp` .
```
lack, line 21:
```suggestion
 1. Adjust all Namespace names to start with prefix  `ztp`.
```
Maybe add an example? Also, is this still needed?

lack (changes requested):

imiller0:
Nice document so far. We should add content to describe other changes (both required and recommended changes). Some areas to include:

1. We have reduced the number of policies. How would the user adjust an existing PolicyGenTemplate to use our recommended grouping of policies
2. We have removed MachineConfigs from the PGT (applied as part of installation)
3. Any other changes that the user needs to make to existing PGT or SiteConfig

serngawy:
> Nice document so far. We should add content to describe other changes (both required and recommended changes). Some areas to include:
> 
> 1. We have reduced the number of policies. How would the user adjust an existing PolicyGenTemplate to use our recommended grouping of policies
> 2. We have removed MachineConfigs from the PGT (applied as part of installation)
> 3. Any other changes that the user needs to make to existing PGT or SiteConfig

Agree, we should add the changes for the policies reduce but let's do it later in another PR especially the  policy reduce PR didn't get merged yet.

ztp/gitops-subscriptions/argocd/Upgrade.md
```diff
@@ -0,0 +1,43 @@
...
+- common-ranGen.yaml
+- group-du-sno-ranGen.yaml
+- example-sno-site.yaml
+
+resources:
+# List of Namespace*.yaml required by the PolicyGenTemplate*.yaml
+- ns.yaml
+``` 
+ Save and push changes to the repository. The ArgoCD application will detect the changes that have been done to the PolicyGenTemplates git repository   
+ and re-sync the ACM policies to the hub cluster. Note: The ACM policies will be recreated under the new Namesapces.
```
lack, line 39: Do users need to delete the old policies from the old namespaces?  Will any changes be pushed out to the clusters immediately? What about the new integration with the LifecycleOperator?

serngawy: No need to delete any thing ArgoCD App take care of that for both PGT & siteConfig. Delete the policies does not affect the provisioned cluster the new policies will get created and then shows complaint or not same as the old one.
regrds LifecycleOperator let's have it work first then we will documented.

PR #874, opened 2021-12-21
[release 4.8] Bump intel operator api

SchSeba:


SchSeba:
```
run command 'oc [exec -i -n openshift-machine-config-operator -c machine-config-daemon --request-timeout 30 machine-config-daemon-qhtt2 -- /bin/bash -c taskset -pc 33]' (err=<nil>):
  stdout=""
```

looks like a known bug when we did not get the return from the exec command there was a fix in PAO for the tests that don't get backported. I will override the lane

PR #881, opened 2021-12-23
[Release 4.7] bump CI image version

SchSeba:


SchSeba:
This openshift-ci image we are using is not working.
we need to bump the golang version in the build image but the image is using the branch HEAD so we need to force this merge.
I will override all the lanes

PR #892, opened 2022-01-05
oot-driver: use DockerImage insted of ImageStreamTag for the driverContainer

sudomakeinstall2:
This commit adds environment variable USE_DOCKER_IMAGE. Setting this environment variable replaces ImageStreamTag with DockerImage. 

Also changes tag of the oot-source-driver and removed the nested namespace. This allows us to use quay which doesn't support nested namespaces.


SchSeba:
the change is not related to the failing tests.

/override ci/prow/e2e-gcp
/lgtm
/approve

PR #900, opened 2022-01-07
Bug 2036990: Adjust ztp-done policy to work with multi-node clusters (plus 50-nto-* MC)

lack:
- ztp: ztp-done policy: Do not rely on specific node counts
- ztp: ztp-done policy: Ensure MCP degradedMachineCount == 0
- ztp: ztp-done policy: remove day-0 MachineConfig checks
- ztp: ztp-done policy: Check for the PAO MachineConfig in any MCP pool
- ztp: ztp-done policy: Remove any kube-apiserver checks
- ztp: ztp-done policy: Remove redundant PAO checks
- ztp: ztp-done policy: Remove redundant SRIOV checks
- ztp: ztp-done policy: Check PTP daemonset, not PTP deployment
- ztp: ztp-done policy: Check for the 50-nto-[worker|master] MachineConfig

### DIFF
--- a/ztp/source-crs/validatorCRs/informDuValidator.yaml
+++ b/ztp/source-crs/validatorCRs/informDuValidator.yaml
@@ -22,14 +22,7 @@ status:
   syncStatus: Succeeded
 ---
 apiVersion: apps/v1
-kind: Deployment
+kind: DaemonSet
 metadata:
-  name: ptp-operator
+  name: linuxptp-daemon
   namespace: openshift-ptp
-status:
-  conditions:
-    - type: Available
-      status: "True"
-      reason: MinimumReplicasAvailable
-  availableReplicas: 1
-  readyReplicas: 1
--- a/ztp/source-crs/validatorCRs/informDuValidator.yaml
+++ b/ztp/source-crs/validatorCRs/informDuValidator.yaml
@@ -9,12 +9,13 @@ spec:
     matchLabels:
       machineconfiguration.openshift.io/role: master
 status:
-  unavailableMachineCount: 0
   conditions:
     - type: Updated
       status: "True"
     - type: Updating
       status: "False"
+  degradedMachineCount: 0
+  unavailableMachineCount: 0
   configuration:
     source:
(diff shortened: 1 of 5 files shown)


ztp/source-crs/validatorCRs/informDuValidator.yaml
```diff
@@ -1,55 +1,23 @@
...
 kind: MachineConfigPool
 metadata:
-  labels:
-    pools.operator.machineconfiguration.openshift.io/master: ""
-  name: master
-spec:
-  machineConfigSelector:
-    matchLabels:
-      machineconfiguration.openshift.io/role: master
+  name: $mcp
```
Missxiaoguo, line 4: Smart to pass $mcp!

lack:
After consultation with the PAO team, the definitive answer is that we need to check both the PAO-created MachineConfig and the NTO-created 50-nto-[master|worker] policy, as they are not created atomically.  Un-holding this PR.

ztp/source-crs/validatorCRs/informDuValidator.yaml
```diff
@@ -59,63 +27,7 @@ status:
   syncStatus: Succeeded
 ---
 apiVersion: apps/v1
-kind: Deployment
+kind: DaemonSet
```
Missxiaoguo, line 30: Is it necessary to check status.numberMisscheduled=0 as well?

lack: We could add that maybe - Right now we're just checking that the daemonset has been created, not that it has been scheduled.

Originally I tried to check that the `numberUnavailable` was zero, but it seems like instead of setting that to 0 it just removes the field, and the policy never goes compliant.  I'm not sure numberMisscheduled is right to check - That counts pods that are running but on the wrong node...  It might more correct to check that the `numberReady`:and `numberActive` are equal to `desiredNumberScheduled` - This is a bit annoying to do with ACM though.

I think just checking if the daemonset is present is good enough for us to claim the DU profile is "done".

Missxiaoguo (approved):
/lgtm

PR #903, opened 2022-01-13
oot-driver: address lingering containers during reboot cycles

novacain1:
addresses known problems experienced with a partner, where reboot cycles leave podman containers resident.  the resident containers can go to 2048, and then it stops working for the partner.  this cleans them up and prevents an infrequent problem with iptables forwarding chains, when switching to the out-of-tree ice driver.


SchSeba:
Hi @novacain1 please add `oot-driver: ` in the commit title to make the CI happy :)

PR #905, opened 2022-01-13
features: Update PTP config to pass validations

imiller0:
The ptp-operator has a new validating webhook which ensures that the
parameters in the PTP config are valid. The parameters used in CI don't
conform so CI fails. In specific the summary_interval and
logSyncInterval need to be set to the same value. The CI is being
updated to use -4 to match the "right" setting when "fast event
notification" is being used.
Validator:
https://github.com/openshift/ptp-operator/blob/master/api/v1/ptpconfig_webhook.go#L122-L124

### DIFF
--- a/feature-configs/cn-ran-overlays/ran-profile/cluster-config/du-common/ptp/ptpconfig-grandmaster.yaml
+++ b/feature-configs/cn-ran-overlays/ran-profile/cluster-config/du-common/ptp/ptpconfig-grandmaster.yaml
@@ -8,7 +8,7 @@ spec:
   - name: "grandmaster"
 # The interface name is hardware-specific    
     interface: "eno1"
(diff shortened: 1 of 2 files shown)


imiller0:
I expect this initial commit to fail CI in the same way as [this run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-kni_cnf-features-deploy/826/pull-ci-openshift-kni-cnf-features-deploy-release-4.9-e2e-gcp-ran-profile/1481613902348292096#1:build-log.txt%3A2365). Once confirmed I'll update the PR to change the parameters and confirm it is fixed.

yuvalk:
can commit msg explain what validation are you looking to pass
and also why changing the code and not the validation is needed?
or better explain why these values are correct?

lack (approved):

PR #907, opened 2022-01-14
Bug 2040705: ztp: Only run PTP daemons on worker nodes for standard clusters

lack:


lack:
This works properly on standard, three-node, and SNO clusters.

PR #910, opened 2022-01-17
Bug 2037874: ztp: Retire DU tuned patch

pixelsoccupied:
The patch needs to be retired as it is a temporary workaround. 

After migrating the required confs from Tuned CR to `PerformanceProfile`, it can be safely retired. 

`cat /proc/cmdline` output below -->

`
BOOT_IMAGE=(hd0,gpt3)/ostree/rhcos-315b5118fb91a6624749b371b769f894a99120da503c907cd32ec6af55c3f072/vmlinuz-4.18.0-305.28.1.rt7.100.el8_4.x86_64 random.trust_cpu=on console=tty0 console=ttyS0,115200n8 ignition.platform.id=metal ostree=/ostree/boot.1/rhcos/315b5118fb91a6624749b371b769f894a99120da503c907cd32ec6af55c3f072/0 ip=enp1s0:dhcp root=UUID=76e5f01f-4b7b-4843-badd-fee918d0fccf rw rootflags=prjquota skew_tick=1 nohz=on rcu_nocbs=2-7 tuned.non_isolcpus=00000003 intel_pstate=disable nosoftlockup tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=managed_irq,2-7 systemd.cpu_affinity=0,1 default_hugepagesz=1G hugepagesz=1G hugepages=1 idle=poll rcupdate.rcu_normal_after_boot=0 nohz_full=2-7 kernel.timer_migration=1
`


ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
```diff
@@ -34,6 +34,11 @@ spec:
       metadata:
         name: openshift-node-performance-profile
       spec:
+        additionalKernelArgs:
+          - "idle=poll"
+          - "rcupdate.rcu_normal_after_boot=0"
+          - "kernel.timer_migration=1"
+          - "nohz_full=2-19,22-39" # same as .spec.cpu.isolated
```
lack, line 41: Since the source-crs version already has most of these, can we just set this last `nohz_full` line?  Or does setting part of a list in PGT reset the whole list @serngawy / @imiller0 ?

pixelsoccupied: Ya the whole list in the source-cr file would get replaced (it's not smart enough to append/prepend a list) if don't the include the whole thing.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
```diff
@@ -43,20 +48,3 @@ spec:
...
-            data: |
-              [main]
-              summary=Configuration changes profile inherited from performance created tuned
-              include=openshift-node-performance-openshift-node-performance-profile
-              [bootloader]
-              cmdline_crash=nohz_full=2-19,22-39
-              [sysctl]
-              kernel.timer_migration=1
-              [service]
-              service.stalld=start,enable
```
imiller0, line 62: How is the start/enable of stalld accounted for?

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
```diff
@@ -34,6 +34,11 @@ spec:
       metadata:
         name: openshift-node-performance-profile
       spec:
+        additionalKernelArgs:
+          - "idle=poll"
+          - "rcupdate.rcu_normal_after_boot=0"
+          - "kernel.timer_migration=1"
```
imiller0, line 40: Does this correctly set the sysctl value when the node is booted?

imiller0 (changes requested):
When retiring a profile we should (unless there are compelling reasons otherwise) leave the source-cr in place (adding a comment that it has been deprecated) but remove it from all the reference PolicyGenTemplates. This helps ensure backwards compatibility (existing deployments using that source-cr will continue to work) for at least 1 release.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
browsell, line 40: Unfortunately tuned will overwrite this when it starts up, so we will need to keep the TunedPerformancePatch for sysctl
[sysctl]
              kernel.timer_migration=1

The updated TunedPerformancePatch will not require a reboot when applied.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
pixelsoccupied: Since we are removing `nohz_full` from `tuned`, there's no need account for the cpu starvation via `stalld`. cc @browsell

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
pixelsoccupied: No actually! Brought back Tuned with this value under [sysctl]

pixelsoccupied:
Need to find a better way to avoid repeating 
```          
          - "idle=poll"
          - "rcupdate.rcu_normal_after_boot=0"
          - "kernel.timer_migration=1"
          - "nohz_full=2-19,22-39" # same as .spec.cpu.isolated
```

But the problem is explained below 

```
# source cr
spec:
  additionalKernelArgs:
  - "idle=poll"
  - "rcupdate.rcu_normal_after_boot=0"
  - "nohz_full=$isolated" 
  cpu:
    isolated: $isolated
    reserved: $reserved
```

this converts to
```
# out/policy           

 spec:
              additionalKernelArgs:
              - idle=poll
              - rcupdate.rcu_normal_after_boot=0
              - nohz_full=$isolated ----> should also replace
              cpu:
                isolated: 2-19,22-39 
                reserved: 0-1,20-21
```
i.e the replacement only seems seems to work for maps

For [list](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/policygenerator/policyGen/policyBuilderOverlay_test.go#L130) you can only [replace](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/policygenerator/policyGen/policyBuilderOverlay_test.go#L164) the whole list (cant append/prepend). With  maps…you can change the value cause the code is looking by matching key. This $ marker (e.g $val) is actually not used for [much](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/policygenerator/policyGen/policyBuilder.go#L248) but rather a place holder. So like “find the key isolated and replace value with 2-19,22-39”  instead of “find $isolated and replace with 2-19,22-39" 

@serngawy would like your opinion on this. I think behaviours is expected and aligned with say...helm. But is there anyway to replace one "val" is part of a list entry?

pixelsoccupied:
The feature to replace an item from a list can be a different PR (https://issues.redhat.com/browse/CNF-4664). Closing this PR as finally Tuned can not be retired (Check BZ for latest).

PR #909, opened 2022-01-14
Bug 2040791: ztp: Change example PGT to generate inform policies

lack:
This is required for integration with the Topology Aware Lifecycle
Operator.

Signed-off-by: Jim Ramsay <jramsay@redhat.com>

/cc @imiller0 @serngawy
/hold for testing

PR #911, opened 2022-01-17
Bug 2040791: ztp: Change the default PolicyGenerator to generate inform, not enforce policies

lack:
By changing the default in the tool, we ensure that users upgrading for
4.9 to 4.10 do not accidentally regenerate new enforce policies that
affect preexitsing clusters.  Likewise no action needs to be taken to
update existing common or group policies from older versions of the
tool (at least not to update the remediationAction part), minimizing the
number of manual actions a user needs to take when upgrading from 4.9 to
4.10.

/cc @imiller0 @serngawy
/hold until the approach is decided (between this and #909)

### DIFF
--- a/ztp/policygenerator/utils/utils.go
+++ b/ztp/policygenerator/utils/utils.go
@@ -42,8 +42,8 @@ type PolicyGenTempSpec struct {
 func (pgt *PolicyGenTempSpec) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	type PolicyGenTemplateSpec PolicyGenTempSpec
 	var defaults = PolicyGenTemplateSpec{
-		WrapInPolicy:      true, //Generate ACM wrapped policies by default
-		RemediationAction: "enforce",
+		WrapInPolicy:      true,     //Generate ACM wrapped policies by default
+		RemediationAction: "inform", // Generate inform policies by default
 		ComplianceType:    DefaultComplianceType,
 	}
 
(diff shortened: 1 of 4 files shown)


ztp/policygenerator/utils/utils.go
```diff
@@ -42,8 +42,8 @@ type PolicyGenTempSpec struct {
 func (pgt *PolicyGenTempSpec) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	type PolicyGenTemplateSpec PolicyGenTempSpec
 	var defaults = PolicyGenTemplateSpec{
-		WrapInPolicy:      true, //Generate ACM wrapped policies by default
-		RemediationAction: "enforce",
+		WrapInPolicy:      true,     //Generate ACM wrapped policies by default
+		RemediationAction: "inform", // Generate inform policies by default
```
lack, line 46: This is the core of the functional change

lack:
/unhold
I think we're agreed on this as the solution.

serngawy (commented):
As general comment, This is API changes. What was the conclusion on changing the generated ACM policies 4.9  from enforce to inform after upgrade to 4.10 ? How we will tell the enduser about it ?

lack:
> As general comment, This is API changes. What was the conclusion on changing the generated ACM policies 4.9 from enforce to inform after upgrade to 4.10 ? How we will tell the enduser about it ?

Good point - I assume it would be in the same place where we talk about the 4.9 to 4.10 upgrade.

imiller0:
> > As general comment, This is API changes. What was the conclusion on changing the generated ACM policies 4.9 from enforce to inform after upgrade to 4.10 ? How we will tell the enduser about it ?
> 
> Good point - I assume it would be in the same place where we talk about the 4.9 to 4.10 upgrade.

I think we should also consider adding documentation (if not already covered somewhere) for the user on how to run the tools stand-alone to generate installation and configuration CRs. The default behavior would be documented there as well.

imiller0 (approved):
/lgtm

PR #915, opened 2022-01-20
oot-driver: use environmet variables for driver version

sudomakeinstall2:
add auxiliary module necessary for recent iavf and ice driver


tools/oot-driver/hack/create-special-resource.sh
```diff
@@ -25,6 +28,8 @@ do
       then
         sed 's/name: \"oot-source-driver:latest\"/name: \"\$\{INTERNAL_REGISTRY\}\/oot-source-driver:latest\"\n          pullsecret: external-registry/g' -i ./special-resource.yaml
         sed 's/ImageStreamTag/DockerImage/g' -i ./special-resource.yaml
+        envsubst < "./special-resource.yaml" > temp
```
SchSeba, line 31: why we need to run this one again? it's already in line 26

sudomakeinstall2: there is a variable inside the `sed` command that I could not replace so I did a dirty fix and used `envsubst` again. However I fixed this using a better `sed` command

tools/oot-driver/charts/iavf-driver-0.0.1/Dockerfile-driver.SRO
```diff
@@ -55,6 +55,7 @@ ENV KERNEL_VERSION=$KVER
 RUN dnf install -y kmod
 
 COPY --from=builder /lib/modules/$KVER/updates/drivers/net/ethernet/intel/iavf/iavf.ko /oot-driver/
+COPY --from=builder /lib/modules/$KVER/updates/drivers/net/ethernet/intel/auxiliary/auxiliary.ko /oot-driver/
```
sudomakeinstall2, line 58: This might not work for old versions of the driver.

SchSeba: is this driver always built?

sudomakeinstall2: from ice version 1.7.7 and iavf version 4.3.12 the auxiliary module is built.

SchSeba:
Hi @sudomakeinstall2 can you please check the logs I think this change break the tests

sudomakeinstall2:
@SchSeba I changed the test, so it uses ice driver version 1.7.16. The version in the test is very old.

yanirq:
@SchSeba   the errors in e2e-gcp-ovn lane should be resolved by this patch alone ? or this is yet again a flakiness we have experienced so far ?

sudomakeinstall2:
@SchSeba I ran the oot-driver tests separately, and the tests passed. I am not sure why the oot-driver tests fail in the pipeline.

PR #917, opened 2022-01-20
Bug 2041639: ztp: Add DisableSnoNetworkDiag to group-du-sno PolicyGenTemplate

daosman:
This is a back port of the fix in master branch that is missing in release-4.9


imiller0 (approved):
/lgtm

PR #918, opened 2022-01-20
Bug 2038253: ztp: Add bindingExcludedRules in PGT

Missxiaoguo:
This commit includes the changes:
  - Add bindingExcludedRules in PGT crd to support selecting clusters
    which do not contain a label
  - Update PGT examples and move validator policy to three new PGTs
    corresponding to each type of cluster. The policy binds to the
    group types and to the clusters without "ztp-done" label.
  - Add a very large wave in validator source CR so validator policy
    will be added at the end of policies in generated UOCR by TALO

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-standard-validator-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-standard-validator-ranGen.yaml
@@ -1,0 +1,21 @@
+apiVersion: ran.openshift.io/v1
+kind: PolicyGenTemplate
+metadata:
+  # The name will be used to generate the placementBinding and placementRule names as {name}-placementBinding and {name}-placementRule
+  name: "group-du-standard-validator"
+  namespace: "ztp-group"
+spec:
+  # This policy will correspond to all clusters with label specified in bindingRules and
+  # without label specified in bindingExcludedRules.
+  bindingRules:
+    group-du-standard: ""
+  bindingExcludedRules:
+    # The ztp-done label is used in coordination with the Topology Aware Lifecycle Operator(TALO).
+    # Please do not change this label.
+    ztp-done: ""
+  mcp: "worker"
+  sourceFiles:
+    # Create inform policy to validate configuration CRs that will be applied to all standard clusters
+    - fileName: validatorCRs/informDuValidator.yaml
+      remediationAction: inform
+      policyName: "du-validator-policy"
--- a/ztp/policygenerator/policyGen/policyHelper.go
+++ b/ztp/policygenerator/policyGen/policyHelper.go
@@ -30,7 +30,40 @@ func CreatePolicySubject(policyName string) utils.Subject {
 	return subject
 }
 
-func CreatePlacementRule(name string, namespace string, matchKeyValue map[string]string) utils.PlacementRule {
+/* Func CheckBindindRules checks the following invalid rules:
+----------------------
+bindingRules:
+	labelKey: ""
+bindingExcludedRules:
+	labelKey: ""
+-----------------------
+bindingRules:
+	labelKey: "labelValue"
+bindingExcludedRules:
+	labelKey: ""
+-----------------------
+bindingRules:
+	labelKey: "labelValue"
+bindingExcludedRules:
+	labelKey: "labelValue"
+*/
+func CheckBindingRules(pgtName string,
+	bindingRules map[string]string, bindingExcludedRules map[string]string) error {
+
+	for key, valueExcludedRules := range bindingExcludedRules {
+		valueRules, found := bindingRules[key]
+		if found && (valueExcludedRules == "" || valueExcludedRules == valueRules) {
+			return fmt.Errorf("Invalid bindingRules and bindingExcludedRules found in PGT %s. "+
+				"Cannot add label (%s:\"%s\") in bindingRules but label (%s:\"%s\") in bindingExcludedRules.",
+				pgtName, key, valueRules, key, valueExcludedRules)
+		}
+	}
+	return nil
+}
+
+func CreatePlacementRule(name string, namespace string,
+	bindingRules map[string]string, bindingExcludedRules map[string]string) utils.PlacementRule {
+
 	placementRule := utils.PlacementRule{}
 	placementRule.ApiVersion = "apps.open-cluster-management.io/v1"
 	placementRule.Kind = "PlacementRule"
@@ -38,11 +71,27 @@ func CreatePlacementRule(name string, namespace string, matchKeyValue map[string
 	placementRule.Metadata.Namespace = namespace
 	expressions := make([]map[string]interface{}, 0)
 
-	for key, value := range matchKeyValue {
+	for key, value := range bindingRules {
 		expression := make(map[string]interface{})
 		expression["key"] = key
-		expression["operator"] = utils.InOper
-		expression["values"] = strings.Split(value, ",")
+		if value == "" {
+			expression["operator"] = utils.ExistOper
+		} else {
+			expression["operator"] = utils.InOper
+			expression["values"] = strings.Split(value, ",")
+		}
+		expressions = append(expressions, expression)
+	}
+
+	for key, value := range bindingExcludedRules {
+		expression := make(map[string]interface{})
+		expression["key"] = key
+		if value == "" {
+			expression["operator"] = utils.DoesNotExistOper
+		} else {
+			expression["operator"] = utils.NotInOper
+			expression["values"] = strings.Split(value, ",")
+		}
 		expressions = append(expressions, expression)
 	}
 
@@ -83,7 +132,7 @@ func BuildObjectTemplate(resource generatedCR) utils.ObjectTemplates {
 
 // We are using Deploywaves to order policies deployment.
 // Each resource needs to be applied via ACM enforce policy controlled
-// by ClusterGroupUpgrade operator should have a Deploywave annotation.
+// by Topology Aware Lifecycle operator should have a Deploywave annotation.
 // For example,
 //   metadata:
 //     annotations:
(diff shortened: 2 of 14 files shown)


ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
```diff
@@ -6,7 +6,8 @@ metadata:
   namespace: "ztp-common"
 spec:
   bindingRules:
-    common: "true"
+    includedClusters:
+      common: "true"
```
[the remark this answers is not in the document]
Missxiaoguo: It's possible to consider the labels added under bindingRules to be "includedClusters" labels and apply the same logic as "includedClusters" labels, but I feel it's more clear to have labels explicitly defined under either includedClusters or excludedClusters to indicate where the policies should be applied, so I prefer to have this added in the list of actions when upgrading to 4.10

Missxiaoguo:
@lack @serngawy @imiller0 @browsell

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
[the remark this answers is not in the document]
lack: For a backwards incompatible change like this one, should we bump the version number of the PGT spec, and consider writing a utility that helps with the upgrade?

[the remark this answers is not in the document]
serngawy: yes, back compatibility will be real issue and our code should work on both API version

serngawy (changes requested):
We must support the old bindingRules structure and our implementation must support the old/new bindingRules . AS Jim comment better to have this change under PGT v2 it is API changes.

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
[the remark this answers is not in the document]
imiller0: I would go a step further and say we need to do everything we can to stay backwards compatible. It causes significant disruption if the customer has to change what they are generating. Maybe we can consider:
```
bindingRules:
    common: "true"
bindingExcludeRules:
    notHere: ""
```

imiller0 (changes requested):

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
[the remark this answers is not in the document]
Missxiaoguo: > I would go a step further and say we need to do everything we can to stay backwards compatible. It causes significant disruption if the customer has to change what they are generating. Maybe we can consider:
> 
> ```
> bindingRules:
>     common: "true"
> bindingExcludeRules:
>     notHere: ""
> ```

I agree. I was thinking the same thing. To make it backward-compatiable, keep the existing bindingRules and have another field for opposite rules.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-and-3node-validator-ranGen.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: ran.openshift.io/v1
+kind: PolicyGenTemplate
+metadata:
+  # The name will be used to generate the placementBinding and placementRule names as {name}-placementBinding and {name}-placementRule
+  name: "group-du-sno-and-3node-validator"
```
lack, line 6: If we want a less cumbersome name, maybe "group-du-master-validator", and the standard cluster one could be "group-du-worker-validator" - That's the main difference, which mcp pool are we watching; master or worker...

lack: But aside from this minor net, this looks great!

Missxiaoguo: Yeah, the only difference is the mcp. I like the idea. To make clarity of which one is for sno&3node and which one is for standard, I would keep the yaml file names and only change the PGT name.

ztp/source-crs/validatorCRs/informDuValidator.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfigPool
 metadata:
   name: $mcp
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "1000000000000"
```
imiller0, line 6: Maybe something not quite so huge? 10000?

Missxiaoguo: Sure, I know it's a little bit crazy :)

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-and-3node-validator-ranGen.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: ran.openshift.io/v1
+kind: PolicyGenTemplate
+metadata:
+  # The name will be used to generate the placementBinding and placementRule names as {name}-placementBinding and {name}-placementRule
+  name: "group-du-master-validator"
+  namespace: "ztp-group"
+spec:
+  bindingExcludedRules:
+    # This policy will correspond to all clusters without this label:
```
serngawy, line 10: This label "ztp-done" will be added by TALCO ? if yes, please add this info to the description and tell end-user must not  change that otherwise TALCO will miss behave.

ztp/policygenerator/policyGen/policyHelper.go
```diff
@@ -30,19 +30,49 @@ func CreatePolicySubject(policyName string) utils.Subject {
 	return subject
 }
 
-func CreatePlacementRule(name string, namespace string, matchKeyValue map[string]string) utils.PlacementRule {
+func CheckBindingRules(pgtName string,
+	bindingRules map[string]string, bindingExcludedRules map[string]string) error {
+
+	for key := range bindingRules {
```
serngawy, line 36: this logic is strange , I can have binding rule like;
- key: "du-sno"
  operator: "Exists"

And exclude binding rule like;
- key: "du-sno"
  operator: NotIn
  value: [east] 

Just check the same set not both.

serngawy (changes requested):
Few comments and please make sure the generated policy with excluded binding rule can get applied on ACM hub without causing error. Thanks

ztp/policygenerator/policyGen/policyHelper.go
Missxiaoguo: Thanks for pointering this out. Right, this is the only valid case for same key in bindingRules & bindingExcludedRules. I have updated the PR to address this.

ztp/policygenerator/policyGen/policyHelper.go
```diff
@@ -30,19 +30,68 @@ func CreatePolicySubject(policyName string) utils.Subject {
...
+----------------------
+bindingRules:
+	labelKey: ""
+bindingExcludedRules:
+	labelKey: ""
+-----------------------
+bindingRules:
+	labelKey: "labelValue"
+bindingExcludedRules:
+	labelKey: ""
```
imiller0, line 43: Is the reverse valid?
binding:
  labelKey: ""
exclude
  labelKey: "abcd"

Missxiaoguo: Yes, this is valid. It checks for the clusters has label labelKey but its value not in abcd.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-and-3node-validator-ranGen.yaml
```diff
@@ -0,0 +1,18 @@
+---
```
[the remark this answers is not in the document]
Missxiaoguo: I saw other examples have the separator at the top so I added it to make consistent, but yes, it's not needed. I have removed.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-standard-validator-ranGen.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: ran.openshift.io/v1
+kind: PolicyGenTemplate
+metadata:
+  # The name will be used to generate the placementBinding and placementRule names as {name}-placementBinding and {name}-placementRule
+  name: "group-du-worker-validator"
+  namespace: "ztp-group"
+spec:
+  bindingExcludedRules:
```
imiller0, line 8: Without binding rules to match some type of common or group label this policy will apply to anything without ztp-done. Should it also include a binding rule for ztp-running so that it doesn't apply to (and fail against) the "local-cluster" and/or any other clusters the user added to the hub outside the scope of ZTP.

Missxiaoguo: If adding bindingRule for ztp-running, validator policy will be applied to cluster with ztp-running label, but this label is added when start remediating auto-generated UOCR, so at the time of creating UOCR, validator policy will be excluded in the list of managedPolicies so we cannot add the ztp-running bindingRules.
But yes, with only bindingExcludedRules, it will apply to the local-cluster and other clusters added outside of ZTP, so we want another bindingRules to exclude them.

It occurs to me that we actually need have 3 validator PGTs for each type of cluster, i.e.
    ztp-group-sno bindingRules for sno
    ztp-group-3node bindingRules for 3node
    ztp-group-standard bindingRules for standard
, then corresponding validator policy will be applied to the right type of cluster, otherwise, for example, group-du-worker-validator will be applied to the sno as well. It's a bit duplicate for sno and 3node validator policy as the only difference is different label...

imiller0: I think I understand and agree that we do need one per cluster type. With the use case where the user deploys both SNO and standard clusters, the hub will have both the SNO and standard DU validator created and right now they will both bind to newly deploying SNO and standard clusters. That could mean that the cluster never finished ZTP because only one of the two DU validators will ever be compliant. By adding a bindingRule for the matching group type (eg group-du-sno and group-du-standard we avoid this.

Missxiaoguo:
retest-required

imiller0 (approved):
/lgtm

imiller0 (approved):
/lgtm

PR #920, opened 2022-01-21
ztp: Remove PtpOperatorConfig from SNO and 3node example PGTs

lack:
This isn't needed for those platforms since all nodes are masters and
all nodes should want the PTP daemon running.  It's only critical for
standard clusters where we want to restrict PTP to workers only.  So
moving this simplifies things for users especially when upgrading from
4.9 to 4.10.

/cc @imiller0


imiller0 (approved):
/lgtm

PR #924, opened 2022-01-25
ztp: ovs-vswitchd running more threads than expected

daosman:
The ovs-vswitchd process on DU-configured SNO runs as part of the
management overhead set of processes which are constrained to the
management cores. When ovs-vswitchd starts up, it samples the system for
number of cores and the cpu-affinity mask to determine how many worker
threads should be created.  That sampling can happen before we constrain
ovs to the set of management cores, which results in OVS still using a
large number of worker threads even after we constrain it to the smaller
set of management cores.

Re-trigger OVS to recalculate the worker threads using the ovs-vsctl
command.  We use the same calculation done by OVS to re-configure it so
that the number of worker threads matches the reduced set of cores.


ztp/extra-manifests-builder/04-accelerated-container-startup/test.sh
```diff
@@ -138,4 +138,22 @@ test_unrestrictedCpuset 0 "0-9" ' '                          "0-9" "Empty file f
...
+  echo "Testing getCPUCount: $4"
+
+  local result
+  result=$(getCPUCount $cpuSet)
+  local rc=$?
+  [[ $rc -eq $expectedRc ]] || fatal "  getCPUCount failed: Expected rc $expectedRC != $rc"
+  [[ $result == $expectedValue ]] || fatal "  getCPUCount failed: Expected return value '$expectedValue' != '$result'"
+}
+
+test_getCPUCount 0 "4" "3-4,54,55"     "Valid CPU List with no ranges"
```
lack, line 152: But this has a range, `3-4`?

ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
```diff
@@ -80,6 +145,7 @@ resetAffinity() {
     logger "Recovery: Failed to re-affine $failcount processes"
     return 1
   fi
+  resetOVSthreads "$(getCPUCount ${cpuset})"
```
lack, line 148: I think we probably want to run this code even if `$failcount` implies some taskset operations above failed.  and if 'resetOVSthreads' fails, we should probably increment `((failcount++))`

lack (changes requested):
This is looking awesome!  Just a couple minor changes to get it to 100%

Also, to pass CI you'll need to run `make` in the `extra-manifests-builder` directory and check in the changes in `source-crs/extra-manifests` too.

ztp/extra-manifests-builder/04-accelerated-container-startup/test.sh
daosman: Will fix, should be a ',' instead of '-'

ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
daosman: I agree. done

ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
```diff
@@ -57,6 +57,75 @@ restrictedCpuset() {
...
+  return 0
+}
+
+resetOVSthreads () {
+  local cpucount="$1"
+  local curRevalidators=0
+  local curHandlers=0
+  local desiredRevalidators=0
+  local desiredHandlers=0
+  local rc=1
```
lack, line 97: I think if I read this correctly, this means that if this function is a no-op (ie, no work was done because `[[ $curRevalidators -ne $desiredRevalidators || $curHandlers -ne $desiredHandlers ]]` evaluated to false), then we return 1 -> I think a no-op should return 0 so we don't count it as a failure?

lack (changes requested):

ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
daosman: yes, good catch.

lack (approved):

PR #926, opened 2022-01-25
ztp: update default subscription CRs

Missxiaoguo:
Update each subscription CR with default "installPlanApproval"
set to Manual. With this configuration, installPlan will not be
installed automatically. TALO needs to find the right installPlan
and approve it.

Also add check of "AtLatestKnown" status in subscriptions, so
a new version of z-stream operator pushed to the registry will
cause the sub policy to be non-compliant then TALO is able to
know when to start upgrade and trigger upgrade.

Jira: https://issues.redhat.com/browse/CNF-4073


Missxiaoguo:
/hold
InstallPlan approval functionality has not been pushed to TALO. Hold this change for now.

PR #929, opened 2022-01-26
ztp: Move KlusterletAddonConfig out of SiteConfig and into PGT

lack:
This will allow customization of the KlusterletAddonConfig for customers
who need to do so.

/hold for testing


serngawy (commented):
I don't believe this is the right way to do it Jim. It mix the logic of siteconfig and PGT. It will lead to miss match errors from end-user and most likely confusion.

lack:
Closing in favor of #937 

This approach has some unfortunate side-effects on the way ArgoCD collides with the actions ACM takes on the policies.  This warrants more thought.

PR #931, opened 2022-01-27
deps: bump to PAO 4.10.0-rc.0

fromanirh:
Update dependencies to consume PAO e2e test changes for 4.10.0
(see:
https://github.com/openshift-kni/performance-addon-operators/pull/830)

unfortunately, this also bring (with little chance to split this huge
change):
- golang 1.17
- kubernetes 1.23.3 (plus deps)

on the bright side we have a explicit tag for PAO.
We used this weird scheme (v0.0.MAJORMINORPATCHSEQ) to be somehow
compliant with the semver golang expects without introducing even more
changes. Using the simpler `v4.10.0-rc.0` scheme would lead to
something very similar to
https://gitlab.com/libvirt/libvirt-go/-/issues/7
which we should discuss and possibly change on a separate PR.


fromanirh:
depends on: https://github.com/openshift-kni/cnf-features-deploy/pull/932

fromanirh:
FYI @cynepco3hahue

fromanirh:
requires: https://github.com/openshift/release/pull/25707

lack:
I've rebased this now that #932 is merged, and re-posted the equivalent change as #954

PR #938, opened 2022-02-02
ztp: Update the SiteConfig template language to be more go-template compatible

lack:
While this does not cut over to go templates yet, it does refactor the
templating language we use to be more future-proof and forward-facing,
given that customers can now override these templates.

/cc @imiller0 @serngawy

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/siteConfig.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfig.go
@@ -27,16 +27,23 @@ func (sc *SiteConfig) GetSiteConfigFieldValue(path string, clusterId int, nodeId
 	for _, key := range keys[1:] {
 		if v.Kind() == reflect.Slice || v.Kind() == reflect.Array {
 			intf := v.Interface()
-			arrClusters, ok := intf.([]Clusters)
 
+			arrClusters, ok := intf.([]Clusters)
 			if ok {
+				if clusterId < 0 || clusterId >= len(arrClusters) {
+					return nil, fmt.Errorf("Cluster ID out of range: %d", clusterId)
+				}
 				v = reflect.ValueOf(arrClusters[clusterId])
 			}
-			arrNodes, ok := intf.([]Nodes)
 
+			arrNodes, ok := intf.([]Nodes)
 			if ok {
+				if nodeId < 0 || nodeId >= len(arrNodes) {
+					return nil, fmt.Errorf("Node ID out of range: %d", nodeId)
+				}
 				v = reflect.ValueOf(arrNodes[nodeId])
 			}
+
 			v = v.FieldByName(key)
 		} else if v.Kind() == reflect.Struct {
 			v = v.FieldByName(key)
--- a/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
@@ -242,8 +242,14 @@ func (scbuilder *SiteConfigBuilder) getClusterCR(clusterId int, siteConfigTemp S
 			}
 			mapIntf[k] = value
 		} else if reflect.ValueOf(v).Kind() == reflect.String &&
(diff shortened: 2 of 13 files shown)


ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -242,8 +242,14 @@ func (scbuilder *SiteConfigBuilder) getClusterCR(clusterId int, siteConfigTemp S
 			}
 			mapIntf[k] = value
 		} else if reflect.ValueOf(v).Kind() == reflect.String &&
-			strings.HasPrefix(v.(string), "siteconfig.") {
-			valueIntf, err := siteConfigTemp.GetSiteConfigFieldValue(v.(string), clusterId, nodeId)
+			strings.HasPrefix(v.(string), "{{") &&
+			strings.HasSuffix(v.(string), "}}") {
+			// We can be cleaner about this, but this translation is minimally invasive for 4.10:
+			key, err := translateTemplateKey(v.(string))
```
lack, line 248: Here's the only change to the parsing logic!  That and the fairly simple (and well-tested) `translateTemplateKey` function below.

lack:
Testing shows that the siteconfig-generator output is identical before and after this change.

serngawy (changes requested):
not sure what the benefits of changing the path but may be better to use go template and evaluate the use of it.

lack:
> not sure what the benefits of changing the path but may be better to use go template and evaluate the use of it.

The main benefit is that if/because we want to move to GoTemplate in the future, we can without affecting an user that's overriding the template.  We only get once chance to release the template format for the first time.

The reason I didn't go all the way to GoTemplate in this particular change was to minimize the change against 4.10.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -27,16 +27,23 @@ func (sc *SiteConfig) GetSiteConfigFieldValue(path string, clusterId int, nodeId
 	for _, key := range keys[1:] {
 		if v.Kind() == reflect.Slice || v.Kind() == reflect.Array {
 			intf := v.Interface()
-			arrClusters, ok := intf.([]Clusters)
 
+			arrClusters, ok := intf.([]Clusters)
 			if ok {
+				if clusterId < 0 || clusterId >= len(arrClusters) {
+					return nil, fmt.Errorf("Cluster ID out of range: %d", clusterId)
+				}
```
lack, line 35: This (and the following) replaces a potential 'panic' with proper error handling.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -291,8 +294,105 @@ func Test_getExtraManifestTemplatedRoles(t *testing.T) {
...
+		input:          `top: { a: "{{ .Site.BaseDomain }}", middle: { b: "{{ .Cluster.ClusterName }}", bottom: "end" } }`,
+		expectedResult: `top: { a: "example.com", middle: { b: "cluster1", bottom: "end" } }`,
+	}, {
+		what:           "Nested structure appending another nested structure",
+		input:          `top: { a: "{{ .Cluster.ClusterLabels }}", middle: { bottom: "end" } }`,
+		expectedResult: `top: { a: {common: "true", "group-du-sno": "", "sites": "test-site"}, middle: { bottom: "end" } }`,
+	}, {
+		// TODO: This should be an error, not a success where the key is removed
+		what:           "Nested recursive structure with invalid key",
+		input:          `top: { middle: { bottom: "{{ .Cluster.NoSuchKey }}" } }`,
```
imiller0, line 390: Is the substitution relatively immune to whitespace as well? Might be good to add a few test cases like
`
key: "  {{ .Cluster.Name}}"
key: "{{ .Cluster.Name}}    "
key: "     {{ .Cluster.Name}}    "
`
etc

PR #939, opened 2022-02-02
ztp: Add a local registry to the SNO DU profile (CNF-3759)

pixelsoccupied:
A SNO DU can have limited BW between the centralized registry and itself. This can lead to long image download times.  This is unavoidable for the initial deployment. For subsequent pulls of an image we are relying on the fact that the image would be in the cache. There is the possibility that it could have been evicted but this would be the exception vs the rule.  

Unfortunately this strategy is flawed as crio will wipe /var/lib/containers/storage in the case on an unclean shutdown 

To address this, a local registry will be required on the SNO. The cluster images would need to be  mirrored to the this registry along with being used as the regisrty for the applications.


pixelsoccupied:
Let me if there's a better place to put these file in.

ztp/image-registry-crs/README.md
```diff
@@ -0,0 +1,43 @@
+Installation
+-
+1. Include `image-registry-partition-mc.yaml` in the extra-sno-manifest folder during SNO installation as specified [here](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/README.md)
+
+   - Make sure to modify values in the mc appropriately as it is dependent on the underlying disk. Refer to the comments on how to pick the right values. 
+2. Once cluster is up, apply `image-registry-pv-pvc.yaml` to create the pv and pvc.
```
lack, line 6: Maybe we should make these available in source-crs so they can use PGT to install these as part of normal day-2 operation?

pixelsoccupied: moved and updated README

browsell:
We want these CR's to be generated automatically by siteconfig and applied as part of the installation. The siteconfig should take the device and partition size as parameters and generate the required CRs

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -413,3 +451,92 @@ func (scbuilder *SiteConfigBuilder) splitYamls(yamls []byte) ([][]byte, error) {
...
+							SizeMiB:  &diskSizeMiB,
+							StartMiB: &diskStartMiB,
+						},
+					},
+				}
+				disks = append(disks, disk)
+
+				fileSystemDevice := "/dev/disk/by-partlabel/" + partitionLable
+				fileSystemLable := "/" + strings.ReplaceAll(partitionLable, "-", "/")
+				fileSystemFormat := "xfs"
```
lack, line 487: Should the filesystem be configurable (but default to xfs)?

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -236,6 +247,7 @@ type Nodes struct {
 	IgnitionConfigOverride string                 `yaml:"ignitionConfigOverride"`
 	Role                   string                 `yaml:"role"`
 	CrTemplates            map[string]string      `yaml:"crTemplates"`
+	DiskPartition		   []DiskPartition        `yaml:"diskPartition"`
```
lack, line 250: It seems like this would be forcing users to define disk partitions uniquely on every node; is that right?  For 3-node and standard clusters, it seems like they'd probably want to define a disk partition and have it appear on every worker...

pixelsoccupied:
Going to remove the code and rely on templating instead

ztp/siteconfig-generator/siteConfig/siteConfig.go
pixelsoccupied: One thing though...I'm assuming this is valid only for single node. But regardless of however many nodes...to create the partition on a node, the user must know its underlying hardware metadata including device path and available space which maybe different for every node. 

Also, in this case, once a partition is made only a single PV and PVC (can't use local storage operator with partition) is made which is the link between the image registry and disk..i.e even if you have multiple partitions user can't really leverage the other partitions on other nodes. 

I guess the bottom line is...users can make as many partitions as they like on however many nodes...only one is ever being is used for image registry. So the question is do we limit the ability to create partitions?

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -228,6 +230,43 @@ type TangConfig struct {
 	Thumbprint string `yaml:"thumbprint" json:"thp"`
 }
 
+type DiskPartition struct {
+	Device     string       `yaml:"device"`
+	Partitions []Partitions `yaml:"partitions"`
+}
+
+type Partitions struct {
+	Label         string `yaml:"label"`
```
lack, line 239: I was thinking about the user input here... Is it better to have them specify the label, and then we guess what the intended mount-point is from that, or would it be rather to have them specify the mount-point and we construct the proper label?

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -228,6 +230,43 @@ type TangConfig struct {
...
+}
+
+func (prt *Partitions) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type PartitionsDefault Partitions
+	var defaults = PartitionsDefault{}
+	err := unmarshal(&defaults)
+
+	// https://docs.openshift.com/container-platform/4.9/installing/installing_bare_metal/installing-bare-metal.html#installation-user-infra-machines-advanced_vardisk_installing-bare-metal
+	if !(strings.HasPrefix(defaults.Label, "var")) {
+		defaults.Label = "var-" + defaults.Label
```
lack, line 253: We should look at what Verizon is doing and see if this meets their needs.  If they're just creating partitions but not mounting them, do we need to enforce a leading "var-"?

ztp/siteconfig-generator/siteConfig/siteConfig.go
pixelsoccupied: I think mount-point would be better instead of Label as the user input. This mount-point path is also the common info between PV and MachineConfig.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
pixelsoccupied: XFS is the default of RHEL 7 and over. But I forget...did we decide to make this configurable?

ztp/siteconfig-generator/siteConfig/siteConfig.go
pixelsoccupied: As discussed, this feature is only for SNO and ok for it stay under `nodes` list.

ztp/siteconfig-generator/siteConfig/siteConfig.go
pixelsoccupied: Changed the input to `mount_point` and generating the `label` instead

ztp/siteconfig-generator/siteConfig/siteConfig.go
lack: It's possible they may be making partitions that they're not mounting; depends a lot on the applications that I don't know anything about.

pixelsoccupied: Gotcha. Removed it.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -228,6 +231,57 @@ type TangConfig struct {
...
+
+type Partitions struct {
+	MountPoint    string `yaml:"mount_point" `
+	Size          int    `yaml:"size"`
+	Start         int    `yaml:"start"`
+	MountFileName string `yaml:"-"`
+	Label         string `yaml:"-"`
+}
+
+type PartitionsUserInput Partitions
```
eggfoobar, line 247: Since this alias is really designed to be used in `UnmarshalYAML`, we should scope it in that function so it doesn't get used outside of its intended purpose.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -228,6 +231,57 @@ type TangConfig struct {
...
+
+func (prt *Partitions) UnmarshalYAML(unmarshal func(interface{}) error) error {
+
+	var userInput = PartitionsUserInput{}
+
+	err := unmarshal(&userInput)
+
+	if userInput.Start < 25000 {
+		// 25000 min value, otherwise root partition is too small and the installation will fail
+		err = fmt.Errorf("start value too small. must be over 25000")
```
eggfoobar, line 257: these `err` statements get over written if multiple user inputs are wrong, would be best to concat them, so we don't shadow errors presented to the user

ztp/siteconfig-generator/siteConfig/siteConfig_test.go
```diff
@@ -370,3 +371,135 @@ spec:
...
+		name           string
+		fields         fields
+		args           args
+		wantErr        bool
+		expectedResult *Partitions
+		expectedError  string
+	}{
+		{
+			name: "check if the mount file name is generated correctly",
+			args: args{unmarshal: func(i interface{}) error {
```
eggfoobar, line 397: It might be more robust to follow the code path that the yaml takes, so instead of mocking out the `unmarshal` function it would help catch more problems if we use yaml string, exactly what the user would define, and follow that code path, what do you think?

go.mod
```diff
@@ -42,6 +43,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.11.0
 )
 
+require github.com/google/go-cmp v0.5.6
```
eggfoobar, line 46: Small nitpick. This looks odd to me, was this manually added or through `go get github.com/google/go-cmp@v0.5.6` ?

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -228,6 +231,57 @@ type TangConfig struct {
...
+
+	// it's a required field
+	if userInput.MountPoint == "" {
+		err = fmt.Errorf("must provide a path for mount_point. e.g /var/path")
+	}
+	// run a clean to ensure path is not malformed
+	userInput.MountPoint = path.Clean(userInput.MountPoint)
+	// ensure path is absolute
+	if !(path.IsAbs(userInput.MountPoint)) {
+		userInput.MountPoint = path.Join("/", userInput.MountPoint)
```
eggfoobar, line 272: If the Path isn't absolute and it's not clear what the intent is, is it safe to assume we're at the root or should we fail here since we don't know if the user made a mistake?

eggfoobar (commented):
Looking good, just had a few comments

go.mod
pixelsoccupied: came up after I ran `go mod tidy && go mod vendor && go mod verify` . Should I manually move it under one of the require?

eggfoobar: Nah, that's fine, was just curious if the go tool created it

ztp/siteconfig-generator/siteConfig/siteConfig_test.go
pixelsoccupied: Updated the tests like we discussed

ztp/siteconfig-generator/siteConfig/siteConfig.go
pixelsoccupied: Now all the errors are concatenated into an array and returned after a join.

ztp/siteconfig-generator/siteConfig/siteConfig.go
pixelsoccupied: Moved it into the method and updated the tests cases

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -2,8 +2,11 @@ package siteConfig
 
 import (
 	"fmt"
+	"path"
 	"reflect"
 	"strings"
+
+	"github.com/coreos/go-systemd/v22/unit"
```
lack, line 9: Is v22 the right version to use?  Does that correspond to the systemd version that's running under RHCOS 8?

Can you use whatever "latest" version was already in use?

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.17
 
 require (
 	github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf // indirect
+	github.com/coreos/go-systemd/v22 v22.3.2
```
lack, line 7: Is it weird that we have a different go-systemd and go-systemd/v22?  How does that work?

And it also doesn't look like you checked in the vendored version of this or the go-cmp that's required below - I think after running 'go mod vendor' you also need to `git add -a vendor/` and check that in.

ztp/source-crs/ImageRegistryPvc.yaml
```diff
@@ -0,0 +1,15 @@
+kind: PersistentVolumeClaim
```
imiller0, line 1: We already have StoragePVC.yaml in the source-crs. Can the user simply use that source-cr (with name/namespace override) to avoid duplication?

ztp/source-crs/extra-manifest/image-registry-partition-mc.yaml.tmpl
```diff
@@ -0,0 +1,52 @@
...
+            Before=local-fs.target
+            [Mount]
+            Where={{ .MountPoint }}
+            What=/dev/disk/by-partlabel{{ .MountPoint }}
+            [Install]
+            WantedBy=local-fs.target
+      {{- end }}
+      {{- end }}
+{{ end -}}
+{{ end -}}
```
imiller0, line 52: Newline at the end please

ztp/source-crs/extra-manifest/image-registry-partition-mc.yaml.tmpl
```diff
@@ -0,0 +1,52 @@
+{{- range .Data.Nodes -}}
+{{ if .DiskPartition -}}
+{{- "# Automatically generated by extra-manifests-builder" }}
+{{ "# Do not make changes directly." }}
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  name: 98-var-imageregistry-partition-{{ $.Role }}
```
imiller0, line 8: If there are multiple nodes does each MachineConfig need a unique name per node?

ztp/image-registry-crs/README.md
```diff
@@ -0,0 +1,47 @@
...
+   - Make sure to modify values in the mc appropriately as it is dependent on the underlying disk.
+2. Use PGT, to apply `ImageRegistryPvPvc.yaml` and create the pv and pvc as part of normal day-2 operation
+   ```
+   sourceFiles:
+      - fileName: ImageRegistryPvPvc.yaml
+        policyName: "config-policy"
+   ```
+3. Enable image registry operator with the following command:
+   ```
+   oc patch configs.imageregistry.operator.openshift.io cluster --type merge --patch '{"spec":{"managementState": "Managed","storage":{"pvc":{"claim":"image-registry-pvc"}}}}'
```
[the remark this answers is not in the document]
pixelsoccupied: That's true. We actually already have [ImageRegistryConfig.yaml ](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/source-crs/ImageRegistryConfig.yaml). I'll remove this

ztp/siteconfig-generator/siteConfig/siteConfig.go
pixelsoccupied: Updating to use latest

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
pixelsoccupied: Defaulting to xfs but is now configurable

ztp/source-crs/ImageRegistryPvc.yaml
pixelsoccupied: Removed the file and using existing source cr for PVC

ztp/source-crs/extra-manifest/image-registry-partition-mc.yaml.tmpl
lack: Future-proof is always prudent +1

lack:
/override ci/prow/e2e-gcp-ovn

These changes do not have anything to do with SRO which is causing the failures.

lack:
Ah, we need approval from someone on the go.mod changes too before this can merge.

fromanirh:
/approve
the go.mod changes seems straightforward
trusting you folks for the content review

PR #942, opened 2022-02-03
Bug 2050393: ztp: Add support for customer features

imiller0:
A local registry is needed to cache container images and reduce the need
to pull images across a potentially slow link at system startup.

The Generic MachineConfig template is needed to support customer
specific tuning.

### DIFF
--- a/ztp/source-crs/ImageRegistryConfig.yaml
+++ b/ztp/source-crs/ImageRegistryConfig.yaml
@@ -1,0 +1,28 @@
+apiVersion: imageregistry.operator.openshift.io/v1
+kind: Config
+metadata:
+  name: cluster
+  annotations:
+    # The registry depends on backend storage being
+    # configured. StoragePVC defaults to wave 10, so this CR defaults
+    # to wave 11 to satisfy that dependency.
+    ran.openshift.io/ztp-deploy-wave: "11"
+spec:
+  logLevel: Normal
+  managementState: Managed
+  observedConfig: null
+  # These fields can be set by the user as desired
+  #operatorLogLevel: Normal
+  #proxy: {}
+  replicas: 1
+  requests:
+    read:
+      maxWaitInQueue: 0s
+    write:
+      maxWaitInQueue: 0s
+  rolloutStrategy: Recreate
+  # User needs to specify proper storage
+  storage: {}
+  #  pvc:
+  #    claim: ''
+  unsupportedConfigOverrides: null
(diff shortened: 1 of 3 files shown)


ztp/source-crs/ImageRegistryConfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+  requests:
+    read:
+      maxWaitInQueue: 0s
+    write:
+      maxWaitInQueue: 0s
+  rolloutStrategy: Recreate
+  # User needs to specify proper storage
+  storage: {}
+  #  pvc:
+  #    claim: ''
```
lack, line 27: Maybe use a template-like example for the parts we expect users to fill in?
```suggestion
  storage:
    pvc:
      claim: '$pvcname'
```

ztp/source-crs/MachineConfigGeneric.yaml
```diff
@@ -0,0 +1,13 @@
...
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: generic
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    #storage:
+    #  files:
+    #    - contents: ""
```
lack, line 13: This is interesting, given that we already have an 'extra-manifests' way of adding MachineConfigs at deployment.

Do we know what customers are adding via MachineConfig? This seems dangerous.

ztp/source-crs/StoragePVC.yaml
```diff
@@ -0,0 +1,15 @@
...
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "10"
+    volume.beta.kubernetes.io/storage-class: $storageclass
+# User should supply all required content
+spec: {}
+  #accessModes:
+  #  - ReadWriteOnce
+  #resources:
+  #  requests:
+  #    storage: 400Gi
```
lack, line 15: I think it's better to have a fully-formed example, if that doesn't break anything:
```suggestion
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 400Gi
```

imiller0: The issue comes in if they don't want to include those sub-sections. We don't have a way of removing them.

ztp/source-crs/MachineConfigGeneric.yaml
imiller0: MachineConfigs definitely have the ability to significantly affect the system. Any use would be at their own risk.

PR #950, opened 2022-02-08
wrong namespace

jgato:
The operator was not been installed in the created NS for that.

### DIFF
--- a/ztp/gitops-subscriptions/argocd/deployment/openshift-gitops-operator.yaml
+++ b/ztp/gitops-subscriptions/argocd/deployment/openshift-gitops-operator.yaml
@@ -7,13 +7,13 @@ apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
 metadata:
   name: openshift-gitops-operator
-  namespace: openshift-operators
+  namespace: openshift-gitops
 ---
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
   name: openshift-gitops-operator
-  namespace: openshift-operators
+  namespace: openshift-gitops
 spec:
   name: openshift-gitops-operator
   source: redhat-operators


ztp/gitops-subscriptions/argocd/deployment/openshift-gitops-operator.yaml
```diff
@@ -7,16 +7,16 @@ apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
 metadata:
   name: openshift-gitops-operator
-  namespace: openshift-operators
+  namespace: openshift-gitops
```
imiller0, line 10: I believe that when the operator is installed via the console the subscription is put into the openshift-operators namespace. Is there a reason to move to openshift-gitops?

jgato: I was following that tutorial: 
https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd
I think sometimes they are using openshift-operators as NS and others openshift-gitops. Following the tutorial, it failed to deploy correctly the operator because of that. So I thought thay maybe, it was better to create the subscription and the OperatorGroup on openshift-gitops. Also is the NS I have seen in other tutorials. 
I guess it is not important the name of the NS, but it should match later in the tutorial.

jgato: here an example of how the documentation points to different NS, is this correct? 
image

serngawy: Hi,
This is not different NS, the operator->subscription get installed under the openshift-operators NS as many other operators but the openshift-gitops operator uses the NS openshift-gitops to deploy all the deployments, pod, service ...etc  that is required.

jgato: Thanks @serngawy, maybe I did something else wrong while installing. The ArgoCD was not been deployed correctly I thought this was the problem.

PR #967, opened 2022-02-15
ztp: ClusterVersion changes for platform upgrade 

nishant-parekh:
ztp: add cluster version status to source-crs

### DIFF
--- a/ztp/source-crs/ClusterVersionUpgradeGraph.yaml
+++ b/ztp/source-crs/ClusterVersionUpgradeGraph.yaml
@@ -3,14 +3,7 @@ kind: ClusterVersion
 metadata:
   name: version
   annotations:
-    ran.openshift.io/ztp-deploy-wave: "100"
+    ran.openshift.io/ztp-deploy-wave: "90"
 spec:
   channel: stable-4.10
   upstream: https://api.openshift.com/api/upgrades_info/v1/graph # default upstream.
-  desiredUpdate:
-    force: false
-    version: $version
-status:
-  history:
-    -  version: $version
-       state: "Completed"


ztp/source-crs/ClusterVersionStatus.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterVersion
+metadata:
+  name: version
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "100"
+status:
+  conditions:
+  - type: Available
+    status: "True"
+
```
yliu127, line 11: 1. This condition will be True even if upgrade did not start at all.
2. How are we verifying the actual version?

yliu127 (changes requested):

ztp/source-crs/ClusterVersionStatus.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterVersion
```
imiller0, line 2: We already have the ClusterVersion CR in source-crs, why do we need a new copy?

Missxiaoguo: I would suggest to keep two ClusterVersion CRs.
1. ClusterVersionUpdateGraph.yaml only contains channel & upstream. This can be used to workaround the CVO issue https://bugzilla.redhat.com/show_bug.cgi?id=2055314 if needed.
2. ClusterVersion.yaml contains the channel, upstream with default values and desiredUpdate.version and status check.

We can easily remove the ClusterVersionUpdateGraph.yaml if the CVO fix merged to 4.9

ztp/source-crs/ClusterVersionStatus.yaml
irinamihai: @yliu127, talked to Nishant and made a "quick" test on my dev env and it seems that we can just use the history status and skip this condition check. For example:
```
status:
    history:
    - state: Completed
      version: 4.9.8
```

yliu127: Yes agreed this could be the solution for 4.10. For 4.11, maybe we can ask if CVO can add a boolean status field as Ian suggested. e.g., 
status:
  atDesiredVersion: true

ztp/source-crs/ClusterVersionStatus.yaml
nishant-parekh: I have updated the PR to split the upstream & channel change from the version change separate CRs ordered by the argoCD wave number.

ztp/source-crs/ClusterVersionUpdateGraph.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: config.openshift.io/v1
```
irinamihai, line 1: Hmm, why do we need 2 files for the same CR?

ztp/source-crs/ClusterVersion.yaml
```diff
@@ -5,9 +5,10 @@ metadata:
   annotations:
     ran.openshift.io/ztp-deploy-wave: "100"
 spec:
-  # The default channel is stable-<version>
-  channel: $channel
-  # The default upstream path is https://api.openshift.com/api/upgrades_info/v1/graph
-  upstream: $upstream
-  # desiredUpdate:
-  #   version: $version
+  desiredUpdate:
```
irinamihai, line 8: Should we plan to add more checks here, like looking at the cluster operators' health? Thus ensuring the cluster is in a healthy state after upgrade.

ztp/source-crs/ClusterVersion.yaml
```diff
@@ -5,9 +5,10 @@ metadata:
...
+    force: false
+    version: $version
+status:
```
Missxiaoguo, line 11: I am afraid of precaching feature needs to look at the "upstream" from this policy. If that's the case, "upstream:https://api.openshift.com/api/upgrades_info/v1/graph" should be added in default source CR.
Probably @vitus133 can confirm

ztp/source-crs/ClusterVersionUpdateGraph.yaml
Missxiaoguo: This provides us the ability to workaround the CVO issue. This CR can be used to update channel/upstream first, and the normal one above to be used for upgrade. After the fix of the CVO bz is back-ported to 4.9, we can safely delete this CR without impact.

nishant-parekh: we are trying to separate updating the channel and version. That is why we are creating 2 different CRs.

ztp/source-crs/ClusterVersion.yaml
nishant-parekh: Yes we want to add more checks here. This change should address the issue we are having right now.. where the policies are compliant without the upgrade completing..

ztp/source-crs/ClusterVersionUpdateGraph.yaml
imiller0: If this were the only CR, could the user accomplish the same thing through overlay with the PGT? They already need to supply the status section above in order to fill in the version.

ztp/source-crs/ClusterVersion.yaml
vitus133: Yes, this is a very valid concern. if you specify `version` you also need `channel` and `upstream`. All three are needed to get the release image pull spec.

ztp/source-crs/ClusterVersionUpdateGraph.yaml
yliu127: Agreed with one CR, and this CR should have channel and upstream explicitly specified. This will help with image precaching of ocp image, which expects channel, upstream and desiredUpdate to be found in one policy.

ztp/source-crs/ClusterVersionUpdateGraph.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterVersion
+metadata:
+  name: version
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "90"
+spec:
+  channel: $channel
```
yliu127, line 8: If we are going to use one source-cr, then we should put an explicit value here (stable-4.10?) to help with image precaching. If we do so, then part of this bz would be resolved: https://bugzilla.redhat.com/show_bug.cgi?id=2058414#c2

yliu127:
I just tried similar steps using existing cv source-cr in repo, which doesn't have a status field. Although "status.history" is included in PGT, but generated policy does not include status field.. Do we know if it will make a difference if status.history is included in source-cr?

nishant-parekh:
> I just tried similar steps using existing cv source-cr in repo, which doesn't have a status field. Although "status.history" is included in PGT, but generated policy does not include status field.. Do we know if it will make a difference if status.history is included in source-cr?

I have tested creating the policy using this and see the status in the generated policy and having status.history in the policy only updates once the upgrade is completed.

yliu127:
> I have tested creating the policy using this and see the status in the generated policy and having status.history in the policy only updates once the upgrade is completed.

I checked with Angie, and "status" has to be included in source-cr in order for it to work, meaning we should have two source-crs like you have it now.

yliu127:
LGTM

PR #969, opened 2022-02-15
ztp: add PGT example for platform upgrade

Missxiaoguo:


Missxiaoguo:
/hold
Not needed if https://github.com/openshift/cluster-version-operator/pull/742 is merged

PR #975, opened 2022-02-22
Update dpdk testpmd dockerfile with centos8 stream and latest dpdk version

alosadagrande:
CentOS 8 is EOL so needs to be replaced by CentOS 8 Stream
DPDK version is upgraded to latest 21.11 (which is not using the stable naming in the file)

cc/ @SchSeba

### DIFF
--- a/tools/s2i-dpdk/Dockerfile
+++ b/tools/s2i-dpdk/Dockerfile
@@ -1,12 +1,12 @@
-FROM quay.io/centos/centos:8
+FROM quay.io/centos/centos:stream8
 
 LABEL maintainer="Sebastian Scheinkman <sebassch@gmail.com>"
 LABEL io.openshift.s2i.scripts-url="image:///usr/libexec/s2i"
 LABEL io.s2i.scripts-url="image:///usr/libexec/s2i"
 
 ENV BUILDER_VERSION 0.1
-ENV DPDK_VER 20.11.1
-ENV DPDK_DIR /usr/src/dpdk-stable-${DPDK_VER}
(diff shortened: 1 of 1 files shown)


tools/s2i-dpdk/Dockerfile
```diff
@@ -1,12 +1,12 @@
...
+FROM quay.io/centos/centos:stream8
 
 LABEL maintainer="Sebastian Scheinkman <sebassch@gmail.com>"
 LABEL io.openshift.s2i.scripts-url="image:///usr/libexec/s2i"
 LABEL io.s2i.scripts-url="image:///usr/libexec/s2i"
 
 ENV BUILDER_VERSION 0.1
-ENV DPDK_VER 20.11.1
-ENV DPDK_DIR /usr/src/dpdk-stable-${DPDK_VER}
+ENV DPDK_VER 21.11
```
SchSeba, line 8: please can we use the same version as we ship in our product?

dpdk-20.11-3.el8.x86_64

tools/s2i-dpdk/Dockerfile
```diff
@@ -36,10 +36,10 @@ RUN yum install -y wget python3\
  yum install -y libibverbs-devel && \
  yum clean all
 
-RUN pip3 install meson ninja
+RUN pip3 install meson ninja pyelftools
```
SchSeba, line 39: why do we need this new package?

SchSeba:
closing in favor of https://github.com/openshift-kni/cnf-features-deploy/pull/1145

/close

PR #977, opened 2022-02-22
ztp: Update container builder to golang 1.17

lack:

### DIFF
--- a/ztp/resource-generator/Containerfile
+++ b/ztp/resource-generator/Containerfile
@@ -1,5 +1,5 @@
 # Builder
-FROM openshift/origin-release:golang-1.16 as builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.17 as builder
 USER root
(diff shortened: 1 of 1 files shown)


ztp/resource-generator/Containerfile
```diff
@@ -1,5 +1,5 @@
 # Builder
-FROM openshift/origin-release:golang-1.16 as builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.17 as builder
```
nishant-parekh, line 2: any particular reason we want to use the ci registry instead of a catalog released build?

lack: It's the same image that cnf-tests uses, no other reason.

lack: Also to use the released build I'd have to make more changes in the CI lane so it builds.

nishant-parekh: makes sense!

PR #985, opened 2022-02-25
Bug 2047871: ztp: Add mechanism to patch inner container references at build

lack:
The argocd deployment files refer to the upstream container images by
default. But downstream builds (or other special-purposw builds) need to
have a facility to override that internal reference. This is that
facility. Setting the IMAGE_REF build argument will patch any internal
references to that argument's value verbatim.

### DIFF
--- a/ztp/resource-generator/README.md
+++ b/ztp/resource-generator/README.md
@@ -23,3 +23,15 @@ out/
 ## Automatic upstream container builds
 The Red Hat Prow infractructure automatically pushes the head of this
 master branch to quay.io/openshift-kni/ztp-site-generator:latest
+
+## Custom builds
+The argocd deployment files refer to the upstream container images by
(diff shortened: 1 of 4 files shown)


ztp/resource-generator/README.md
```diff
@@ -23,3 +23,15 @@ out/
 ## Automatic upstream container builds
 The Red Hat Prow infractructure automatically pushes the head of this
 master branch to quay.io/openshift-kni/ztp-site-generator:latest
+
+## Custom builds
+The argocd deployment files refer to the upstream container images by
+default. But downstream builds (or other special-purposw builds) need
```
imiller0, line 29:
```suggestion
default. But downstream builds (or other special-purpose builds) need
```

imiller0 (commented):
/lgtm

PR #989, opened 2022-03-01
Bug 2059663: ztp: Suppress NMStateConfig when nodeNetwork empty

imiller0:
If the user does not supply any networking data in the nodeNetwork field
for a given node we should not generate the NMStateConfig CR. The
resultant NMStateConfig CR would be empty (no config/interfaces) and
causes the install to not complete (crash noted in infraenv). This patch
detects the empty nodeNetwork configuration in the user supplied
SiteConfig and suppresses the generation of NMStateConfig for that node.

### DIFF
--- a/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
+++ b/ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
@@ -99,12 +99,18 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp 
 			}
 			clusterCRs = append(clusterCRs, crValue)
 		} else if mapSourceCR["kind"] == "BareMetalHost" || mapSourceCR["kind"] == "NMStateConfig" {
-			for ndId := range siteConfigTemp.Spec.Clusters[clusterId].Nodes {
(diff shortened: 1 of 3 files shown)


lack:
/lgtm

Except that `go fmt` doesn't seem to like your formatting...

ztp/ztp-policy-generator/kustomize/plugin/policyGenerator/v1/policygenerator/siteConfig/siteConfigBuilder.go
```diff
@@ -99,12 +99,18 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
...
-			for ndId := range siteConfigTemp.Spec.Clusters[clusterId].Nodes {
+			for ndId, node := range siteConfigTemp.Spec.Clusters[clusterId].Nodes {
 				crValue, err := scbuilder.getClusterCR(clusterId, siteConfigTemp, mapSourceCR, ndId)
 				if err != nil {
 					return clusterCRs, err
 				}
-				clusterCRs = append(clusterCRs, crValue)
+				// BZ 2028510 -- Empty NMStateConfig causes issues and
+				// should simply be left out.
+				if mapSourceCR["kind"] == "NMStateConfig" && node.nodeNetworkIsEmpty() {
```
lack, line 109: Maybe move this before we call getClusterCR so we don't process something we don't intend to use?

PR #990, opened 2022-03-01
Bug 2034926: ztp: Readme updates for ztp-deploy-wave

Missxiaoguo:

### DIFF
--- a/ztp/source-crs/README.md
+++ b/ztp/source-crs/README.md
@@ -1,0 +1,16 @@
+## Source CRs
+This folder contains a full list of CRs for ZTP RAN solution.
+
+* ./ contains a list of configruation CRs to be deployed via the ACM policies
+* ./extra-manifest contains the MachineConfig CRs to be applied during the cluster installation
+* ./validatorCRs contains the validation CRs to be deployed via the ACM policies to validate some configuration
+
+### Waves
+Each configuration CR has a default `ran.openshift.io/ztp-deploy-wave` annotation that represents the deployment order of the resources wrapped in the ACM inform policies.
+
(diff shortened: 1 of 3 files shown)


ztp/source-crs/README.md
```diff
@@ -0,0 +1,42 @@
...
+This folder contains a full list of CRs for ZTP RAN solution.
+
+* / contains a list of configruation CRs to be deployed via the ACM policies
+* /extra-manifest contains the MachineConfig CRs to be applied during the cluster installation
+* /validatorCRs contains the validation CRs to be deployed via the ACM policies to validate some configuration
+
+### Waves
+Each configuration CR has a default `ran.openshift.io/ztp-deploy-wave` annotation that represents the deployment order of the resources wrapped in the ACM inform policies.
+
+The default waves in source CRs:
```
lack, line 11: I don't like keeping this list here in a README - As soon as we add or change source-crs, this list will be out of date.  Instead, can we come up with an easy command users can run to list the wave of each policy?  Maybe something like:
```
grep ztp-deploy-wave source-crs
```

It may not be bad to describe why we have the specific orderings we do, at a high level here.  For example, maybe just a brief description around each existing wave number:

1 - Catalog sources and common policy configuration
2 - Namespaces and operator subscriptions
10 - group-level configuration
11 - ??  What is this?
100 - site-specific configuration

ztp/source-crs/README.md
```diff
@@ -0,0 +1,42 @@
+## Source CRs
```
lack, line 1: I really like having a README in here.  Good idea!

lack (changes requested):

ztp/source-crs/README.md
Missxiaoguo: Great idea to run a simple command to check the waves for source-crs! I totally agree that it's not good to document this waves here as it takes effort to maintain.

lack (approved):

PR #997, opened 2022-03-07
ztp: Bug 2021688: add mutex lock for ArgoCD apps

vitus133:
Adds mutex lock causing all concurrent ArgoCD apps
in a given namespace to execute serially.
This addresses an issue happenning when a concurrent
application deletes and re-creates the resource version
configmap, corrupting the ongoing application data.

The lock is realized with a configmap, created
in the pre-sync hook and deleted just before the post-sync
hook exits. All concurrent pre-sync hooks will block
until the configmap is deleted


serngawy (commented):
The fix sounds good, but how did you validate that fixed the bug ? Would you add steps how you test it. It will help the QA as well

PR #999, opened 2022-03-08
ztp: bug 2002960: handle siteconfig deletion

vitus133:
Works around ACM limitation that requires first delete
a managedcluster object before removing the correspondent
namespace.
/cc @serngawy  @imiller0 
This was tested by creating a 100 siteconfigs and then removing one.
The branch used to setup the environment is https://github.com/vitus133/cnf-features-deploy/tree/siteconfig-delete-setup
Please note that if adding sites is limited to batches of 100, removing is limited to 50. It's because ArgoCD creates two transactions, modify and delete per each deletion, and the cache is still limited to 100.
This can be merged independently of https://github.com/openshift-kni/cnf-features-deploy/pull/997


ztp/resource-generator/src/watcher.py
```diff
@@ -164,6 +165,41 @@ def __init__(self, api_response, resourcename="siteconfigs", debug=False):
...
+            with open(os.path.join(self.del_path, item), "r") as yi:
+                obj = yaml.safe_load(yi)
+            if obj.get("kind", {}) == "SiteConfig":
+                # Assuming here that name and namespace of the
+                # managedcluster are identical
+                name = obj.get("metadata", {}).get("namespace")
+                try:
+                    status = None
+                    cmd = ["oc", "delete", f"managedcluster/{name}",
+                           "-n", f"{name}"]
```
serngawy, line 181: the managedCluster is a cluster scope resource no namespace for it

vitus133: Fixed

serngawy (approved):

PR #1000, opened 2022-03-08
Bug 2061782: ztp: Add mechanism to patch inner container references at build 

lack:
The argocd deployment files refer to the upstream container images by
default. But downstream builds (or other special-purpose builds) need to
have a facility to override that internal reference. This is that
facility. Setting the IMAGE_REF build argument will patch any internal
references to that argument's value verbatim.

- Manual cherry-pick of #985 to release-4.10


imiller0 (approved):
lgtm

PR #1001, opened 2022-03-08
Bug 2058997: Set ice-ptp threads to FIFO 10

josephdrichard:


ztp/source-crs/TunedPerformancePatch.yaml
```diff
@@ -18,6 +18,8 @@ spec:
         cmdline_crash=nohz_full=${isolated_cores}
         [sysctl]
         kernel.timer_migration=1
+        [scheduler]
+        group.ice-ptp=0:f:10:*:ice-ptp.*
```
imiller0, line 22: Unfortunately the same content is replicated in the reference group PolicyGenTemplates. To ensure that customers pick this up can you make the same change in:
* https://github.com/openshift-kni/cnf-features-deploy/blob/d39442342eaaf0514ca935fbe43d4cda1d1fdaa8/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml#L99-L107
* https://github.com/openshift-kni/cnf-features-deploy/blob/d39442342eaaf0514ca935fbe43d4cda1d1fdaa8/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml#L50-L58
* https://github.com/openshift-kni/cnf-features-deploy/blob/d39442342eaaf0514ca935fbe43d4cda1d1fdaa8/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-standard-ranGen.yaml#L50-L58

josephdrichard: done.

imiller0 (approved):
lgtm

PR #1004, opened 2022-03-09
Bug 2050391: ztp: delete wave annotation from the policy-wrapped CRs

Missxiaoguo:
Strip the wave annotation from the built CR after the wave is
assigned to the policy.


Missxiaoguo:
/hold
Testing is in progress

serngawy (changes requested):
Its not right to set the annotation then remove it by code. As enduser I expect to see what i set in the source CRs. Please, add the wave annotation to the policies and remove it from the source-crs.

lack:
> Its not right to set the annotation then remove it by code. As enduser I expect to see what i set in the source CRs. Please, add the wave annotation to the policies and remove it from the source-crs.

The problem with this approach is that it requires that customers create the right values in their PGTs.  The benefit of shipping these in the source-crs is that they are completely invisible to the customer (unless they introduce a mismatch).

Maybe longer-term we need to figure out a better way to more global policy-based things like this in a way such that customers don't have to change their PGTs to take advantage of them.

PR #1007, opened 2022-03-10
Bug 2062493: ztp: fix StoragePVC

serngawy:


imiller0:
Closing PR. This bug is fixed in 4.10 and later.

PR #1016, opened 2022-03-14
Add policy Annotation  & Remove wave annotation from source-crs

serngawy:


lack:
So it seems like we've got a couple different potential paths forward with how we want to manage these wave annotations.

I can see that this approach alleviates one concern:
- The source-crs that are the input to the PGTs arrive in the policies unchanged.

However, I see this as introducing a new problem:
- We would be fundamentally moving a piece of data that we own (wave CR-to-policy mapping values) and moving it into an area that the administrator owns (the PGT CRs). This moves this responsibility from something we largely own (with the ability for the user to override using overlays if needed) to something that is now the responsibility of the end-user (with us only able to make suggestions in our example PGTs).

Perhaps we can find a middle-ground somewhere; Can we find a better place to put these annotations that's not in source-crs but also not in PGT?

imiller0:
As I understand it this PR is trying to resolve the issue where an upgrade from the 4.9 solution (no waves) to the 4.10 solution (with waves) results in all the CRs needing an update on the spoke cluster (due to the new wave annotation). PR #1004 attempts to solve this by removing the wave annotation from the CR when it gets wrapped in policy. Both 1004 and this PR have pros and cons. 
This PR has the benefit of not stripping an annotation. That's good and a "con" against 1004. However there are a couple "cons" (my opinion) to this approach. The biggest issue (same issue as @lack mentioned) is that this approach puts the responsibility onto the user to assemble CRs into waves without allowing for us to provide any default value. What makes it worse is that the ordering is an internal implementation detail that improves robustness and shortens install time. The end user really doesn't care about the ordering, so placing the responsibility to get it right entirely onto them (via their annotations in the PGT) feel overly burdensome.
The second "con" I see is that the wave (order that CRs should be applied) is really an attribute of the CR and not the policy. Specifically, if policies weren't being used as the mechanism to push CRs to the spoke clusters, the concept of waves would still apply. We'd want to apply Namespaces, followed by OperatorGroups, followed by Subscriptions (etc). Because we combine multiple CRs into a single policy we need to "elevate" the waves to the policy so that they still apply in the right order. 

One resolution to the initial issue would be to simply do nothing and let all the policies be "non compliant" due to the added annotation, but from a user experience that also seems like causing pain for no end-user benefit. This annotation is one we introduced for helping to manage the ZTP flow in an end-to-end solution. Managing that annotation (including removing it) as part of the solution is not ideal, but I think acceptable.

imiller0:
Hi @serngawy
Unfortunately I gave a bad example in my previous comment by using Namespaces and Subscriptions because we define these within the same wave. I should have used the CatalogSource and Subscription CRs instead (default waves 1 and 2). I still believe that placing the wave annotation on the source-crs is correct. The order in which CRs should be applied depends on what the CRs are. The order that policies should be applied matters only because of the specific CRs that are inside that policy. Where CRs have the same wave (as in the Namespace and Subscription) it means they should be applied and reconciled at the same time without dependency or ordering between them. In our implementation, where they are in the same policy, this means that the policy engine will continue to apply these CRs concurrently until they are all compliant. We could have annotated the Namespace CRs with “1” and Subscriptions with “2” and enforced a rigid ordering, however we chose to optimize the default solution to use fewer policies and overall speed up installation.

As I said in my earlier comment, I agree with you that removing this annotation from the CRs is not generally a good thing. There isn’t a technical problem with doing so (the spoke cluster does not make any use of this annotation) but it isn’t ideal. PR #1004 was intended to improve the user’s experience during upgrade. Perhaps we need to reconsider the impact of stripping the waves and back-out PR #1004.

I think we need to avoid any solution to this issue (adding order to the applied CRs) which puts the responsibility onto the user as they create their PolicyGenTemplates. The current solution gives the user a default configuration with all the correct waves set. Moving the spec of the waves into PolicyGenTemplate means users would need to update all of their existing PGT (and any newly created PGT) to add this field. 

If PR #1004 did not strip the annotation would that address your concerns?

Hope that helps explain my thinking. Happy to discuss more.

imiller0 (commented):
The ability to set policy annotations is supported in the ACM policy generator. Can we close this PR in anticipation of generating policies using ACM's verion?

PR #1017, opened 2022-03-14
ztp: Add ACM PolicyGenerator to ztp-site-generator container image

serngawy:

### DIFF
--- a/ztp/resource-generator/Containerfile
+++ b/ztp/resource-generator/Containerfile
@@ -15,17 +15,27 @@ WORKDIR $PGT_ROOT
 RUN make build
 WORKDIR $SC_ROOT
 RUN make build
+WORKDIR /go/src/
+RUN curl -L $(curl https://api.github.com/repos/stolostron/policy-generator-plugin/releases/latest | jq -r '.tarball_url') -o policy-generator-plugin.tar.gz
+RUN mkdir policy-generator-plugin && tar -xzvf policy-generator-plugin.tar.gz -C policy-generator-plugin --strip-components=1
+WORKDIR /go/src/policy-generator-plugin
+RUN go mod vendor
+RUN make build-binary
 
 # Container image
 FROM ubi8-minimal
 USER root
 ENV BUILDER_ZTP=/go/src/cnf-features-deploy/ztp
 ENV ZTP_HOME=/home/ztp
+ENV ACM_POLICY_GEN=/kustomize/plugin/policy.open-cluster-management.io/v1/policygenerator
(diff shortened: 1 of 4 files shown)


ztp/resource-generator/Containerfile
```diff
@@ -15,17 +15,27 @@ WORKDIR $PGT_ROOT
 RUN make build
 WORKDIR $SC_ROOT
 RUN make build
+WORKDIR /go/src/
+RUN curl -L $(curl https://api.github.com/repos/stolostron/policy-generator-plugin/releases/latest | jq -r '.tarball_url') -o policy-generator-plugin.tar.gz
+RUN mkdir policy-generator-plugin && tar -xzvf policy-generator-plugin.tar.gz -C policy-generator-plugin --strip-components=1
```
imiller0, line 20: Should we have a more explicit versioning here other than "latest". For example, should the 4.11 version of the ztp-site-generator container always pull (example) version 1.2 (no z-stream listed so the latest from 1.2) of stolostron. Then when a new y-stream version comes out there is an explicit switch up to 1.3, etc? This way the dependency between ztp-site-generate container and underlying stolostron/acm version is explicit (and can be tested).

ztp/resource-generator/Containerfile
```diff
@@ -15,17 +15,27 @@ WORKDIR $PGT_ROOT
...
+WORKDIR /go/src/policy-generator-plugin
+RUN go mod vendor
+RUN make build-binary
 
 # Container image
 FROM ubi8-minimal
 USER root
 ENV BUILDER_ZTP=/go/src/cnf-features-deploy/ztp
 ENV ZTP_HOME=/home/ztp
+ENV ACM_POLICY_GEN=/kustomize/plugin/policy.open-cluster-management.io/v1/policygenerator
```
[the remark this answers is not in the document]
serngawy: yes

ztp/resource-generator/Containerfile
```diff
@@ -15,17 +15,27 @@ WORKDIR $PGT_ROOT
 RUN make build
 WORKDIR $SC_ROOT
 RUN make build
+WORKDIR /go/src/
+RUN curl -L $(curl https://api.github.com/repos/stolostron/policy-generator-plugin/releases/latest | jq -r '.tarball_url') -o policy-generator-plugin.tar.gz
```
lack, line 19: Running Curl as part of 'podman build' won't work for downstream builds; it needs to be completely offline-capable.

serngawy: is there a way we can validate the downstream build ?

PR #1018, opened 2022-03-14
Bug 2062450: ztp: Enable kdump in the DU profile

daosman:
To enable the Crash recovery kernel arming service, we need to add the
kernel command line argument 'crashkernel=256M' and enable the
'kdump.service' systemd service through a MachineConfig.

Addresses Bug 2062450


ztp/extra-manifests-builder/06-kdump/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+GOPATH=${GOPATH:-${HOME}/go}
+GOBIN=${GOBIN:-${GOPATH}/bin}
+MCMAKER=${MCMAKER:-${GOBIN}/mcmaker}
+MCPROLE=${MCPROLE:-master}
+
+${MCMAKER} -stdout -name 06-kdump-enable -mcp ${MCPROLE} \
+	unit -source kdump.service -name kdump.service -enable=true
```
lack, line 9: Looks like I should update mcmaker so it can just enable a unit without having to supply the source of it...

ztp/source-crs/PerformanceProfile.yaml
```diff
@@ -8,6 +8,7 @@ spec:
   additionalKernelArgs:
   - "idle=poll"
   - "rcupdate.rcu_normal_after_boot=0"
+  - "crashkernel=256M"
```
lack, line 11: I would rather we put this in the 06-kdump-enable-master file and not here...  Which would take another edit to mcmaker :)

ztp/extra-manifests-builder/06-kdump/build.sh
daosman: Yes, all we need is to enable the service. Not include anything

ztp/source-crs/PerformanceProfile.yaml
lack: That would be fine! Just remember that we need to make a master and worker, so maybe still use a generator script like sctp does to avoid duplication.

lack (approved):

PR #1022, opened 2022-03-16
ztp: bug 2002628: gracefully reconcile deleted policies

vitus133:
Removing one of the policies from the PGT does
not lead to removal of the generated policy from the
hub and does not remove the related objects from the spokes.
This commit adds a comparison between the generated policies
and the policies present on the hub. If some of the
policies have been removed from the PGT list but still configured
on the hub, they are being deleted.

/cc @serngawy @imiller0
This is still not covering a case when a PGT is deleted in full. In this case the policies are deleted from the hub, but the objects remain on the spokes. This is WIP.

### DIFF
--- a/ztp/resource-generator/src/watcher.py
+++ b/ztp/resource-generator/src/watcher.py
@@ -11,6 +11,12 @@ from kubernetes import client, config
 import logging
 
 
+def find_files(root):
+    for d, dirs, files in os.walk(root):
+        for f in files:
+            yield os.path.join(d, f)
+
+
 class Logger():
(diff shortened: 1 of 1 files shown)


serngawy (changes requested):
I believe its better to remove the ManagedClusterAction code  and move it to another PR. 
lets keep the code clean and separate the 2 issues

vitus133:
> I believe its better to remove the ManagedClusterAction code and move it to another PR. lets keep the code clean and separate the 2 issues

Done

PR #1026, opened 2022-03-17
ztp: Enable kdump in the DU profile

daosman:
To enable the Crash recovery kernel arming service, we need to add the
kernel command line argument 'crashkernel=256M' and enable the
'kdump.service' systemd service through a MachineConfig.

Addresses Bug 2062450


lack:
/lgtm

I also verified that running the site-config generator with this in place adds it to the final ConfigMap.

imiller0 (approved):
lgtm

PR #1029, opened 2022-03-18
ztp: secure boot doc for ZTP 

pixelsoccupied:
Upstream doc secure boot. Includes instruction on how to use mokutil 
CNF-[4157](https://issues.redhat.com/browse/CNF-4157)
CNF-[3096](https://issues.redhat.com/browse/CNF-3096)

### DIFF
--- a/ztp/gitops-subscriptions/argocd/SecureBoot.md
+++ b/ztp/gitops-subscriptions/argocd/SecureBoot.md
@@ -1,0 +1,115 @@
+# Enable and Verify for Secure Boot for SNO
+
+## Enabling Secure Boot 
+
+ - Configure `SiteConfig` with `UEFIScureBoot` for `bootMode`. E.g part of yaml below 
+   ```yaml
+    nodes:
+    - hostName: "myhost"
+      bootMode: "UEFISecureBoot"
+   ```
+
+- This value should now be available in the Hub (cluster w/ ACM) cluster's appropriate BareMetalHost (bmh) CR. e.g part of the yaml below 
+  ```yaml
+  spec:
+    bootMode: UEFISecureBoot
+  ```
+
+## Verifying Secure Boot
+
+1. log into the node
+
+    - With `oc`
+      ```shell
+      oc debug node/myhost.rh.com
+      
+      # don't forget to link the logs 
+      sh-4.4#  chroot /host
+      ```
+  
+    - **With `ssh` command** 
+       ```shell
+       ssh -i key core@myhost.rh.com # key: private key associated with the node
+       ```
+  
+2. Once logged into the node there are a couple of ways to verify. 
+
+    - **With `journalctl`**
+      ```shell
+      sh-4.4# journalctl -g secureboot
+      -- Logs begin at Wed 2022-03-23 17:14:14 UTC, end at Fri 2022-03-25 16:46:26 UTC. --
+      Mar 23 17:14:14 localhost kernel: secureboot: Secure boot enabled
+      -- Reboot --
+      Mar 23 17:20:00 localhost kernel: secureboot: Secure boot enabled
+      -- Reboot --
+      Mar 23 17:54:41 localhost kernel: secureboot: Secure boot enabled
+      -- Reboot --
+      Mar 23 18:04:16 localhost kernel: secureboot: Secure boot enabled
+      ```
+
+    - **With `mokutil`**
+      ```shell
+      mokutil --sb-state
+      SecureBoot enabled
+      ```
+
+**Debugging mokutil**
+
+If you're not using the latest set of source-crs and are running the real-time kernel, `mokutil` may return the following error message:
+
+```shell
+mokutil --sb-state
+EFI variables are not supported on this system
+```
+
+There are multiple ways to enable kernel access to the EFI variables that are required by `mokutil`: 
+
+- Update to the latest set of source-cr
+
+- Append `PerformanceProfile.yaml`'s  `additionalKernelArgs` from PGT
+
+  ```yaml
+  spec:
+    additionalKernelArgs:
+      - ...
+      - "efi=runtime"
+  ```
+- Use SiteConfig's `sno-extra-manifest` feature.
+
+  1. Create MachineConfig CR `99-efi-runtime-path-kargs.yaml`
+
+   ```yaml
+   apiVersion: machineconfiguration.openshift.io/v1
+   kind: MachineConfig
+   metadata:
+     labels:
+       machineconfiguration.openshift.io/role: master
+     name: 99-efi-runtime-path-kargs
+   spec:
+     kernelArguments:
+     - "efi=runtime"
+  ```
+  2. Include the `MC` with your `SiteConfig` as part of extra manifest. E.g part of the CR below
+  
+     ```yaml
+     clusters:
+     - clusterName: "myhost"
+       extraManifestPath: myhost/sno-extra-manifest/
+     ```
+
+     File dir may look like below 
+
+     ```shell
+     ➜ tree .
+       └── install
+       ├── myhost
+       │   ├── myhost.yaml
+       │   ├── secret.yaml
+       │   └── sno-extra-manifest
+       │       └── 99-efi-runtime-path-kargs.yaml
+       └── kustomization.yaml
+
+     ```
+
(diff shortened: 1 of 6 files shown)


ztp/gitops-subscriptions/argocd/SecureBoot.md
```diff
@@ -0,0 +1,80 @@
...
+```yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 99-efi-runtime-path-kargs
+spec:
+  kernelArguments:
+  - "efi=runtime"
```
lack, line 58: Maybe it would be better to add this to the PAO performance profile where we enable the RT kernel, since it's only needed if you have the RT kernel installed anyway...

ztp/gitops-subscriptions/argocd/SecureBoot.md
```diff
@@ -0,0 +1,80 @@
...
+    bootMode: "UEFISecureBoot"
+```
+
+- This value should now be available in the Hub (cluster w/ ACM) cluster's appropriate BareMetalHost (bmh) CR. e.g part of the yaml below 
+```yaml
+spec:
+  bootMode: UEFISecureBoot
+```
+
+## How Verify Secure Boot Status
```
lack, line 18:
```suggestion
## Verifying Secure Boot
```

ztp/gitops-subscriptions/argocd/SecureBoot.md
```diff
@@ -0,0 +1,80 @@
...
+
+Right now the installation verification is done by `ssh`-ing into the node. Credentials such as private key associated with the node are required.
+E.g `ssh` command 
+```shell
+ssh -i key core@myhost.rh.com
+```
+
+Once logged into the node there are a couple of ways to verify. 
+  - **With `journalclt`**
```
lack, line 27:
```suggestion
  - **With `journalctl`**
```

ztp/gitops-subscriptions/argocd/SecureBoot.md
```diff
@@ -0,0 +1,80 @@
...
+Feb 25 01:25:02 localhost kernel: secureboot: Secure boot enabled
+```
+
+  - **With `mokutil`**
+```shell
+mokutil --sb-state
+SecureBoot enabled
+```
+
+`motutil` may not work correctly and see something like this below
```
lack, line 39:
```suggestion
If you are running the real-time kernel, `motutil` may return the following error message:
```

ztp/gitops-subscriptions/argocd/SecureBoot.md
```diff
@@ -0,0 +1,80 @@
...
+```shell
+mokutil --sb-state
+EFI variables are not supported on this system
+```
+
+To use `mokutil` do the following:
```
lack, line 45:
```suggestion
To enable kernel access to the EFI variables that are required by `mokutil`, do the following: 
```

ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml
```diff
@@ -45,6 +45,7 @@ spec:
         bmcCredentialsName:
           name: "example-node1-bmh-secret"
         bootMACAddress: "AA:BB:CC:DD:EE:11"
+        # Use UEFISecureBoot to enable secure boot
```
lack, line 48: We should maybe add this same comment to all the example SiteConfigs?

ztp/gitops-subscriptions/argocd/SecureBoot.md
```diff
@@ -0,0 +1,80 @@
+# Enable and Verify for Secure Boot for SNO
+
+## How to Enable Secure Boot
```
lack, line 3:
```suggestion
## Enabling Secure Boot 
```

ztp/gitops-subscriptions/argocd/SecureBoot.md
```diff
@@ -0,0 +1,80 @@
...
+  bootMode: UEFISecureBoot
+```
+
+## How Verify Secure Boot Status
+
+Right now the installation verification is done by `ssh`-ing into the node. Credentials such as private key associated with the node are required.
+E.g `ssh` command 
+```shell
+ssh -i key core@myhost.rh.com
+```
```
lack, line 24: Another way to do this besides ssh could be `oc debug mode/nodename`; This is maybe a more OCP-feiendly way of getting a shell on a node?

lack (changes requested):

ztp/gitops-subscriptions/argocd/SecureBoot.md
pixelsoccupied: Node debug does pick up logs from node's journalctl or dmesg 
```
sh-4.4# journalctl | grep secureboot 
No journal files were found.
````

File: ztp/gitops-subscriptions/argocd/SecureBoot.md (hunk 1 above)
Comment (pixelsoccupied) [reply ztp/gitops-subscriptions/argocd/SecureBoot.md:58]: So appending to [this](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/source-crs/PerformanceProfile.yaml#L8) list? That would mean the user would enable the runtime regardless of using SecureBoot right? Is that what we would want now?

File: ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml (hunk 6 above)
Comment (pixelsoccupied) [reply ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml:48]: I tried this config only for SNOs. But since this config ultimately is part of BMH CR I think it's safe to add the comment.

File: ztp/gitops-subscriptions/argocd/SecureBoot.md (hunk 1 above)
Comment (lack) [reply ztp/gitops-subscriptions/argocd/SecureBoot.md:58]: I don't think there's any harm in enabling this, though I think we'd have to be careful and check what happens if you add this entry to the non-realtime kernel.   But it's much more a feature of the RT kernel and whether it has access to any EFI variables, regardless of whether or not secure boot is enabled.

Comment (lack) [reply ztp/gitops-subscriptions/argocd/SecureBoot.md:58]: To put it another way, if you install the RT kernel and don't add this kernel argument, and try to use mokutils to verify that secureboot is NOT enabled, you would have the same error regardless of whether secureboot is enabled on not enabled!

Comment (pixelsoccupied) [reply ztp/gitops-subscriptions/argocd/SecureBoot.md:58]: Can confirm adding `efi=runitme` to `PerformanceProfile` and setting RT Kernel to false, mokutil works as expected 

```
[core@cnfdf04 ~]$ uname -r     
4.18.0-305.34.2.el8_4.x86_64
[core@cnfdf04 ~]$  mokutil --sb-state
SecureBoot enabled

```

Comment (lack) [reply ztp/gitops-subscriptions/argocd/SecureBoot.md]: Excellent!

File: ztp/gitops-subscriptions/argocd/SecureBoot.md (hunk 8 above)
Comment (lack) [reply ztp/gitops-subscriptions/argocd/SecureBoot.md:24]: You might need to link up /var/log properly - not sure how that's normally done for 'oc debug'...  You may need to also run `chroot /host` first.

Also, `journalctl -g secureboot` may be slightly more efficient than `journalctl | grep secureboot`

Comment (pixelsoccupied) [reply ztp/gitops-subscriptions/argocd/SecureBoot.md:24]: You're 100%! The message was right there 😶 

```
[core@cnfdf04 ~]$ oc debug node/cnfdf04.ran.dfwt5g.lab
Starting pod/cnfdf04randfwt5glab-debug ...
To use host binaries, run `chroot /host`
```

`-g` looks good 

```
sh-4.4# journalctl -g secureboot
-- Logs begin at Wed 2022-03-23 17:14:14 UTC, end at Fri 2022-03-25 16:46:26 UTC. --
Mar 23 17:14:14 localhost kernel: secureboot: Secure boot enabled
-- Reboot --
Mar 23 17:20:00 localhost kernel: secureboot: Secure boot enabled
-- Reboot --
Mar 23 17:54:41 localhost kernel: secureboot: Secure boot enabled
-- Reboot --
Mar 23 18:04:16 localhost kernel: secureboot: Secure boot enabled
```

Updating the docs

PR #1037, opened 2022-03-24
[release-4.9] ztp: bug 2068098: Remove workload management annotation from AMQ namespace

aneeshkp:
AMQ interconnect operators are not pinned to platform cores, and the AMQ namespace has an annotation to manage workload.
So removing workload management annotations from the namespace


nishant-parekh (approved):

lack:
Shouldn't the AMQ interconnect operators be pinned to the management cores instead?

PR #1038, opened 2022-03-24
Bug 2026130: fix creation of empty policy objects

sakhoury:
When a source CR for policyGen contains a trailing "---" separator, the generated policy contains an empty object. ACM complains that this object does not have a "kind" attribute and the policy never goes compliant.

This commit checks that the decoded YAML objects are not null prior to appending them for further processing.

Signed-off-by: Sharat Akhoury sakhoury@redhat.com
/cc @imiller0 @lack

### DIFF
--- a/ztp/policygenerator/policyGen/policyBuilder.go
+++ b/ztp/policygenerator/policyGen/policyBuilder.go
@@ -328,12 +328,18 @@ func (pbuilder *PolicyBuilder) splitYamls(yamls []byte) ([][]byte, error) {
 		if err != nil {
 			return nil, err
 		}
-		resBytes, err := yaml.Marshal(resIntf)
 
-		if err != nil {
-			return nil, err
+		// Check that resIntf is not nil in order to mitigate appending an empty
+		// object as a result of redundant trailing seperator(s) "---""
+		if resIntf != nil {
+			resBytes, err := yaml.Marshal(resIntf)
+
+			if err != nil {
+				return nil, err
+			}
+
+			resources = append(resources, resBytes)
 		}
-		resources = append(resources, resBytes)
 	}
 	return resources, nil
 }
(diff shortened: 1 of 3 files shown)


lack:
As mentioned on slack, it would be great to have an associated unit test as part of the change please :)

ztp/policygenerator/policyGen/policyBuilder.go
```diff
@@ -333,7 +333,12 @@ func (pbuilder *PolicyBuilder) splitYamls(yamls []byte) ([][]byte, error) {
 		if err != nil {
 			return nil, err
 		}
-		resources = append(resources, resBytes)
+
+		// Check that resBytes is not null in order to mitigate appending an empty
+		// object as a result of redundant trailing seperator(s) "---""
+		if !bytes.Equal(resBytes, []byte("null\n")) {
```
lack, line 339: Rather than checking whether the marshaled result has this somewhat odd string value, would it be better to check resIntf instead before we marshal it?

sakhoury: Good suggestion! I have updated the code to check before the object is marshalled!

ztp/policygenerator/policyGen/testData/GenericSourceFiles/GenericCRWithTrailingSeparators.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: operators.coreos.com/v1alpha1
```
lack, line 1: Perhaps having some leading empty docs would also be a good idea for a test?

sakhoury: Good idea! I have added that.

imiller0 (approved):
lgtm

PR #1040, opened 2022-03-25
ztp: Annotate generated CRs for traceability

tliu2021:
This update annotates all CRs generated by siteconfig-generator,
which also includes the MachineConfig CRs that embedded in the
ConfigMap.


ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// Add ztp deploy annotation to all siteconfig generated CRs
+// A dynamic value might be added later
+func addZTPAnnotation(clusterCRs []interface{}) ([]interface{}, error) {
+
+	for _, v := range clusterCRs {
+		cr := v.(map[string]interface{})
+		metadata := cr["metadata"].(map[string]interface{})
+		annotations, _ := metadata["annotations"].(map[string]interface{})
```
lack, line 16: Do we need the `, _` if we're just trying to get the value?

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,44 @@
...
+		annotations[ztpAnnotation] = nil
+		if cr["kind"] == "ConfigMap" {
+			// Add ztp annotation to embedded machine config CRs
```
lack, line 19: I would much rather see us not special-case the "look at the contents of the ConfigMap" here, but instead call `addZTPAnnotation` on each of the individual MachineConfigs when we process them in siteConfigBuilder.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Add ztp deploy annotation to all siteconfig generated CRs
+// A dynamic value might be added later
+func addZTPAnnotation(clusterCRs []interface{}) ([]interface{}, error) {
+
+	for _, v := range clusterCRs {
+		cr := v.(map[string]interface{})
+		metadata := cr["metadata"].(map[string]interface{})
+		annotations, _ := metadata["annotations"].(map[string]interface{})
+		annotations[ztpAnnotation] = nil
```
lack, line 17: In the testdata below, it looks like we're a bit inconsistent - We have
```
ran.openshift.io/ztp-gitops-generated: null
```
in some places and 
```
ran.openshift.io/ztp-gitops-generated: ""
```
in others.  I'd vote for using the empty string instead of nil everywhere, please!

Or maybe even:
```
ran.openshift.io/ztp-gitops-generated: "true"
```
Or
```
ran.openshift.io/ztp-gitops-generated: "{}"
```

Not sure what a good starting value is, if we want to eventually have the ability for it to be structured data?

lack (changes requested):
Perhaps a unit test for `siteConfigHelper` would be prudent too?

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -18,6 +18,7 @@ const cpuset = "$cpuset"
 const SNO = "sno"
 const Standard = "standard"
 const Master = "master"
+const ztpAnnotation = "ran.openshift.io/ztp-gitops-generated"
```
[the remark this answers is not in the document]
tliu2021: Changed.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: No, it is not needed.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: Ok, addZTPAnnotationToManifest() is called on the individual  manifest.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: It has been changed to 'ran.openshift.io/ztp-gitops-generated: "{}"'.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -54,6 +54,8 @@ func MergeManifests(individualMachineConfigs map[string]interface{}, doNotMerge
 		merged.SetName(fmt.Sprintf("%s-%s", McName, roleName))
 		merged.ObjectMeta.Labels = make(map[string]string)
 		merged.ObjectMeta.Labels[machineconfigv1.MachineConfigRoleLabelKey] = roleName
+		merged.ObjectMeta.Annotations = make(map[string]string)
+		merged.ObjectMeta.Annotations[ZtpAnnotation] = ZtpAnnotationDefaultValue
```
lack, line 58: I know it would only be used in one place for now, but maybe it's worth adding a `addZtpAnnotationToObject(object)` and pass in `merged`?

lack (commented):
Awesome!  One small question (optional) and I think I'm happy to approve this!

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: I don't see it as added value.

lack (approved):

PR #1042, opened 2022-03-25
Bug 2060492: ztp: Update network_transport to L2

josephdrichard:
Currently, only L2 as network_transport is supported.
Example yamls currently set network_transport UDPv4, and then override
with -2 command-line option.
Update to set network_transport to L2 in yaml.


vitus133:
Would it be less confusing to remove this line from all instances of `ptp4lConf`? It's overridden by `ptp4lOpts` anyway.

josephdrichard:
> Would it be less confusing to remove this line from all instances of `ptp4lConf`? It's overridden by `ptp4lOpts` anyway.

My understanding is that the person who reported this bug initially, it's because they removed the -2 from their ptp4lOpts, and then it stopped working.  So I think it is better leaving it in to prevent that.

nishant-parekh (approved):

PR #1043, opened 2022-03-28
Bug 2051540: cnf-tests:cyclictest: remove smi option

Tal-or:
cyclictest does not yet support the --smi option for newer processors (e.g. ice lake).

This will require an update to the rt-tests package to support newer processor models.
Until that is done, the change to add must be backed out, or this test won't work on ice lake processors.

A NOTE: this change is not temporary - it probaly would never be safe to add --smi option here, because of the way cyclictest is written, it is always going to fail on new processor models.


MarSik (approved):
/lgtm

PR #1047, opened 2022-03-30
ztp: Merge extra manifests to one MC

tliu2021:
This update merges all statically defined extra manifest to a single Machine
Config object per role.


ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+	"log"
+
+	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	mcfgctrlcommon "github.com/openshift/machine-config-operator/pkg/controller/common"
+
+	"github.com/lack/yamltrim"
+	yaml "gopkg.in/yaml.v3"
+)
+
+const mcRoleKey = "machineconfiguration.openshift.io/role"
```
lack, line 15: I recently learned we can use `machineconfigv1.MachineConfigRoleLabelKey` instead of copying this value around by hand all the time :)

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+const mcName = "predefined-extra-manifests"
+
+func isExcluded(s []string, str string) bool {
```
lack, line 18: This is a useful function, but maybe better named `listContains`, since it really doesn't have anything to do with exclusion per se?

Also (optional), if you use a `map[string]bool` instead of a `[]string`, you don't need this function at all and the code below could just be:
```
if excludes[k] {
  continue
}
```

(Assuming you insert them as `excludes["filename"] = true` when you populate the map)

This is a cool feature of go maps -> The default value of `bool` is false, and if you only call the map get function with a single-argument return, you get the default value if the key doesn't exist in the map!

Example: https://go.dev/play/p/vZVaxZxXyjm

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+			return nil, err
+		}
+
+		merged.SetName(fmt.Sprintf("%s-%s", mcName, k))
+		merged.ObjectMeta.Labels = make(map[string]string)
+		merged.ObjectMeta.Labels[mcRoleKey] = k
+		merged.TypeMeta.APIVersion = machineconfigv1.GroupVersion.String()
+		merged.TypeMeta.Kind = "MachineConfig"
+
+		// Marshal to json
```
lack, line 61: Maybe a bit of a note would be good about why we want to marshal to json, then unmarshal from json?

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+		err := addMachineConfig(v, configs)
+		if err != nil {
+			return dataMap, err
+		}
+		// remove the individual file entries
+		delete(dataMap, k)
+	}
+
+	for k, v := range configs {
```
lack, line 47: In this case `k`, `v` may be better as `roleName`, `machineConfigs`?

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// merge the spec fields of all MC manifests except the ones that are in the excluded list
+func MergeManifests(dataMap map[string]interface{}, excludes []string) (map[string]interface{}, error) {
+	// key is role, value is a list of MCs
+	configs := make(map[string][]*machineconfigv1.MachineConfig)
```
lack, line 33: To make the purpose more clear, maybe call this variable `mergableMachineConfigs`?

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	return dataMap, nil
+}
+
+// convert yaml data to MC
+func convertToMC(data interface{}) (*machineconfigv1.MachineConfig, error) {
+	var m map[string]interface{}
+
+	err := yaml.Unmarshal([]byte(data.(string)), &m)
```
lack, line 96: A comment about why we're doing an `Unmarshal` followed immediately by a `Marshal` and then another `Unmarshal` might be nice.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -346,6 +346,10 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
 	if err != nil {
 		return nil, err
 	}
+
+	// Manifests to be excluded from merging
+	var excluded_manifests []string
```
lack, line 351: Perhaps "doNotMerge" would be a better name, given the fact that @pixelsoccupied is working on a different kind of exclusion?

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+	mc, err := convertToMC(data)
+	if err != nil {
+		return err
+	}
+	role := mc.ObjectMeta.Labels[mcRoleKey]
+	if _, ok := configs[role]; ok {
+		configs[role] = append(configs[role], mc)
+	} else {
+		var mcPtr []*machineconfigv1.MachineConfig
+		configs[role] = append(mcPtr, mc)
```
lack, line 126: This syntax looks like it works, but feels weird to append to an empty list you just made - Is it cleaner to just set configs[role] to a new slice that contains mc instead? Maybe just:
```
    configs[role] = []*machineconfigv1.MachineConfig{mc}
```

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+	for _, v := range s {
+		if str == v {
+			return true
+		}
+	}
+	return false
+}
+
+// merge the spec fields of all MC manifests except the ones that are in the excluded list
+func MergeManifests(dataMap map[string]interface{}, excludes []string) (map[string]interface{}, error) {
```
lack, line 31: I know it ends up being used as the 'data' map in the ConfigMap somewhere else, but for the purposes of this function, maybe we should call the variable `individualMachineConfigs` instead of `dataMap` for clarity?

lack (changes requested):
Looks great!  I think there are a couple minor cleanups we could do to make the code even easier to understand than it already is.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
pixelsoccupied: Yep so the code I'm working on to filter is run after listing all the default and user provided MCs. So basically on the `dataMap` map and before [L168](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go#L168) 

`doNotMerge` list is something that's provided by the user? Would it be the same thing as a filename under the [exclude](https://github.com/openshift-kni/cnf-features-deploy/pull/1048/files#diff-5a7eeb6977f4204a7d5c1322a7e1fe7957d13801d22c91348770ef8a71c22466R144) list? 

I think the @tliu2021's `func MergeManifests(dataMap map[string]interface{},`'s `dataMap` should be the one after it's run through the filter? [Algorithm](https://docs.google.com/document/d/1xYD-cVlSw6HXUAdWTLad-JailWkaXO6I_u2Fq6VNpyc/edit) here might make more sense

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: Thank you for your insight.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: Thanks for the suggestion.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: I added a note.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -0,0 +1,129 @@
...
+	// key is role, value is a list of MCs
+	configs := make(map[string][]*machineconfigv1.MachineConfig)
+
+	for k, v := range dataMap {
```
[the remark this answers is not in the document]
tliu2021: Agree, k and v have been renamed to filename and machineConfig.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: Yes.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: OK,  the variable has been renamed.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: OK.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
tliu2021: Yes, it has been renamed. 

@pixelsoccupied as we discussed, both actions can be done inside 'getExtraManifests' .

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: Agree, the code has been updated.

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
tliu2021: OK, the variable has been renamed.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -341,6 +341,135 @@ func Test_getExtraManifestTemplatedRoles(t *testing.T) {
 	}
 }
 
+func Test_mergeExtraManifests(t *testing.T) {
```
lack, line 344: These are great tests, but the ones testing the `MergeManifests` function should be in `siteConfigHelper_test.go` instead.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -347,6 +351,7 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
 				}
 				if value != "" {
 					dataMap[filename] = value
+					doNotMerge[filename] = true
```
lack, line 354: It's probably worth a comment describing why we're excluding all templated MCs: just because the only templated MC is an installation-only MC.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
tliu2021: The ones testing the MergeManifests function have been moved to siteConfigHelper_test.go.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -361,6 +367,13 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
 		}
 	}
 
+	// merge the pre-defined manifests
```
lack, line 370: Maybe it's not that important, but for completeness I'd actually prefer this get moved until after the "workload partitions" MC, and the explicitly exclude that one from merging as well.  So basically it groups:
(1) All our provided MCs (some of which should not be merged)
(2) The merge step
(3) All customer-provided MCs

lack (approved):

PR #1048, opened 2022-03-30
Bug 2065969: ztp: filter ztp extra manifest

pixelsoccupied:
When installing clusters using SiteConfig, RAN team provides a specific set of MachineConfigs (though other types are also valid) as well as a set of extra CRs from the user (dir specified under `ExtraManifestPath `)

This PR should help users with controlling the files they would like to add. (check doc for more info)


ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -464,3 +472,38 @@ func (scbuilder *SiteConfigBuilder) splitYamls(yamls []byte) ([][]byte, error) {
...
 	return resources, nil
 }
+
+func filterZtpExtraManifests(dataMap map[string]interface{}, manifests *ZtpExtraManifests) (map[string]interface{}, error) {
+	// return if there's no ZtpExtraManifests initialized
+	if manifests == nil {
+		return dataMap, nil
+	}
+
+	// default value is include. treat use of `exclude` as an advanced feature
```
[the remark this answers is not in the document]
pixelsoccupied: Yep definitely prone to headache. but the use case is exactly what you just mentioned imo i.e admins who want to pre-emptively exclude future capabilities can use `inclusionDefault: Exclude` and then explicitly include anything they want the cluster to have.

Are we saying we don't want this level feature? @imiller0 @serngawy agreed?

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -139,24 +139,31 @@ type SshPrivateKeySecretRef struct {
...
+	ClusterLabels          map[string]string  `yaml:"clusterLabels"`
+	NetworkType            string             `yaml:"networkType"`
+	ClusterNetwork         []ClusterNetwork   `yaml:"clusterNetwork"`
+	IgnitionConfigOverride string             `yaml:"ignitionConfigOverride"`
+	DiskEncryption         DiskEncryption     `yaml:"diskEncryption"`
+	ProxySettings          ProxySettings      `yaml:"proxy,omitempty"`
+	ExtraManifestPath      string             `yaml:"extraManifestPath"`
+	ClusterImageSetNameRef string             `yaml:"clusterImageSetNameRef,omitempty"`
+	BiosConfigRef          BiosConfigRef      `yaml:"biosConfigRef"`
+	ZtpExtraManifests      *ZtpExtraManifests `yaml:"ztp_extra_manifests,omitempty"`
```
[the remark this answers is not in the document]
pixelsoccupied: how about `extraManifestsFilter` (similar to ExtraManifestPath)? Though nicest would be combine them together.
```
 extraManifest: 
  path: "bla/"
  filter: 
    exclude: 
      - myfile.yaml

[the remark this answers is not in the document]
Comment (lack) [reply ztp/siteconfig-generator/siteConfig/siteConfig.go:166]: Ah, that's nice too!

File: ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (serngawy) [reply ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go:482]: Its better to keep exclude, otherwise we will need to list every MC to exclude them all

File: ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go [hunk 3]
```diff
@@ -418,6 +411,20 @@ func (scbuilder *SiteConfigBuilder) getExtraManifest(dataMap map[string]interfac
...
+		return dataMap, err
+	}
+
+	// merge the pre-defined manifests
+	dataMap, err = MergeManifests(dataMap, doNotMerge)
+	if err != nil {
+		log.Printf("Error could not merge extra-manifest %s.%s %s\n", clusterSpec.ClusterName, clusterSpec.ExtraManifestPath, err)
+		return dataMap, err
+	}
+
```
pixelsoccupied, line 427: @tliu2021 could take a look at this when you get a chance? The new filter function needs to have the full list before you can merge.

pixelsoccupied: Actually moving it causes tests to fail. Need your help with this

lack: By moving the merge function down here you're implicitly adding all the user manifests to the merged MachineConfig.  If you set the `doNotMerge[file]=true` inside the user-supplied extra manifests loop (line 388) I bet it will work again.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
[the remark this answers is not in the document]
lack: But why would we want to support that use case?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -478,3 +485,84 @@ func (scbuilder *SiteConfigBuilder) splitYamls(yamls []byte) ([][]byte, error) {
...
+	inclusionDefaultInclude := "include"
+	inclusionDefaultExclude := "exclude"
+	// default value is include. treat use of `exclude` as an advanced feature
+	if filter.InclusionDefault == nil {
+		filter.InclusionDefault = &inclusionDefaultInclude
+	} else if strings.EqualFold(*filter.InclusionDefault, inclusionDefaultExclude) {
+		filter.InclusionDefault = &inclusionDefaultExclude
+	} else if strings.EqualFold(*filter.InclusionDefault, inclusionDefaultInclude) {
+		filter.InclusionDefault = &inclusionDefaultInclude
```
lack, line 525: Rather than doing string comparisons again below (line 552), you could probably use a boolean internally like `excludeAllByDefault`.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -478,3 +485,84 @@ func (scbuilder *SiteConfigBuilder) splitYamls(yamls []byte) ([][]byte, error) {
...
+			value, exists := dataMap[fileToInclude]
+			if exists {
+				temp[fileToInclude] = value
+			} else {
+				errStr := fmt.Sprintf("Filename %s under include array is invalid. Valid files names are: %s",
+					fileToInclude, getDataMapFileNameInStrings(dataMap))
+				return dataMap, errors.New(errStr)
+			}
+		}
+		return temp, nil
```
lack, line 564: If you move this whole block up before the previous one, I think it might flow better:

1. if we're in "exclude" mode, do this block and return the copied  `temp` map.

2. If we bypass this short-circuit return, we know we're in "include" mode...  So just delete any entries in the exclude list from dataMap (line 541-549) and return the dataMap.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -478,3 +485,84 @@ func (scbuilder *SiteConfigBuilder) splitYamls(yamls []byte) ([][]byte, error) {
...
 }
+
+func filterExtraManifests(dataMap map[string]interface{}, filter *Filter) (map[string]interface{}, error) {
+	// return if there's no filter initialized
+	if filter == nil {
+		return dataMap, nil
+	}
+
+	// check if there is an intersection in filenames
+	if filter.Include != nil && filter.Exclude != nil && len(filter.Exclude) > 0 && len(filter.Include) > 0 {
```
lack, line 496: A simpler sanity check would be:
- if the default is "include", the include list must be empty (because it is only valid to exclude specific files in this mode)
- elif the default is "exclude", the exclude list must be empty (because it is only valid to include specific files in this mode)

(I think that would cover everything, without traversing the lists?)

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
pixelsoccupied: Yep tried that. Some kind of path issues it seems (tests pass if I move it back)...investigating.

pixelsoccupied: Had to revert the test file used to assert `Pre-defined extra-manifest cannot be over written`. Looking good now.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
pixelsoccupied: Yep I think so too. But the purpose of the traversal is get the names of all the offending files to help users debug quickly -- `Can not have same file names in both exclude and include lists. Files %s are present in both include and exclude lists", stringFiles`

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
pixelsoccupied: done

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
pixelsoccupied: added a bool!

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
lack: If we short-circuit it to say "When in exclude mode you can't have anything in the exclude list", we don't even need to report the list contents though...

pixelsoccupied: done

lack (approved):

PR #1055, opened 2022-04-05
Bug 2054407: ztp: add the Operator source CR for each subscription

Missxiaoguo:
The new operator resource added is used for determining
the successful operator install/upgrade.

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
@@ -16,11 +16,15 @@ spec:
       policyName: "subscriptions-policy"
     - fileName: SriovSubscriptionOperGroup.yaml
       policyName: "subscriptions-policy"
+    - fileName: SriovOperatorStatus.yaml
+      policyName: "subscriptions-policy"
     - fileName: PtpSubscription.yaml
       policyName: "subscriptions-policy"
     - fileName: PtpSubscriptionNS.yaml
       policyName: "subscriptions-policy"
     - fileName: PtpSubscriptionOperGroup.yaml
+      policyName: "subscriptions-policy"
+    - fileName: PtpOperatorStatus.yaml
       policyName: "subscriptions-policy"
     - fileName: PaoSubscription.yaml
       policyName: "subscriptions-policy"
@@ -31,17 +35,23 @@ spec:
       policyName: "subscriptions-policy"
     - fileName: PaoSubscriptionOperGroup.yaml
       policyName: "subscriptions-policy"
+    - fileName: PaoOperatorStatus.yaml
+      policyName: "subscriptions-policy"
     - fileName: ClusterLogNS.yaml
       policyName: "subscriptions-policy"
     - fileName: ClusterLogOperGroup.yaml
       policyName: "subscriptions-policy"
     - fileName: ClusterLogSubscription.yaml
       policyName: "subscriptions-policy"
+    - fileName: ClusterLogOperatorStatus.yaml
+      policyName: "subscriptions-policy"
     - fileName: StorageNS.yaml
       policyName: "subscriptions-policy"
     - fileName: StorageOperGroup.yaml
       policyName: "subscriptions-policy"
     - fileName: StorageSubscription.yaml
+      policyName: "subscriptions-policy"
+    - fileName: StorageOperatorStatus.yaml
       policyName: "subscriptions-policy"
     - fileName: ReduceMonitoringFootprint.yaml
       policyName: "config-policy"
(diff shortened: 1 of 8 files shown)


Missxiaoguo:
Have tested ZTP and operator upgrade workflow.

lack:
Auto-generating these without using an explicit PGT entry would also remove the fact that these are all nearly-identical copies of eachother, which always makes me worried.

ztp/source-crs/AmqOperatorStatus.yaml
```diff
@@ -0,0 +1,26 @@
+# This CR verifies the installation/upgrade of the Amq Operator
```
lack, line 1: These are all copies of each other, which seems like a tricky thing to manage; Can we either auto-generate them at runtime, or auto-generate them at build time from a common template, instead of just copying them multiple times?

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
```diff
@@ -16,12 +16,16 @@ spec:
       policyName: "subscriptions-policy"
     - fileName: SriovSubscriptionOperGroup.yaml
       policyName: "subscriptions-policy"
+    - fileName: SriovOperatorStatus.yaml
+      policyName: "subscriptions-policy"
```
lack, line 20: Requiring a user to add to their PGT is something we should try to minimize.

In this case, since there's no spec that would need overriding, can we just autogenerate these for every Subscription, instead of making it opt-in on a per-subscription basis like this?

lack (changes requested):

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
Missxiaoguo: Yeah, it's a fighting between the options: user to add them in PGT vs auto-append these CRs to the policy.  Both have pros and cons. If user would like to install/upgrade operators with a successful indication of when the operators is installed/upgraded, then these resources should be included, otherwise, the deploy of these subscriptions only means the CRs are applied. Auto-generating these CRs in policy removes the possibility to delete these CRs from the policy in the case that one operator has issue that prevents the sub policy to become compliant and the rest of configuration to be applied.  Another consideration is it's unlikely to do this when transfering to use ACM policy generator.  Right, it's an effort for user to adjust their PGTs but from a different perspective, this is the way PGT works. User includes resources via PGT to generates policies. Thinking about if the operator status is considered to be auto-created so why not for ns, operatorgroup and even subscription,  no much difference for those CRs. Currently, the pattern is one ns, one operatorgroup and one subscription per operator, so have one Operator source CR for each associated subscription to avoid user to override the name and namespace.

lack: > Yeah, it's a fighting between the options: user to add them in PGT vs auto-append these CRs to the policy.

Always!

> Auto-generating these CRs in policy removes the possibility to delete these CRs from the policy in the case that one operator has issue that prevents the sub policy to become compliant and the rest of configuration to be applied.

This is a good point.  Perhaps we could be more choosy with which status we check?  Do we need to check all three Subscription/InstallPlan/CSV statuses?  On the other hand, if the operator isn't behaving as we expect, do we want the user to go on?  Or would a better user action to be to either call support or remove the operator's subscription?

> Another consideration is it's unlikely to do this when transfering to use ACM policy generator. Right, it's an effort for user to adjust their PGTs but from a different perspective, this is the way PGT works. User includes resources via PGT to generates policies.

I actually think this is a major point AGAINST the user using the ACM PGs directly, and a good case for why we should continue to provide a better user experience by converting user input (maybe PGT, maybe something else) to ACM PGs (or whatever formats we decide on in the future)

> Thinking about if the operator status is considered to be auto-created so why not for ns, operatorgroup and even subscription, no much difference for those CRs. Currently, the pattern is one ns, one operatorgroup and one subscription per operator, so have one Operator source CR for each associated subscription to avoid user to override the name and namespace.

I agree, and I think we should actually move in that direction :)  I'd love if the way the user told us about what operator they wanted consisted of the bare minimum and we auto-generated the rest.  Or better still - We start with all the operators we recommend, and just give a user to opt-out or override parts of them if they need to.  But probably not today.

For today I really want to see if we can move in the right direction with this object specifically, because it's not something the user needs to have in order for the operator to install, and it's not anything the user needs to specify what they need to happen - It's really something we need for our own purposes (of knowing when the operator is installed).  And I think for something like that, it's better to not have to involve the user at all!

As a middle ground, to maybe mitigate the legitimate danger you raised about an operator behaving badly, could we default to auto-generating this Operator object in the policy, and invent some opt-out mechanism to either remove it or override it on a case-by-cases basis?

lack (approved):

PR #1056, opened 2022-04-06
Bug 2064825: add sanity checks prior to instantiating CR

sakhoury:
The clusters app panics when klusterletaddonconfig via siteconfig has missing attributes / components such as the argocd sync-wave annotation.

This commit adds sanity checks to verify the existence of these attributes prior to instantiating the CR to avoid the application from panicking. If there are missing attributes, an appropriate error is returned instead!

Signed-off-by: Sharat Akhoury [sakhoury@redhat.com](mailto:sakhoury@redhat.com)
/cc @imiller0 @lack

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
@@ -230,14 +230,27 @@ func (scbuilder *SiteConfigBuilder) instantiateCR(target string, originalTemplat
 
 	// Sanity-check the resulting metadata to ensure it's valid compared to the non-overridden original CR:
 	originalMetadata := originalCR["metadata"].(map[string]interface{})
-	overriddenMetadata := overriddenCR["metadata"].(map[string]interface{})
+
+	// Sanity-check the overridden metadata object to ensure that it's instantiated correctly
+	overriddenMetadata, ok := overriddenCR["metadata"].(map[string]interface{})
+	if !ok {
+		return override, fmt.Errorf("Overriden template metadata in %q is not specified!", overridePath)
+	}
+
 	for _, field := range []string{"name", "namespace"} {
 		if originalMetadata[field] != overriddenMetadata[field] {
 			return overriddenCR, fmt.Errorf("Overridden template metadata.%s %q does not match expected value %q", field, overriddenMetadata[field], originalMetadata[field])
 		}
 	}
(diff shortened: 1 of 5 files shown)


ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -230,15 +230,32 @@ func (scbuilder *SiteConfigBuilder) instantiateCR(target string, originalTemplat
...
+	if overriddenMetadata["annotations"] == nil {
+		return override, fmt.Errorf("Overriden template metadata annotations in %q is not specified!", overridePath)
+	}
 	overriddenAnnotations := overriddenMetadata["annotations"].(map[string]interface{})
+
+	// Sanity-check the argocd annotation
 	argocdAnnotation := "argocd.argoproj.io/sync-wave"
+	if overriddenAnnotations[argocdAnnotation] == nil {
+		return override, fmt.Errorf("Overriden template metadata argocd annotation in %q is not specified!", overridePath)
+	}
```
lack, line 257: Should this be allowed if the `originalAnnotations[argocdAnnotation]` is also nil?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -230,15 +230,32 @@ func (scbuilder *SiteConfigBuilder) instantiateCR(target string, originalTemplat
 
 	// Sanity-check the resulting metadata to ensure it's valid compared to the non-overridden original CR:
 	originalMetadata := originalCR["metadata"].(map[string]interface{})
+
+	// Sanity-check the overridden metadata object to ensure that it's instantiated correctly
+	if overriddenCR["metadata"] == nil {
+		return override, fmt.Errorf("Overriden template metadata in %q is not specified!", overridePath)
+	}
```
lack, line 237: Another way to do this same check is to use the more lenient type assertion syntax like this:

```
overriddenMetadata, ok := overriddenCR["metadata"].(map[string]interface{})
if !ok {
  return override, err
}
```
(https://go.dev/tour/methods/15)

In fact, you probably don't even have to check `if !ok` since in that more lenient check, it actually returns an empty map without panicking, so you can just change line 238 to:

```
overriddenMetadata, _ := overriddenCR["metadata"].(map[string]interface{})
```
and it may just avoid panic (and hit the pre-existing error case just below) if there's a mismatch...

lack (approved):

PR #1062, opened 2022-04-07
ztp: add latest label for policies in master branch.

serngawy:

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,10 @@ This repository follows the same version numbering and release branching schedul
 
 1. Create a new 'release-x.y' branch and push it into Git
 2. Advance the CI configuration in openshift/release to create new lanes for the release branch and move the master along to the next release version number
-   - Example: https://github.com/openshift/release/pull/26172
-3. Pin all midstream branches to the new release branch, and create new midstream branches for the next release
+    - Example: https://github.com/openshift/release/pull/26172
+3. Change the ZTP templates examples label `du-profile: latest` under [gitops-subscriptions/argocd/example](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example) to match the branch release-x.y
+    - PGT Example: https://github.com/openshift-kni/cnf-features-deploy/pull/1063
+    - SiteConfig Example: https://github.com/openshift-kni/cnf-features-deploy/pull/1064
+4. Pin all midstream branches to the new release branch, and create new midstream branches for the next release
    - [cnf-tests](https://code.engineering.redhat.com/gerrit/admin/repos/cnf-tests)
    - [ztp-site-generate](https://code.engineering.redhat.com/gerrit/admin/repos/ztp-site-generate)
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
@@ -2,12 +2,13 @@
 apiVersion: ran.openshift.io/v1
 kind: PolicyGenTemplate
 metadata:
-  name: "common"
+  name: "common-latest"
   namespace: "ztp-common"
 spec:
   bindingRules:
     # These policies will correspond to all clusters with this label:
     common: "true"
+    du-profile: "latest"
   sourceFiles:
     # Create operators policies that will be installed in all clusters
     - fileName: SriovSubscription.yaml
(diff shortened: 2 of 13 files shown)


lack:
We should also add a note to the top-level `README.md` in the section about branching, reminding our future selves to set these to a version-specific value in the new release branch.

serngawy:
> We should also add a note to the top-level `README.md` in the section about branching, reminding our future selves to set these to a version-specific value in the new release branch.

Agree, it could be better if we can automate that similar to the [script](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/resource-generator/tools/patchImageReference.sh) we have for the container image

lack:
> > We should also add a note to the top-level `README.md` in the section about branching, reminding our future selves to set these to a version-specific value in the new release branch.
> 
> Agree, it could be better if we can automate that similar to the [script](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/resource-generator/tools/patchImageReference.sh) we have for the container image

Sure, automation is even better; but the bare minimum should be a note in the README :)

serngawy:
@lack would you review/Approve this PR (I forget about it)

lack (approved):

lack (approved):

README.md
```diff
@@ -93,7 +93,10 @@ This repository follows the same version numbering and release branching schedul
 
 1. Create a new 'release-x.y' branch and push it into Git
 2. Advance the CI configuration in openshift/release to create new lanes for the release branch and move the master along to the next release version number
-   - Example: https://github.com/openshift/release/pull/26172
-3. Pin all midstream branches to the new release branch, and create new midstream branches for the next release
+    - Example: https://github.com/openshift/release/pull/26172
+3. Change the ZTP templates examples label `du-profile: latet` under [gitops-subscriptions/argocd/example](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example) to match the branch release-x.y
```
alosadagrande, line 97: s/latet/latest

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
```diff
@@ -3,11 +3,12 @@ apiVersion: ran.openshift.io/v1
 kind: PolicyGenTemplate
 metadata:
   name: "common"
-  namespace: "ztp-common"
+  namespace: "ztp-common-latest"
```
imiller0, line 6: I think we want to change the name rather than the namespace. This has a couple of impacts to the user:
1. If you change the namespace you will need to copy any hub side templating resources (eg configmap) to the new namespace. This results in duplication or work by the user to remove cluster-abc from one configmap and into another when they change the binding version.
2. TALM and ACM both have restrictions on two policies with the same name but different namespaces if they get bound to the same cluster. With versioning like this it should not happen but I think we should use unique names to stay far away from this limititation. TALM's limitation is in the policy name, ACM's is in the policy-template name.

serngawy: /test e2e-gcp-ovn

README.md
serngawy: done, thanks

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
serngawy: okay, done

imiller0 (approved):
/lgtm

README.md
```diff
@@ -93,7 +93,10 @@ This repository follows the same version numbering and release branching schedul
 
 1. Create a new 'release-x.y' branch and push it into Git
 2. Advance the CI configuration in openshift/release to create new lanes for the release branch and move the master along to the next release version number
-   - Example: https://github.com/openshift/release/pull/26172
-3. Pin all midstream branches to the new release branch, and create new midstream branches for the next release
+    - Example: https://github.com/openshift/release/pull/26172
+3. Change the ZTP templates examples label `du-profile: latest` under [gitops-subscriptions/argocd/example](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example) to match the branch release-x.y
+    - PGT Example: https://github.com/openshift-kni/cnf-features-deploy/pull/1063
```
yliu127, line 98: @serngawy  these examples are changing namespace instead of name - can you please update README?

PR #1065, opened 2022-04-08
Bug 2073439: ztp: Disable podman logging when extracting container data

lack:
As described in https://github.com/containers/podman/issues/13779,
podman by default logs all of a container's stdout in order to retain
info to make 'podman logs' work. But this extra copy of stdout can be
overrun and cause hangs when shipping large amounts of data through
stdout (as in the case when piping out a tar stream).  Changing the
instructions and documentation to add '--log-driver=none' to the podman
commandline disables this extra unneeded logging and prevents hangs.

/cc @imiller0 @serngawy

### DIFF
--- a/ztp/gitops-subscriptions/argocd/README.md
+++ b/ztp/gitops-subscriptions/argocd/README.md
@@ -17,7 +17,7 @@ Create a GIT repository for hosting site configuration data. The ZTP pipeline wi
 2. Export the argocd directory from the ztp-site-generator container image by executing the following commands:
 ```
     $ mkdir -p ./out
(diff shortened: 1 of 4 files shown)


lack:
Tested by running this in a loop to make sure of no hangs, even when extracting the larger `/kustomize` directory structure:

```
i=0; while [[ $((++i)) -lt 1000 ]]; do echo "Starting $i"; podman run --rm --log-driver=none ztp-site-generator:latest extract /kustomize --tar | tar t > k-$i.output; echo "Done $i"; done
```

serngawy (approved):

PR #1066, opened 2022-04-08
Bug 2073498: ztp: Disable podman logging when extracting container data

lack:
As described in https://github.com/containers/podman/issues/13779,
podman by default logs all of a container's stdout in order to retain
info to make 'podman logs' work. But this extra copy of stdout can be
overrun and cause hangs when shipping large amounts of data through
stdout (as in the case when piping out a tar stream).  Changing the
instructions and documentation to add '--log-driver=none' to the podman
commandline disables this extra unneeded logging and prevents hangs.

(cherry picked from commit a2124e068a72c5f21a3ef874a8cb872ba086993d)

/cc @imiller0 @serngawy


lack:
Also tested in a 1000x loop; stress test passed.

imiller0 (approved):
lgtm

PR #1081, opened 2022-04-18
Bug 2057566: ztp: add doc explaining mustonlyhave use when modifying list via PGT

pixelsoccupied:
/cc @lack

### DIFF
--- a/ztp/gitops-subscriptions/argocd/IssueWithCRGeneratedViaPolicy.md
+++ b/ztp/gitops-subscriptions/argocd/IssueWithCRGeneratedViaPolicy.md
@@ -1,0 +1,72 @@
+## Duplicated list entries when modifying Policy
+
+If you're using the default configurations and update a CR's list via PolicyGenTemplate,  you may notice that the "final" CR generated doesn't match the Policy.
+
+e.g After updating `TunedPerformancePatch.yaml`'s `data` list, the final `tuned` CR maybe look like the one below.
+
+```yaml
+[core@cnfdf02 ~]$ oc get tuned -n openshift-cluster-node-tuning-operator performance-patch -o yaml
+apiVersion: tuned.openshift.io/v1
+kind: Tuned
+metadata:
+  name: performance-patch
+  namespace: openshift-cluster-node-tuning-operator
+spec:
+  profile:
+    - data: |
+        ... # rest of data
+
+        [sysctl]
+        kernel.timer_migration=1
+      name: performance-patch
+    - data: |
+        ...  # rest of data
+
+        [sysctl]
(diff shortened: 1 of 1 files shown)


ztp/gitops-subscriptions/argocd/IssueWithCRGeneratedViaPolicy.md
```diff
@@ -0,0 +1,68 @@
+## Issue With CR generated via Policy
+
+If you're using the default configurations and update a CR's list via PGT,  you may notice that the "final" CR generated doesn't match the Policy.
```
imiller0, line 3:
```suggestion
If you're using the default configurations and update a CR's list via PolicyGenTemplate,  you may notice that the "final" CR generated doesn't match the Policy.
```

ztp/gitops-subscriptions/argocd/IssueWithCRGeneratedViaPolicy.md
```diff
@@ -0,0 +1,68 @@
+## Issue With CR generated via Policy
```
imiller0, line 1:
```suggestion
## Duplicated list entries when modifying Policy
```

ztp/gitops-subscriptions/argocd/IssueWithCRGeneratedViaPolicy.md
```diff
@@ -0,0 +1,68 @@
...
+To work-around it just at `mustonlyhave` to the file under your PGT. E.g 
+
+```yaml
+- fileName: TunedPerformancePatch.yaml
+  complianceType: mustonlyhave
+```
+
+## Impact on CPU and system
+
+Adding `mustonlyhave` didn't have any impact on CPU consumption, but it's recommended to use `mustonlyhave` on an as-needed basis. When added keep an eye on the CPU consumption and any possible race conditions.
```
imiller0, line 68: Do we expect that a race condition is possible?

lack: I think "race condition" may be a short-hand for "ACM policy reconciliation loop" if any external entity adds annotations.

imiller0: Can we clarify that line? It isn't a typical race condition. This is the case where there are two controllers ACM policy controller and a second controller both updating the same CR which causes an endless battle over the state of the CR.

pixelsoccupied: Updated with a better expiation + added that helpful PromQL query

lack (approved):

PR #1083, opened 2022-04-20
ztp: update disk partitioning docs to include use of rootDeviceHint

pixelsoccupied:
update docs for disk partitioning to include use of rootDeviceHints and other warnings. also add the source crs file entries  in examples files

/cc @lack

### DIFF
--- a/ztp/gitops-subscriptions/argocd/ImageRegistry.md
+++ b/ztp/gitops-subscriptions/argocd/ImageRegistry.md
@@ -1,24 +1,42 @@
 Installation
 -
-1. Use SiteConfig to generate MachineConfig for disk partitioning. Make sure to modify values in the mc appropriately as it is dependent on the underlying disk.
+1. Use SiteConfig to generate MachineConfig for disk partitioning. Make sure to modify values in the mc appropriately as it is dependent on the underlying disk.   
+
+   It is important use persistent naming for device, especially if you have more than one disk as names like `/dev/sda` and `/dev/sdb` may switch at every reboot. You can use `rootDeviceHints` to choose the bootable device and then use same the device for further partitioning, in this case, for Image registry. More info on persistent naming [here](https://wiki.archlinux.org/title/persistent_block_device_naming#Persistent_naming_methods).
+
+   wwn is used below. 
+
+
 ```yaml
 nodes:
-  - diskPartition:
-    - device: /dev/sda
+  - rootDeviceHints:
+      wwn: "0x62cea7f05c98c2002708a0a22ff480ea"
+    diskPartition:
+    - device: /dev/disk/by-id/wwn-0x62cea7f05c98c2002708a0a22ff480ea # depends on the hardware. can also serial num or device name. recommend using wwn. match with the rootDeviceHint above
       partitions:
        - mount_point: /var/imageregistry
-         size: 102500
-         start: 344844
+         size: 102500 # min value 100gig for image registry
+         start: 344844 # 25000 min value, otherwise root partition is too small. Recommended startMiB (Disk - sizeMiB - some buffer), if startMiB + sizeMiB  exceeds disk size, installation will fail
 ```
    
-3. Use PGT, to apply the following to create the pv and pvc and patch imageregistry config as part of normal day-2 operation
+3. Use PGT, to apply the following to create the pv and pvc and patch imageregistry config as part of normal day-2 operation. Select the appropriate PGT for each source-cr and refer to `wave` doc for more help. Below is as example if you would like to test it at the site level.
    ```yaml
    sourceFiles:
+      # storage class
+      - fileName: StorageClass.yaml
+        policyName: "sc-for-image-registry"
+        metadata:
+          name: image-registry-sc
+          annotations:
+            ran.openshift.io/ztp-deploy-wave: "100" # remove this when moved to the right PGT (site/group/common)
+     # persistent volume claim
      - fileName: StoragePVC.yaml
        policyName: "pvc-for-image-registry"
        metadata:
          name: image-registry-pvc
          namespace: openshift-image-registry
+         annotations:
+            ran.openshift.io/ztp-deploy-wave: "100"  # remove this when moved to the right PGT (site/group/common)
        spec:
          accessModes:
            - ReadWriteMany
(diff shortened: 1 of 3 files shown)


ztp/gitops-subscriptions/argocd/example/policygentemplates/example-sno-site.yaml
```diff
@@ -48,3 +48,34 @@ spec:
...
+        storageClassName: image-registry-sc
+        volumeMode: Filesystem
+    - fileName: ImageRegistryPV.yaml
+      policyName: "config-policy"
+    - fileName: ImageRegistryConfig.yaml
+      policyName: "config-policy"
+      spec:
+        storage:
+          pvc:
+            claim: "image-registry-pvc"
```
Missxiaoguo, line 78: The source crs have different wave numbers means they have dependency between each other so they cannot be applied in the same policy. Should revisit the location for these CRs based on their waves.

pixelsoccupied: I've put them in the same file because the configurations only apply at the SNO level (i.e as opposed to common/group). What should I change in order to keep them in this file as they need to be bunched up together...wave numbers maybe? Or are you saying split them apart to different files?

pixelsoccupied: I tried with the same PGT above and installation completed successfully. Was an error expected somewhere?

Missxiaoguo: > I tried with the same PGT above and installation completed successfully. Was an error expected somewhere?

You probably are using ZTP 4.9? In ZTP 4.10, yes, the Argocd should throw out an error like 
"ran.openshift.io/ztp-deploy-wave annotation in Resource StorageClass.yaml (wave 10) doesn't match with Policy example-sno-config-policy (wave 100)"

Missxiaoguo: > I've put them in the same file because the configurations only apply at the SNO level (i.e as opposed to common/group). What should I change in order to keep them in this file as they need to be bunched up together...wave numbers maybe? Or are you saying split them apart to different files?

I am not sure what's expected order to apply these resources. But ideally if the waves they currently have are expected(ie. StorageClass: 10, StoragePVC:10, ImageRegistryPV:2, ImageRegistryConfig:11), then StorageClass and StoragePVC should go to group PGT and ImageRegistryConfig goes to group PGT as well but in a different policy other than config-policy and ImageRegistryPV goes to common PGT.

pixelsoccupied: Ah yep see the ArgoCD error. `/tmp/kust-plugin-config-009758476: ran.openshift.io/ztp-deploy-wave annotation in Resource StoragePVC.yaml (wave 10) doesn't match with Policy cnfdf01-config-policy (wave 100) `....Though installation went through? 

Sorry I'm a little lost as to how wave is working. I see the "100" in the final`Policy` but not on the PGT? Can you tell where this is coming from? And does that mean all the source-crs need to be 100 for it to be valid policy? 

I don't really need to set them in any order but they all need be at the node level. Meaning the 4 CRs need to applied only when there's a disk partition done via SiteConfig. Otherwise there would random crashes everywhere. 

Do you think it makes sense to override the annotation in the PGT and set them 100(?)

pixelsoccupied:
```
    - fileName: StoragePVC.yaml
      policyName: "config-policy"
      metadata:
        name: image-registry-pvc
        namespace: openshift-image-registry
        annotations:
      spec:
        accessModes:
          - ReadWriteMany
        resources:
          requests:
            storage: 100Gi
        storageClassName: image-registry-sc
        volumeMode: Filesystem
    - fileName: ImageRegistryPV.yaml
      policyName: "config-policy"
      metadata:
        annotations:
    - fileName: StorageClass.yaml
      policyName: "config-policy"
      metadata:
        name: image-registry-sc
        annotations:
    - fileName: ImageRegistryConfig.yaml
      policyName: "config-policy"
      metadata:
        annotations:
      spec:
        managementState: Managed
        storage:
          pvc:
            claim: "image-registry-pvc"
```

Can I update the example with the one above? This got rid of the error.

Missxiaoguo: These are the references
[1] https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/policygenerator/README.md#policy-waves
[2] https://deploy-preview-43890--osdocs.netlify.app/openshift-enterprise/latest/scalability_and_performance/ztp-deploying-disconnected.html#ztp-talo-integration_ztp-deploying-disconnected (See waves)

From the registryconfig yaml, the comment mentions registryconfig depends on storagePVC to be configured so wave 11 is set. Having them at the node level means user will be required to add these CRs in each site PGT that needs a local registry.

lack: Just because these CRs correspond to a specific node configuration doesn't necessarily mean they need to be in the site-specific PGT, though.  For example, if the configuration is identical for every SNO, it's perfectly valid to put them in the SNO group.

lack: At a higher level, I think maybe it's better to have the relevant sections commented-out in our examples.

The SiteConfig example would have a comment block mentioning that if you want ImageRegistry, you need to uncomment the disk partition and the relevant (group and/or site TBD) PGT example.

Likewise the PGT example would have a comment block mentioning that if you want ImageRegistry you need to uncomment this section and also make sure your SiteConfig has the proper disk partition set up.

How does that sound?

pixelsoccupied: Commenting out sounds good. Users can always refer to the doc (which I can point to from here) for more info. Will make the changes now.

lack (approved):

PR #1090, opened 2022-04-25
Bug 2066701: ztp: Match PerformanceProfile name with informDuValidator

sudomakeinstall2:
To avoid mistakes from the end user we hardcode the PerformanceProfile
name to match the value of informDuValidator

### DIFF
--- a/ztp/source-crs/TunedPerformancePatch.yaml
+++ b/ztp/source-crs/TunedPerformancePatch.yaml
@@ -8,12 +8,14 @@ metadata:
 spec:
   profile:
     - name: performance-patch
-      # The 'include' line must match the associated PerformanceProfile name
-      # And the cmdline_crash CPU set must match the 'isolated' set in the associated PerformanceProfile
+      # The 'include' line must match the associated PerformanceProfile name, following below pattern
+      # include=openshift-node-performance-${PerformanceProfile.metadata.name}
+
+      # The 'cmdline_crash' CPU set must match the 'isolated' set in the associated PerformanceProfile
       data: |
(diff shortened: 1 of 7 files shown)


ztp/source-crs/TunedPerformancePatch.yaml
```diff
@@ -13,7 +13,7 @@ spec:
       data: |
```
Missxiaoguo, line 13: Can we explain a bit more in line12 to indicate the part that should match the PerformanceProfile name?

lack: Agreed.  A comment explaining that if you change the PerformanceProfile name you need to change this to match is probably important both here and in `PerformanceProfile.yaml`

lack (changes requested):
One thing I don't see is a change to our example policies to match - can we update them too, please?

ztp/source-crs/TunedPerformancePatch.yaml
sudomakeinstall2: Added comment in both files.

sudomakeinstall2:
> One thing I don't see is a change to our example policies to match - can we update them too, please?


@lack 
Of course, do you mean removing 
```
      metadata:
        name: openshift-node-performance-profile
```
from examples?

lack:
Exactly :)

sudomakeinstall2:
Done, I also added the same comment for `informDuValidator`. I would be happy if you review that also.

lack (approved):

PR #1100, opened 2022-05-03
Add netcat to cnftests image

zeeke:
`nc` comes in handy when implement tests for multi-networkpolicy, in order to check if
a host/port is reachable from specific pods.

I'm filing this PR because images should be already available on quay.io prior to filing the PR that uses them.


mmirecki:
lgrm

PR #1105, opened 2022-05-09
WIP Add script to analyze reboots

rcarrillocruz:
@lack @tliu2021 @alosadagrande I don't know where to put this, if you have better ideas let me know.
Also, please comment things as you see fit, I figure it's better to drive it in a PR than on a gdoc .

### DIFF
--- a/ztp/utils/openshift-reboot-analysis.sh
+++ b/ztp/utils/openshift-reboot-analysis.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+NODE_REBOOTS=$(last -x shutdown reboot |grep reboot | wc -l)
+NODE_SHUTDOWNS=$(last -x shutdown reboot |grep shutdown | wc -l)
+NODE_UNGRACEFUL_REBOOTS=$(($NODE_REBOOTS - $NODE_SHUTDOWNS))
+SRIOV_REBOOTS=$(oc -n openshift-sriov-network-operator logs -l app=sriov-network-config-daemon -c sriov-network-config-daemon | grep "reqReboot true")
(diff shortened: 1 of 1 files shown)


tliu2021 (commented):
As the rendered MachineConfig(MC) object contains all previously applied MC objects, it would be hard for the user to find out which MC triggered the reboot. It would be nice to show the diff between each reboot.

rcarrillocruz:
> As the rendered MachineConfig(MC) object contains all previously applied MC objects, it would be hard for the user to find out which MC triggered the reboot. It would be nice to show the diff between each reboot.

Can you expand on showing diff? Which two things should be diffed, the one shown the logs against...

tliu2021:
@rcarrillocruz For example, find the first boot rendered MC object “rendered-master-aed6f420dea74d85a71e857fc5d280f7” from the following logs
Jan 12 14:53:46 cnfocto1.ptp.lab.eng.bos.redhat.com root[3981]: machine-config-daemon[3796]: initiating reboot: Completing firstboot provisioning to rendered-master-aed6f420dea74d85a71e857fc5d280f7

Then get the second reboot rendered MC objects “rendered-master-1dd38629a8e669a5bf61f1509bab354b” from the logs shown below:
Jan 12 15:29:32 cnfocto1.ptp.lab.eng.bos.redhat.com root[145716]: machine-config-daemon[12144]: initiating reboot: Node will reboot into config rendered-master-1dd38629a8e669a5bf61f1509bab354b

The difference between the two files would show the new configurations that triggered the second reboot.

rcarrillocruz:
So what you are saying is getting the history of all the reboot rendered MCs and diff them consecutively. Ok, thanks.

PR #1106, opened 2022-05-10
ztp: Redo the ContainerFile for ztp-site-generator

nishant-parekh:
This removes the reliance on an external tarball so the Prow CI image
builder can build this for us.

It also simplifies the build process a lot, relying on the existing
makefiles in our source control to set up the kustomize plugin directory
structure rather than duplicating it, and removing a lot of extra
copying.

ztp: fixed merge conflicts from cherry-pick

### DIFF
--- a/ztp/resource-generator/tools/patchImageReference.sh
+++ b/ztp/resource-generator/tools/patchImageReference.sh
@@ -7,7 +7,7 @@ set -e
 
 BASEDIR=$1
 REPLACEMENT_IMAGE=$2
-UPSTREAM_IMAGE="quay.io/redhat_emp1/ztp-site-generator:latest"
+UPSTREAM_IMAGE="quay.io/openshift-kni/ztp-site-generator:4.10.0"
 
 if [[ $1 == "-h" || $1 == "--help" ]]; then
     echo "Usage:"
--- a/ztp/gitops-subscriptions/argocd/README.md
+++ b/ztp/gitops-subscriptions/argocd/README.md
@@ -8,7 +8,7 @@ The GitOps ZTP infrastructure relies on the ztp-site-generator container to prov
 
 ## Obtaining pre-built image
 ```
-    $ podman pull quay.io/redhat_emp1/ztp-site-generator:latest
+    $ podman pull quay.io/openshift-kni/ztp-site-generator:4.10.0
 ```
 
 ### Preparation of ZTP GIT repository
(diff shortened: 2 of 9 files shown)


lack:
You need to also backport d5167f0ca9a9ba6de5aa9d55ed9ae2b174933471 in order to not lose the image patching logic, and a2124e068a72c5f21a3ef874a8cb872ba086993d to fix a 'make export` issue in the Makefile.

lack:
Analyzing the difference between the new and old container image contents:
- Old container has `/tmp/ztp-temp/...` but the new container does not.
  - This is actually good, this temporary directory is not needed and inflates the image size.
- New container has `/home/ztp/source-crs/README.md` and `/home/ztp/source-crs/extra-manifest/*`
  - This is benign

Everything else looks pretty good!

lack:
> we do need the Image reference change in this PR correct?

Either here or very soon into this branch, yes.

ztp/resource-generator/tools/patchImageReference.sh
```diff
@@ -7,7 +7,7 @@ set -e
 
 BASEDIR=$1
 REPLACEMENT_IMAGE=$2
-UPSTREAM_IMAGE="quay.io/redhat_emp1/ztp-site-generator:latest"
+UPSTREAM_IMAGE="quay.io/openshift-kni/ztp-site-generator:4.10.0"
```
nishant-parekh, line 10: just want to make sure we should refernce the 4.10.0 tag here instead of latest right?

lack: Is it `4.10.0` or `v4.10.0`?

nishant-parekh: I looked up the existing tags here:
https://quay.io/repository/openshift-kni/ztp-site-generator?tab=tags and seems like it will 4.10.0 instead of v4.10.0.. as that is the format for the 4.11.0 tags.

lack: Just checked!  `4.10.0` is correct.

lack: Aha - But this needs to match the string that's currently in ztp/gitops-subscriptions/argocd/deployment/argocd-openshift-gitops-patch.json 

Right now it's `quay.io/redhat_emp1/ztp-site-generator:latest`, but we should probably edit THAT to match the new upstream to be `quay.io/openshift-kni/ztp-site-generator:4.10.0`

So do a quick search for `quay.io/redhat_emp1/ztp-site-generator:latest` under ztp/ and replace it with `quay.io/openshift-kni/ztp-site-generator:4.10.0` in this PR, and I think we're good to go.

lack:
/unhold

Testing looks good!

lack:
/override ci/prow/images

(The image we're talking about here doesn't even build at this stage)

PR #1109, opened 2022-05-12
cnf-tests: update the documentation about image pull

SchSeba:
We introduce a new container to validate the secure boot status on the nodes.

That pod runs in the openshift-sriov-network operator namespace so there is a need to allow pulling disconnected images from the cnftest namespace to the sriov one.


SchSeba:
doc only change

/override ci/prow/e2e-gcp-ovn

PR #1112, opened 2022-05-16
ztp: add an entrypoint for single-site resources generation

Missxiaoguo:
The new entrypoint generator is added to provide the ability to generate policy or non-policy wrapped configurationCRs based on PolicyGenTemplate(s) and a full set of installationCRs or extra manifests resources based on SiteConfig(s).


ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -20,6 +20,8 @@ const Standard = "standard"
 const Master = "master"
 const ZtpAnnotation = "ran.openshift.io/ztp-gitops-generated"
 const ZtpAnnotationDefaultValue = "{}"
+const UnsetStringValue = "__unset_value__"
```
lack, line 23: I've never been a big fan of this convention... I'd rather use a `*string` and have `nil` represent "not set"...

That said, I know we use this elsewhere in the codebase, so I suppose it can stay :(

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,417 @@
...
+
+printenv MOUNT_DIR &> /dev/null
+if [[ $? -ne 0 ]]; then
+  # default dest dir in the container that is mounted from the host
+  MOUNT_DIR="/home/ztp_resources"
+fi
+
+function help {
+    cat >&2 <<EOF
+ztp-site-generator single node resources generator
```
lack, line 23: Maybe this should be "single site" instead of "single node" since this will work for multi-node clusters as well!

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,417 @@
...
+  podman run -it --rm -v \$host_dir:\$container_dir:Z,U -e MOUNT_DIR=\$container_dir \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} siteconfigs/mycluster.yaml out/generated_installCRs
+  podman run -it --rm -v \$host_dir:\$container_dir:Z,U -e MOUNT_DIR=\$container_dir \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} siteconfigs out/generated_installCRs
+
+  # generate extra manifests only with -e option
+  podman run -it --rm -v \$host_dir:\$container_dir:Z,U \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} -e true \$container_dir/siteconfigs \$container_dir/out/generated_installCRs
+
+  Note: With podman, to grant the user running in the container permission to read and write the folder mounted in from the host,
+        add :U volume option to the "-v" flag or run the container as root with the flag "-u root"
+
+Options:
```
lack, line 88: I would generally expect the options to come before the examples.

lack (commented):
Looking good! Just a couple minor suggestions.

ztp/siteconfig-generator/siteConfig/siteConfig.go
Missxiaoguo: flag.String needs to set a default value and it's a string type which cannot be the value of a nil pointer. I'd rather to keep it as "__unset_value__" to differentiate between "flag is set with an empty string" and "flag is unset" if needed

ztp/resource-generator/Containerfile
```diff
@@ -22,7 +22,9 @@ USER root
 ENV BUILDER_ZTP=/go/src/cnf-features-deploy/ztp
 ENV ZTP_HOME=/home/ztp
 # Install utilities needed by the entrypoint scripts
-RUN microdnf install tar
+RUN microdnf install tar util-linux wget
```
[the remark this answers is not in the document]
Missxiaoguo: Yep, updated

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,420 @@
...
+  -e, --extra-manifest-only=false      If true, generate extra manifests only
+
+Examples:
+  # If you have SiteConfig(s) stored in /home/ztp_resources/siteconfigs on the host,
+  host_dir=/home/ztp_resources
+  container_dir=/home/ztp_resources
+
+  # read from \$container_dir/siteconfigs or \$container_dir/siteconfigs/mycluster.yaml,
+  # the generated resources are outputted to \$container_dir/out/generated_installCRs which are accessible from \$host_dir/out/generated_installCRs on the host
+  podman run -it --rm -v \$host_dir:\$container_dir:Z,U \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} \$container_dir/siteconfigs \$container_dir/out/generated_installCRs
```
lack, line 73: In all of these examples, does it really need `-it`?  You need `-i` to send stdin to the container, and `-t` to allocate a ptty; I think in all of these unattended modes of operation, that's probably not needed?

Missxiaoguo: You are absolutely right! `-it` is not needed. I used the interactive mode to test at the begining and didn't think about to remove it in the examples. 

I have removed it and updated the note for that use

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -54,10 +54,31 @@ func (scbuilder *SiteConfigBuilder) Build(siteConfigTemp SiteConfig) (map[string
...
+				if err != nil {
+					log.Printf("Error: could not unmarshal string:(%s) (%s)\n", manifest.(string), err)
+				}
+				clusterCRs = append(clusterCRs, dataResult)
+			}
+		} else {
+			clusterCRs, err = scbuilder.getClusterCRs(id, siteConfigTemp)
+			if err != nil {
+				return clustersCRs, err
+			}
```
lack, line 81: I'd like to think about refactoring this a bit - If the `scbuilder.getClusterCRs(...)` function took the dataMap of extra manifests as input to the call, this logic (and the logic in the other calls) would be a lot cleaner.

I don't think we necessarily need to do it in this PR, but what do you thank of that approach in general?

lack:
This is looking good!  Do we have any unit tests around the new functionality?  Or maybe new functional tests to exercise the functionality?

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+  * NMStateConfig
+  * KlusterletAddonConfig
+  * ManagedCluster
+  * InfraEnv
+  * BareMetalHost
+  * HostFirmwareSettings
+  * extra-manifest configurations
+
+Environment variables:
+    \$MOUNT_DIR          set the dest path of the volume mounted in from the host (default "/home/ztp_resources"). It's the base location for storing the input and output resources.
```
lack, line 54: Rather than hard-coding the default here and above, it's probably better for this text to be generated by the default value in code above.

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+The DEST_DIR should be a directory and if DEST_DIR is not sepecified, by default the generated resources will be outputted to \$MOUNT_DIR/out/generated_installCRs.
+
+Usage: ${PROG} ${SUB_COMMAND} SRC_PATH [DEST_DIR] [options]
+
+Options:
+  -h, --help                           help
+  -e, --extra-manifest-only=false      If true, generate extra manifests only
+
+Examples:
+  # If you have SiteConfig(s) stored in /home/ztp_resources/siteconfigs on the host,
```
lack, line 67: I think if I had my SiteConfigs in a directory like that, I'd want to just mount that directly into the container and not have to specify a sub-directory.  For example:

```
podman run -v /home/ztp_resources/siteconfigs:/resources -v /home/ztp_resources/siteconfigs/outputs:/output generator siteconfig -e
```

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
+#!/usr/bin/env bash
```
lack, line 1: This is a big enough script that we should consider some kind of automated test for it.  I don't know if 'unit test' is quite the right thing, but maybe some kind of integration test that runs as part of the CR (and can be run locally) to check the generator functionality?

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+
+  podman run --rm -v \$host_dir:\$container_dir:Z,U \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} \$container_dir/policygentemplates
+  podman run --rm -v \$host_dir:\$container_dir:Z,U \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} policygentemplates out/generated_configCRs
+  podman run --rm -v \$host_dir:\$container_dir:Z,U \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} policygentemplates/mycluster.yaml out/generated_configCRs
+
+  # when DEST_DIR is not given, or relative SRC_PATH and DEST_DIR are given but \$container_dir is other than /home/ztp_resources,
+  # the container environment variable MOUNT_DIR needs to be given with podman flag "-e"
+  podman run --rm -v \$host_dir:\$container_dir:Z,U -e MOUNT_DIR=\$container_dir \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} \$container_dir/policygentemplates
+  podman run --rm -v \$host_dir:\$container_dir:Z,U -e MOUNT_DIR=\$container_dir \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} policygentemplates/mycluster.yaml out/generated_configCRs
+  podman run --rm -v \$host_dir:\$container_dir:Z,U -e MOUNT_DIR=\$container_dir \$THIS_CONTAINER ${PROG} ${SUB_COMMAND} policygentemplates out/generated_configCRs
```
imiller0, line 135: Same comment on reducing options for simplicity.

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+        exit 1
+    fi
+
+    eval set -- "$opts"
+    while true;do
+        case "$1" in
+            -h|--help)      installCRs_help
+                            exit 0
+                            ;;
+            -e|--ems-only)  EMS_ONLY="$2"
```
imiller0, line 163: If the user uses the short form do they need to use it as "-e true"? Typically a short form of a boolean can be stand-alone where "-e" means "--ems-only true". Not a huge issue but may confuse some users if they do "-e mySiteConfig.yaml" and it complains that there is no input file given (since the -e swallowed the mySiteConfig.yaml)

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+        exit 1
+    fi
+
+    eval set -- "$opts"
+    while true;do
+        case "$1" in
+            -h|--help)                configurationCRs_help
+                                      exit 0
+                                      ;;
+            -w|--wrap-in-policy)      WRAP_IN_POLICY="$2"
```
imiller0, line 231: Same "short form" comment as above

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+                            ;;
+        esac
+    done
+
+    if [ ${EMS_ONLY} != "false" ] && [ ${EMS_ONLY} != "true" ]; then
+        echo "Invalid value ${EMS_ONLY} for \"--ems-only\" flag. Expected value is true or false." >&2
+        exit 1
+    fi
+
+    outDir="/out/generated_installCRs"
```
imiller0, line 181: Maybe add a comment explaining that verify_arguments uses outDir to set destDir. The setting of destDir as a side-effect of the function is hard to track when reading through.

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+                                      ;;
+        esac
+    done
+
+    if [ ${WRAP_IN_POLICY} != "false" ] && [ ${WRAP_IN_POLICY} != "true" ]; then
+        echo "Invalid value ${WRAP_IN_POLICY} for \"--wrap-in-policy\" flag. Expected value is true or false." >&2
+        exit 1
+    fi
+
+    outDir="/out/generated_configCRs"
```
imiller0, line 248: Same note on adding a comment

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+EMS_ONLY="false"
+WRAP_IN_POLICY="true"
+
+ztpTmpDir="/home/ztp/tmp"
+basePath="/kustomize/plugin/ran.openshift.io/v1"
+policyGenPath=${basePath}/policygentemplate
+siteConfigPath=${basePath}/siteconfig
+
+printenv MOUNT_DIR &> /dev/null
+if [[ $? -ne 0 ]]; then
```
imiller0, line 16: Another way to do this is to use variable substitution:
`MOUNT_DIR=${MOUNT_DIR:-/home/ztp_resources}`

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+        echo "Error: $srcPath is not a file or a directory" >&2
+        exit 1
+    else
+        echo "Error: $srcPath is not found" >&2
+        exit 1
+    fi
+}
+
+# Check if a directory is mounted
+function is_dir_mounted {
```
imiller0, line 327: It looks like this is no longer used?

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+    kind=$(yq eval -o=y '.kind' $file | head -1)
+    if [ $kind != $expected_cr ]; then
+        echo "${file} does not have ${expected_cr} CR. Skipping..."
+        return 1
+    fi
+    return 0
+}
+
+# Copy user defined additional resources in SiteConfig to /home/ztp/tmp
+# if they are relative paths to the parent directory of SiteConfig file
```
imiller0, line 353: Why do these files need to be copied? Does this mean that SiteConfig is not using the correct search path for finding the files? It would seem like it should always be able to find the files relative to the SiteConfig that it is processing.

Missxiaoguo: Based on our SiteConfig code, in theory, the directory of additional manifests or bios path is the relative path to the directory where the siteconfig is called.
With kustomize, it should be relative to the kustomization.yaml
For this usecase, we need to either go to the siteconfigs path that mounted from host and execute siteconfig there or copy the resources to another directory we want to execute from. I couldn't recall why I didn't choose to go to the siteconfigs directory at the first place... but looks like it's no disavantage and much easier than copying

Also may be we want to restrict the extra manifest path in the SiteConfig CR to be relative to the siteconfig yaml instead of the running path? @imiller0 @lack

lack: I think I like the no-copy idea, too.

As for restricting extra-manifest, that may be a nice thing to do in order to prevent mistakes, sure.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
Missxiaoguo: Updated

ztp/resource-generator/entrypoints/generator
```diff
@@ -0,0 +1,425 @@
...
+
+ztpTmpDir="/home/ztp/tmp"
+basePath="/kustomize/plugin/ran.openshift.io/v1"
+policyGenPath=${basePath}/policygentemplate
+siteConfigPath=${basePath}/siteconfig
+
+printenv MOUNT_DIR &> /dev/null
+if [[ $? -ne 0 ]]; then
+  # default dest dir in the container that is mounted from the host
+  MOUNT_DIR="/home/ztp_resources"
```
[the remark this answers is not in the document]
Missxiaoguo: Updated input to `/resources` but kept the default output differently for installCR and configCR

ztp/resource-generator/entrypoints/generator
lack: I see it called on line 316&317 now

lack (approved):

PR #1115, opened 2022-05-17
[release-4.10] Bug 2064585: cnf-tests: remove mainaffinity option

Tal-or:
The cyclictest binary which is called by the runner doesn't have the support for the mainaffinity flag in its current release.

Hence, we should remove the arg entirely

### DIFF
--- a/cnf-tests/pod-utils/cyclictest-runner/main.go
+++ b/cnf-tests/pod-utils/cyclictest-runner/main.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"k8s.io/klog"
-	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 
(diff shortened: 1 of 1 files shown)


fromanirh:
so do we want a 4.10-specific fix, right? IOW, we want the `mainaffinity` feature only on 4.11 and onwards, right?

fromanirh:
/lgtm
I see a plain revert was not really easier than this change. I crosschecked the revert and looks OK.

Tal-or:
> /lgtm I see a plain revert was not really easier than this change. I crosschecked the revert and looks OK.

We couldn't do plain revet because it contained a small fix to another function which both oslat and cyclictest was dependent on.
We don't need to for cyclictest anymore but oslat still has it.

PR #1123, opened 2022-05-30
cnf-tests: Fix dpdk tests

SchSeba:
This change is needed because `-w` option in dpdk-testpmd was removed.
instead we need to use the new option `-a`


SchSeba:
backport for https://github.com/openshift-kni/cnf-features-deploy/pull/1121

PR #1133, opened 2022-06-08
infra: use a specific podman version to build the index images

SchSeba:


yuvalk:
IMHO, this fix is not enough to prevent issues in the future, we're still using non-specific, latest podman here
https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/setup-build-index-image.sh#L46

PR #1136, opened 2022-06-09
Bug 2087219: Fix image registry partition template on multi node cluster

rcarrillocruz:
If the SiteConfig contains multiple nodes per cluster, the siteconfig
generator fails as it tries to create the same manifest for each one
of the nodes.
Since the config for the registry should be the same for all the nodes
in the cluster, this patch removes the range on all clusters in the
template and it only consumes the SiteConfig data of the first node.


alosadagrande:
> Since the config for the registry should be the same for all the nodes
> in the cluster, this patch removes the range on all clusters in the
> template and it only consumes the SiteConfig data of the first node.

If it only consumes the data from the first node and the configuration will be the same for all of them, does not make more sense to be a clusters spec instead of a node spec inside the siteConfig?

rcarrillocruz:
Per slack convo, let's sync up offline in the lab, compare SNO vs compact cluster registry deployments to see how it would affect in case nodes go on and off.

alosadagrande:
I am having problems to execute the provisioning with the fix. ArgoCD plugin is complaining:

`rpc error: code = Unknown desc = `kustomize build .demos/ztp-policygen/site-configs --enable-alpha-plugins` failed exit status 1: 2022/06/17 10:29:49 Error: could not unmarshal string:({{- range .Data.Nodes -}} {{ if .DiskPartition -}} {{- "# Automatically generated by extra-manifests-builder" }} {{ "# Do not make changes directly." }} apiVersion: machineconfiguration.openshift.io/v1 kind: MachineConfig metadata: name: 98-partition-{{ $.Role }} labels: machineconfiguration.openshift.io/role: {{ $.Role }} spec: config: ignition: version: 3.2.0 storage: disks: {{- range .DiskPartition }} - device: {{ .Device }} wipeTable: false partitions: {{- range .Partitions }} - sizeMiB: {{ .Size }} startMiB: {{ .Start }} label: {{ .Label }} {{- end }} {{- end }} filesystems: {{- range .DiskPartition }} {{- range .Partitions }} - path: {{ .MountPoint }} device: /dev/disk/by-partlabel/{{ .Label }} format: {{ .FileSystemFormat }} {{- end }} {{- end }} systemd: units: {{- range .DiskPartition }} {{- range .Partitions }} - name: {{ .MountFileName }} enabled: true contents: | [Unit] Before=local-fs.target [Mount] Where={{ .MountPoint }} What=/dev/disk/by-partlabel/{{ .Label }} [Install] WantedBy=local-fs.target {{- end }} {{- end }} {{ end -}} {{ end -}} ) (yaml: did not find expected node content) 2022/06/17 10:29:49 Error could not create extra-manifest cnf23.compact-clus3a-t5g-lab/extra-manifests yaml: did not find expected node content 2022/06/17 10:29:49 Error: could not build the entire SiteConfig defined by /tmp/kust-plugin-config-2792808213: yaml: did not find expected node content Error: failure in plugin configured via /tmp/kust-plugin-config-2792808213; exit status 1: exit status 1
`

rcarrillocruz:
@serngawy @pixelsoccupied I chatted with @alosadagrande and he tells me you chatted the past few days about this bug ( I was on PTO ) and there has been some discussion about setting registry info at cluster level.
Can you please comment ?

alosadagrande:
Hi, my point is that the disk partition must be independent of the image registry configuration since it can be used for the TALO recovery folder. for instance or any other configuration the user wants. One thing is the device layout and another is using this disk partition to set up a persistent registry configuration or any other config. Specifically, I mean:

- https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/source-crs/extra-manifest/image-registry-partition-mc.yaml.tmpl -> the name of this extra manifest which is responsible for creating a partition on a device, should not be named as 98-var-imageregistry-partition-{{ $.Role } neither the file name  image-registry-partition.mc.yaml.tmpl since it can be used for multiple purposes.
- Next, if in a compact cluster we are going to only take into account the diskpartition config of the first node and all the disk partition layout is going to be the same then it makes more sense to me for this feature to be a cluster setting, e.g, move diskpartition to cluster spec.

serngawy:
I believe the PR mixing two issues, the MC name in the [image-registry-partition-mc.yaml.tmpl](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/source-crs/extra-manifest/image-registry-partition-mc.yaml.tmpl) and the disk partition config. The MC uses the Node->diskPartition to set its configuration. Are we assuming that all the Nodes in the cluster will have same disk partition config ? (not sure if that a right assumption) . If **Yes** then we can level up the registry config to cluster level. If **No**, the registry config needs to be created per node (MC for each node however will need different role for each node than worker) **OR** can we list all the nodes disk partition config in the same MC then it will work with the one exist in the node ?

alosadagrande:
Taking into account that this problem is targeting only 3NC (they are the currently supported configuration for a multinode cluster with ZTP) I would assume the disk layout/disk partition is the same in all 3 nodes. At least on telco, we encourage users to run the same hw and they usually do. 

As you mentioned in the case of different partition layouts we would require an MCP for each node so we can assign its own specific machineConfig, which does not escalate. 

I would encourage to include diskPartition at the cluster level and document appropriately.  Also the template should be renamed since the name is confusing, it is not only about image registry setup.

PR #1143, opened 2022-06-15
Bug 2051443: cnf-tests: verify minimal CPUs requirement for oslat test

Tal-or:
We should handle the amount of minimal CPUs for oslat test more carefully.
When HT disabled oslat needs 2 CPUs.

But when HT enabled oslat needs at least 4 CPUs:
One CPU should be used to run oslat's main thread.
The second is the sibling of the first one, which should be excluded from the list of the tested CPUs,
Because it might cause to false spikes (noisy-neighbor issue).
The third one is the actual CPU to be tested, but due to SMT alignment restriction we need its sibling too.
Four in total.

### DIFF
--- a/cnf-tests/pod-utils/oslat-runner/main.go
+++ b/cnf-tests/pod-utils/oslat-runner/main.go
@@ -32,10 +32,24 @@ func main() {
 	}
 
 	mainThreadCPUs := selfCPUs.ToSlice()[0]
+	klog.Info("oslat main thread cpu: %d", mainThreadCPUs)
+
 	siblings, err := node.GetCPUSiblings(mainThreadCPUs)
 	if err != nil {
 		klog.Fatalf("failed to get main thread CPU siblings: %v", err)
 	}
+	klog.Info("oslat main thread's cpu siblings: %v", siblings)
+
+	// siblings > 1 means Hyper-threading enabled
+	if len(siblings) > 1 && selfCPUs.Size() == 2 {
+		// one CPU should be used to run oslat's main thread.
+		// the second is the sibling of the first one, which should be excluded from the list of the tested CPUs,
+		// because it might cause to false spikes (noisy-neighbor issue).
+		// the third one is the actual CPU to be tested, but due to SMT alignment restriction we need its sibling too.
+		// four in total.
+		klog.Fatalf("when hyper-threading enabled oslat pod requires at least 4 CPUs")
+	}
+
 	cpusForLatencyTest := selfCPUs.Difference(cpuset.NewCPUSet(siblings...))
 	mainThreadCPUSet := cpuset.NewCPUSet(mainThreadCPUs)
 


cnf-tests/pod-utils/oslat-runner/main.go
```diff
@@ -32,10 +32,24 @@ func main() {
...
 	mainThreadCPUs := selfCPUs.ToSlice()[0]
+	klog.Info("oslat main thread cpu: %d", mainThreadCPUs)
+
 	siblings, err := node.GetCPUSiblings(mainThreadCPUs)
 	if err != nil {
 		klog.Fatalf("failed to get main thread CPU siblings: %v", err)
 	}
+	klog.Info("oslat main thread's cpu siblings: %v", siblings)
+
+	// siblings > 0 means Hyper-threading enabled
```
shajmakh, line 43: can you elaborate why?

Tal-or: This is how Hyper-threading works, one physical core is treated as two logical.
two logical cores of the same physical one, are treated as siblings.
https://stackoverflow.com/questions/7274585/linux-find-out-hyper-threaded-core-id

fromanirh (approved):
/approve

shajmakh:
@Tal-or looks good, thanks for the change. I expect there should be a following change on PAO/NTO side in case CPUs are less than 4:
https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/4_latency/latency.go#L78

Tal-or:
> @Tal-or looks good, thanks for the change. I expect there should be a following change on PAO/NTO side in case CPUs are less than 4: https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/4_latency/latency.go#L78

Why? The check is system-dependent so it must be done inside the test pod.
The code you are referring to prior to the test pod creation.

shajmakh:
> > @Tal-or looks good, thanks for the change. I expect there should be a following change on PAO/NTO side in case CPUs are less than 4: https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/4_latency/latency.go#L78
> 
> Why? The check is system-dependent so it must be done inside the test pod which runs on the target system. The code you are referring to prior to the test pod creation.

because customers runs latency tools tests via the cnf-tests, and if they run them on HT env with less than 4 CPUs the test will fail to run properly because the test pod will not start running as it hits SMT error. Thus such cases should be skipped with a relevant message to the user.  Similarly to https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/4_latency/latency.go#L96

Tal-or:
> > > @Tal-or looks good, thanks for the change. I expect there should be a following change on PAO/NTO side in case CPUs are less than 4: https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/4_latency/latency.go#L78
> > 
> > 
> > Why? The check is system-dependent so it must be done inside the test pod which runs on the target system. The code you are referring to prior to the test pod creation.
> 
> because customers runs latency tools tests via the cnf-tests, and if they run them on HT env with less than 4 CPUs the test will fail to run properly because the test pod will not start running as it hits SMT error. Thus such cases should be skipped with a relevant message to the user. Similarly to https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/4_latency/latency.go#L96

This check runs in ```BeforeEach``` we'll hit that anyway, aren't we?

PR #1144, opened 2022-06-15
Remove disable ironic inspection annotation

tliu2021:
This update removes “inspect.metal3.io: disabled” annotation from ZTP generated BMH.
In ACM 2.5 or earlier, BMAC in assisted-service sets this annotation if it is not set.
In ACM 2.6, BMAC decides whether to set this annotation based on the ConvergedFlow
feature flag.


Missxiaoguo:
Looking good!

PR #1146, opened 2022-06-17
Bug 2098156: ptp offset metrics for master count should be one

aneeshkp:
Recent change to check for master offset narrows down to find only one metrics for master offset.
The fix will check count of master offset to one
signed-off-by: Aneesh Puttur <aneeshputtur@gmail.com>


SchSeba:
The changes in the PR are not related to the failing test 

/override ci/prow/e2e-gcp-ovn

PR #1150, opened 2022-06-22
infra: Improve feature-wait timeout

sabinaaledort:
The current timeout can result a wait of ~5 hours if one of the validation tests is failing.
In a working run on BM the max time is takes for all the CNF features to be ready is 50 minutes.
We can reduce the timeout to 90 minutes and allow setting the timeout variable.

### DIFF
--- a/hack/feature-wait.sh
+++ b/hack/feature-wait.sh
@@ -12,14 +12,16 @@ if [ "$FEATURES" == "" ]; then
 	exit 1
 fi
 
-ATTEMPTS=0
-MAX_ATTEMPTS=200
 all_ready=false
 export TEST_SUITES="validationsuite"
 export FAIL_FAST="-ginkgo.failFast"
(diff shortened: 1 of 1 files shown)


hack/feature-wait.sh
```diff
@@ -12,15 +12,20 @@ if [ "$FEATURES" == "" ]; then
 	exit 1
 fi
 
-ATTEMPTS=0
-MAX_ATTEMPTS=200
 all_ready=false
 export TEST_SUITES="validationsuite"
 export FAIL_FAST="-ginkgo.failFast"
 export DONT_REBUILD_TEST_BINS=true
+export TIMEOUT="${TIMEOUT:-5400}" # results in 90 minutes timeout
```
SchSeba, line 19: please just add a small `echo` to tell how much time we are going to wait until timeout

PR #1152, opened 2022-06-22
Bug 2095289: ztp: Update cpuset calculation in accelerated-container-startup.sh

donpenney:
Updated the calculation of unrestricted cpuset in
accelerated-container-startup.sh to include reserved system cpus,
which may not be included in the default cpuset.

Signed-off-by: Don Penney <dpenney@redhat.com>

/cc @browsell


imiller0 (approved):
/lgtm

PR #1154, opened 2022-06-26
features: gatekeeper: update source package name

sabinaaledort:
The name of the gatekeeper-operator package in redhat-operators catalogsource was updated.


sabinaaledort:
/hold
trying to understand why the name of the package was changed.

sabinaaledort:
failing because the v0.2.3 gatekeeper is pointing to an u/s image that doesn't exist
 https://github.com/stolostron/gatekeeper-operator/blob/v0.2.3/bundle/manifests/gatekeeper-operator.clusterserviceversion.yaml#L321

hack/setup-build-index-image.sh
```diff
@@ -90,8 +90,10 @@ spec:
           cd ..
 
 
-          git clone --single-branch --branch GATEKEEPER_VERSION https://github.com/open-cluster-management/gatekeeper-operator.git
+          git clone --single-branch --branch GATEKEEPER_VERSION https://github.com/stolostron/gatekeeper-operator.git
           cd gatekeeper-operator
+          # replace u/s image with d/s
+          sed -i 's/quay.io\/gatekeeper\/gatekeeper-operator/registry.redhat.io\/rhacm2\/gatekeeper-rhel8-operator/g' bundle/manifests/gatekeeper-operator.clusterserviceversion.yaml
```
SchSeba, line 96: I don't think we should do something like that here

sabinaaledort: I don't think we need this change anymore. I did that to fix ci errors, but the issue was resolved, and anyway we are checking if gatekeeper should still be tested, so maybe we shouldn't touch or change is now. I'm closing this PR.

sabinaaledort:
Did that to fix ci errors, but the issue was resolved, and anyway we are checking if gatekeeper should still be tested, so maybe we shouldn't touch or change is now.

PR #1166, opened 2022-07-05
Bug 2087219: ztp: Renames the partition template and moves diskPartition to cluster spec

alosadagrande:
Signed-off-by: Alberto Losada [alosadag@redhat.com](mailto:alosadag@redhat.com)

This PR:

Renames source-crs/extra-manifest/image-registry-partition-mc.yaml.tmpl to ../source-crs/extra-manifest/var-device-partition-mc.yaml.tmpl since it can be used for multiple purposes. Notice that adding a separate storage to a partition is officially supported for mounting /var or a subdirectory of /var [[1]](https://docs.openshift.com/container-platform/4.10/installing/installing_bare_metal/installing-bare-metal-network-customizations.html#installation-user-infra-machines-advanced_disk_installing-bare-metal-network-customizations). Test code has been modified accordingly

diskPartition field is moved from node spec to cluster spec since it renders an MC that will be applied to all the nodes in the cluster. We expect all nodes in a 3 node cluster to use the same hardware and storage configuration.

If this PR is finally merged we probably need to review the imageregistry documentation to update the MC name and siteConfig examples for disk partitioning. We have also to document that diskPartition is a cluster spec instead of a node spec and the rationale behind it.


ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -1002,18 +1002,18 @@ spec:
   clusterImageSetNameRef: "openshift-v4.8.0"
   clusters:
   - clusterName: "cluster1"
+    diskPartition:
+      - device: /dev/sda
+        partitions:
+          - mount_point: /var/imageregistry
```
vitus133, line 1008: Does it make sense to make this mount point name more generic (since an additional partition can be used for any purpose)? You have already refactored "image-registry" to "var-device" in the MC name, so it's only here

vitus133 (commented):
Hi,
Could it be possible to fix the associated Readme files, e.g [this](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/ImageRegistry.md)?
/cc @pixelsoccupied

serngawy (commented):
sounds good to me, just a little concern about back compatibility for previous version.It could be not an issue if we back port those changes.

pixelsoccupied:
Reviving this PR. 

I think change is straightforward and also makes sense now that we are trying it out the multinode cluster. But as mentioned..originally this feature was scoped for SNO and image-registry which is why I think all relevant docs/ examples should be part of this PR. 

I think it's important to at least update [docs](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/ImageRegistry.md) (preferably in this PR) as TELCO docs team is referring to this and is about to wrap up their [work](https://github.com/openshift/openshift-docs/pull/50904).

FYI also we need to backport this to 4.10...another reason why it would super convenient to put everything here

pixelsoccupied:
@alosadagrande -- Ian just made a great point about backwards compatibility. We need things to always be backwards compatible otherwise we get very unhappy customers. 

I think we should revert SiteConfig/supporting CRs/docs to what it was SNO and instead use this PR as a new feature. So basically multi-node use `Clusters` and for SNO use `Nodes`

FYI @imiller0

alosadagrande:
Ok, but does it make sense to add new code for multinode clusters? or just update the documentation and say something like _"If you want a device partition in all nodes of a multinode cluster, only one node must have the diskPartition configuration. The rest of the nodes of the cluster will apply it too."_

That would just work, but IMO it is not very user-friendly. I would not expect a node configuration to be applied to all nodes in a cluster,e.g. it's a specific configuration for a node. Moving the diskPartition to the cluster level, makes sense since SNOs, compact or multicluster installations are just a cluster.

jgato:
We are having an use case where it is need to wipe some disks on some nodes, for a cluster re-installation. From the previous installation, some disks would start in a "dirty" stage. This avoids to re-install, for example, CEPH. If during the installation you can select some nodes and some disks to be wiped-out, it would be a cool feature. 

So I like @alosadagrande proposal of making the 'source-crs/extra-manifest/image-registry-partition-mc.yaml.tmp' more generic. 

IMHO, The diskParition should be at Node level, this give us a lot much more flexibility. In this case, we only want to wipe disks at masters (because CEPH disks will be placed there). But there are more use cases, multinode clusters can be pretty heterogeneous. Different nodes will have different disks. So it is good if you can play with them at node level. Actually, this is happening now. As an alternative to diskPartition, we are using an Ignition MC. The MC affect to all masters, but different masters have different disks.

PR #1167, opened 2022-07-05
Allow running cnf-tests in parallel 

sabinaaledort:
Doing this to allow running more than one instance of the tests in parallel on the same cluster, per feature, under different namespaces.

### DIFF
--- a/cnf-tests/testsuites/e2esuite/test_suite_test.go
+++ b/cnf-tests/testsuites/e2esuite/test_suite_test.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security"
 	"log"
+	"os"
 	"path"
 	"testing"
 	"time"
@@ -60,13 +61,19 @@ import (
 // see - https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/framework/framework.go
 
 var (
-	junitPath  *string
-	reportPath *string
+	junitPath          *string
+	reportPath         *string
+	skipTestNSCreation bool
 )
 
 func init() {
 	junitPath = flag.String("junit", "junit.xml", "the path for the junit format report")
 	reportPath = flag.String("report", "", "the path of the report file containing details for failed tests")
+
+	skipTestNSCreation = false
+	if os.Getenv("SKIP_TEST_NAMESPACES_CREATION") == "true" {
+		skipTestNSCreation = true
+	}
 }
 
 func TestTest(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cnf-tests/testsuites/pkg/namespaces/namespaces.go
```diff
@@ -41,6 +41,9 @@ var SroTestNamespace = "oot-driver"
 
 var BondTestNamespace = "bond-testing"
 
+// NamespaceTesting is the namespace for CNF features testing
+var NamespaceTesting string
```
fedepaol, line 45: Just "Testing" as it's gonna be namespaces.Testing

hack/feature-wait.sh
```diff
@@ -13,7 +13,7 @@ if [ "$FEATURES" == "" ]; then
 fi
 
 all_ready=false
-export TEST_SUITES="validationsuite"
+export TEST_SUITES="validationsuite configsuite"
```
fedepaol, line 16: typo in the commit: configusuit

fedepaol: How comes we need this? Isn't the config step called after the validation?
The configuration step is meant to happen right after the validation and before running the suite

cnf-tests/testsuites/e2esuite/test_suite_test.go
```diff
@@ -93,47 +94,49 @@ func TestTest(t *testing.T) {
...
-	Expect(testclient.Client).NotTo(BeNil())
-	// create test namespace
-	ns := &corev1.Namespace{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: namespaces.NamespaceTesting,
-		},
-	}
-	_, err := testclient.Client.Namespaces().Create(context.Background(), ns, metav1.CreateOptions{})
-	Expect(err).ToNot(HaveOccurred())
+	if os.Getenv("SKIP_TEST_NAMESPACES_CREATION") != "true" {
```
fedepaol, line 97: Can we read this in init and set a flag?

cnf-tests/testsuites/e2esuite/test_suite_test.go
```diff
@@ -97,7 +97,7 @@ var _ = BeforeSuite(func() {
 	// create test namespace
 	ns := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: testutils.NamespaceTesting,
+			Name: namespaces.NamespaceTesting,
```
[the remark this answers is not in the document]
fedepaol: Then why this must be overridden? I am probably missing something

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: I set it as a flag and then realized that I'm breaking the run-functests.sh script that way. The same script is used to run the validation, configuration and functests suites but we only have that flag in the functests suite so the other suites are failing for that. I can add a check to the run-functests.sh script which suite we are running, but if we run two I will have to slpit the command and I think that's ugly. I want to leave it this way unless you have a cleaner idea.

PR #1168, opened 2022-07-05
add a new s2i FEATURE

SchSeba:


sabinaaledort:
> i'm wondering about the name of the new feature, since this is not a general s2i i'ts dpdk-s2i

oh we won't be able to focus if it will be with dpdk

SchSeba:
> oh we won't be able to focus if it will be with dpdk

right that is why I didn't want to add dpdk into the name :)

PR #1174, opened 2022-07-12
cnf-tests: add mtu to nad builder

mmirecki:
Adding an mtu parameter to the nad builder.
Note that mtu is a mandatory parameter, to avoid issues
with leaving the deafult value, which causes issues in
some test environments.


fedepaol:
It was enough to have it in a separate commit, no need for another PR. In any case, can you extend the PR descritption / the commit body?

mmirecki:
> It was enough to have it in a separate commit, no need for another PR. In any case, can you extend the PR descritption / the commit body?

updated

PR #1186, opened 2022-07-22
Bug 2110015: Revert ptp interval changes

imiller0:
The underlying reason to align the summary_interval and logSyncInterval
was due to a validating webhook in the PTP operator. This validation has
been removed and the values no longer need to be aligned. This commit is
being reverted because the change to the source CRs would cause an
update to the customer's configurations on upgrade which is no longer
required. To avoid this update and potential disruption the change is
being reverted. The customer can supply whatever values are appropriate
for their use case in the PolicyGenTemplate override.

### DIFF
--- a/feature-configs/5g-ref-nw/profile-base/du-general/ptp/ptpconfig-grandmaster.yaml
+++ b/feature-configs/5g-ref-nw/profile-base/du-general/ptp/ptpconfig-grandmaster.yaml
@@ -36,7 +36,7 @@ spec:
       # Port Data Set
       #
       logAnnounceInterval -3
-      logSyncInterval 0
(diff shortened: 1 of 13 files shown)


lack:
@imiller0 : Given that ZTP can deploy any version of OCP, including old ones, is there a chance we might install an older version of PTP which would then fail because these values aren't in alignment?

imiller0:
I believe that the webhook was removed prior to any release of the PTP operator with it in place. So I think the answer is no. @aneeshkp or @kenyis can you confirm?

aneeshkp:
@imiller0 when bug was reported we patched it  to removed the bug in same version of OCP .

PR #1192, opened 2022-07-26
features: Update kustomization yamls for the new s2i feature

sabinaaledort:
Following https://github.com/openshift-kni/cnf-features-deploy/pull/1168 there are some kustomization yamls that were not updated.


SchSeba:
nice catch!

PR #1193, opened 2022-07-26
cnf-tests: bond cni over sriov

mmirecki:
Verifies a bond interface over sriov interfaces starts correctly.

This adds some code from another PR not being backported.
Also it does not use the utils introduced in the PR elsewhere in the code.


mmirecki:
@SchSeba @fedepaol

PR #1195, opened 2022-07-27
cnf-tests: Split setup and cleanup by feature

sabinaaledort:
Doing this for parallel run of the tests. 
To be able running more than one instance of cnf-tests at a time, by feature, we need to split the setup and cleanup for every tested feature, so a running instance of the tests won't delete the resources of another tested feature.


cnf-tests/testsuites/e2esuite/test_suite_test.go
```diff
@@ -4,57 +4,41 @@
...
-	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/ovs_qos"    // this is needed otherwise the ovs_qos test won't be executed
-	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/ptp"        // this is needed otherwise the ptp test won't be executed
-	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/s2i"        // this is needed otherwise the dpdk test won't be executed
-	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/sctp"
-	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/sctp"     // this is needed otherwise the sctp test won't be executed
-	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security" // this is needed otherwise the security test won't be executed
-	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/sro"      // this is needed otherwise the sro test won't be executed
-	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/vrf"
```
SchSeba, line 30: we didn't call the vrf tests?

sabinaaledort: seems like it

SchSeba: can you please add it?

```suggestion
_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/vrf
```

sabinaaledort: done

cnf-tests/testsuites/e2esuite/test_suite_test.go
```diff
@@ -74,6 +76,15 @@ func init() {
 	if os.Getenv("SKIP_TEST_NAMESPACES_CREATION") == "true" {
 		skipTestNSCreation = true
 	}
+
+	featuresVar := os.Getenv("FEATURES")
```
fedepaol, line 81: I'd rather accept a comma separated value (and error if the given value is not a valid one). This won't catch typos.

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,266 @@
...
+func (p *DpdkManager) BeforeSuite() error {
+	return namespaces.Create(p.ns(), testclient.Client)
+}
+
+func (p *DpdkManager) AfterSuite() error {
+	return namespaces.Delete(p.ns(), testclient.Client)
+}
+
+func (p *DpdkManager) ns() string {
+	return dpdkNS
```
fedepaol, line 75: Why can't you just hardcode the ns above? Same for the others.

cnf-tests/testsuites/e2esuite/test_suite_test.go
```diff
@@ -101,70 +112,25 @@ func TestTest(t *testing.T) {
...
-
-		err = namespaces.Create(perfUtils.NamespaceTesting, testclient.Client)
-		Expect(err).ToNot(HaveOccurred())
-
-		err = namespaces.Create(namespaces.DpdkTest, testclient.Client)
-		Expect(err).ToNot(HaveOccurred())
-
-		err = namespaces.Create(testutils.GatekeeperTestingNamespace, testclient.Client)
-		Expect(err).ToNot(HaveOccurred())
+		namespaces.Create(testutils.NamespaceTesting, testclient.Client)
```
fedepaol, line 116: Can this be another feature?

cnf-tests/testsuites/e2esuite/test_suite_test.go
fedepaol: Also, calling the variable "FEATURES" is misleading. What that really drives is what setup / teardown, plus we may have inconsistencies between focus and FEATURES.

Ideally, focus / skip is what should drive this too. One dumb way to do so could be parse os.Args and find the gingko.focus / ginkgo.skip parameter and update the map accordingly, not sure if reading directly the argument will work too.

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,266 @@
...
+	vrfNS            = vrf.TestNamespace
+	sriovNS          = sriovNamespaces.Test
+	xtu32NS          = namespaces.XTU32Test
+	ovsqosNS         = namespaces.OVSQOSTest
+	tuningcniNS      = security.TestNamespace
+	tuningcniSriovNS = security.SriovTestNamespace
+	bondcniNS        = namespaces.BondTestNamespace
+)
+
+type SuiteManager interface {
```
fedepaol, line 39: SuiteFixtures

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,266 @@
...
+	BeforeSuite() error
```
fedepaol, line 40: Setup / Teardown

cnf-tests/testsuites/e2esuite/test_suite_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/sctp"                       // this is needed otherwise the sctp test won't be executed
 	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security"                   // this is needed otherwise the security test won't be executed
 	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/sro"                        // this is needed otherwise the sro test won't be executed
+	_ "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/vrf"                        // this is needed otherwise the performance test won't be executed
```
fedepaol, line 27: Wasn't it removed in the previous commit? If so, can we adjust it there?
Also, the comment is wrong

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: you mean in the ns function? i can, but i thought it will be easier to change when needed if all the namespaces are above in a list.

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: the import of the package was removed since we don't need it. what i added here is for execution, what we previously didn't have.

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: yes

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: but if gingko.focus is set to test name? how do you get the feature name from it? FEATURES var is always features, not tests.

cnf-tests/testsuites/pkg/features/features.go
fedepaol: not sure I understood what it'd be simpler

cnf-tests/testsuites/e2esuite/test_suite_test.go
fedepaol: But this assumes you are calling --focus with a list of features. Otherwise,the tests would be failing because the ns is not there.
So, --focus / --skip and FEATURES (or whatever the right name is) must always be in sync, otherwise it's misconfiguration. What I'd like to see, is avoiding doing the configuration twice exposing the risk of misconfigs

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: instead old looking at this huge file. i can change that. not critical.

fedepaol: The thing I was talking about is the ns() function. Why would we need it?

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: if you set --focus to a test and FEATURES to the feature name you get the ns created. but if you don't have FEATURES var but only use --focus, which is set to a test name, how do you know which ns to create?

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: isn't it like beforeSuite and afterSuite?

fedepaol: There's no suite here. I'd give names to function about what they do

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: because we use the ns for some of the features twice (in before and after), so when changing the ns we will have to change it in two places instead of just changing it once in the var.

cnf-tests/testsuites/e2esuite/test_suite_test.go
fedepaol: But this is fixing something that was broken in the previous commit because the import was removed?

sabinaaledort: we onlt had an import there to get the ns, which we don't need anymore. we never had the import for test execution.

fedepaol: Still, that commit broke the execution of vrf tests

cnf-tests/testsuites/pkg/features/features.go
fedepaol: how comes, if we have the list of constants above? And even if it was simpler, it can be a field of the manager instead of a method.

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: they ran here successfully before i even added the second commit with the execution import https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-kni_cnf-features-deploy/1195/pull-ci-openshift-kni-cnf-features-deploy-master-e2e-gcp-ovn/1553643919286210560

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: sorry, was talking about the constants

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: i introduced a new var to focus on tests (FOCUS_TESTS), so FEATURES can be just features names.

cnf-tests/testsuites/e2esuite/test_suite_test.go
fedepaol: yup, but they didn't run the vrf tests I guess. That's what I meant by breaking it.

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+	tuningcniSriovNS = security.SriovTestNamespace
+	bondcniNS        = namespaces.BondTestNamespace
+)
+
+type ConfigManager interface {
+	Setup() error
+	Cleanup() error
+}
+
+type PerformanceManager struct {
```
fedepaol, line 44: Sorry I stress on naming but...
PerformanceManager is not managing Performance , DpdkManager is not managing dpdk and so on. They are not doing what their name promises, they just do setup / teardown. 

That's why I suggested the name Fixture
See for example https://en.wikipedia.org/wiki/Test_fixture or https://code-craftsmanship-saturdays.gitbook.io/software-testing-fundamentals/test-fixtures 

If you find another name that suggests setting up / tearing down I am either fine, but I think manager is not the right term here.

Same for featureManagerMap above.

cnf-tests/testsuites/e2esuite/test_suite_test.go
fedepaol: Fair enough, thanks!

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (p *PerformanceManager) Cleanup() error {
+	if !discovery.Enabled() {
+		perfClean.All()
+	}
+
+	return namespaces.Delete(performanceNS, testclient.Client)
+}
+
+type DpdkManager struct {
```
fedepaol, line 59: nit: DPDKManager (or what comes in place of manager), same for all the others following when using acronyms (PTP, SCTP)
See https://github.com/golang/go/wiki/CodeReviewComments#initialisms

(sorry I didn't notice it before)

hack/run-functests.sh
```diff
@@ -34,6 +32,9 @@ elif [ "$FEATURES" == "" ]; then
   exit 1
 else
   FOCUS="-ginkgo.focus="$(echo "$FEATURES" | tr ' ' '|')
```
cgoncalves, line 34: In a earlier comment above you wrote:
> i introduced a new var to focus on tests (FOCUS_TESTS), so FEATURES can be just features names.

Should this line be removed?

hack/run-functests.sh
```diff
@@ -34,6 +32,9 @@ elif [ "$FEATURES" == "" ]; then
...
+  if [ "$FOCUS_TESTS" != "" ]; then
```
cgoncalves, line 35: This new environment variable should be documented in `cnf-tests/README` as a replacement for "-ginkgo.focus"

cgoncalves (changes requested):

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: i meant the vrf tests were running in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-kni_cnf-features-deploy/1195/pull-ci-openshift-kni-cnf-features-deploy-master-e2e-gcp-ovn/1553643919286210560

hack/run-functests.sh
sabinaaledort: removed why? we focus on features (with FEATURES) or on tests (with FOCUS_TESTS)

hack/run-functests.sh
sabinaaledort: done

cnf-tests/testsuites/e2esuite/test_suite_test.go
```diff
@@ -4,75 +4,79 @@
...
 
 var (
-	junitPath          *string
-	reportPath         *string
-	skipTestNSCreation bool
+	junitPath  *string
+	reportPath *string
 )
 
+var suiteFixtureMap = map[string]features.SuiteFixture{
```
SchSeba, line 52: I think we are missing s2i no?

cnf-tests/testsuites/e2esuite/test_suite_test.go
```diff
@@ -4,75 +4,79 @@
...
-	if os.Getenv("SKIP_TEST_NAMESPACES_CREATION") == "true" {
-		skipTestNSCreation = true
+	featuresVar := os.Getenv("FEATURES")
+	if featuresVar != "" {
+		for feature := range suiteFixtureMap {
+			if !strings.Contains(featuresVar, feature) {
+				delete(suiteFixtureMap, feature)
+			}
+		}
 	}
```
SchSeba, line 80: nit: can you please print here the features we are going to run? this can help for debugs

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
+package features
+
+import (
+	"fmt"
+
+	sriovClean "github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/clean"
```
SchSeba, line 6: nit: please split the imports to

1. golang packages
2. external projects
3. internal

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/namespaces"
+	testutils "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/utils"
+	numaserialconf "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
+	perfUtils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	perfClean "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/clean"
+	ptpClean "github.com/openshift/ptp-operator/test/utils/clean"
+)
+
+var (
+	paoNS            = perfUtils.NamespaceTesting
```
SchSeba, line 25: this is a bit of a mess can you please just move all the namespaces as const into the namespace package?

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/discovery"
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/namespaces"
+	testutils "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/utils"
+	numaserialconf "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
+	perfUtils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	perfClean "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/clean"
+	ptpClean "github.com/openshift/ptp-operator/test/utils/clean"
+)
+
+var (
```
SchSeba, line 24: there is no need for this lets just one everything from the namespaces package

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func (p *SriovFixture) Setup() error {
+	return nil
+}
+
+func (p *SriovFixture) Cleanup() error {
+	sriovClean.All()
+
+	err := namespaces.Delete(tuningcniSriovNS, testclient.Client)
```
SchSeba, line 115: this should go into the tuning no?

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return namespaces.Delete(sctpNS, testclient.Client)
+}
+
+type XTU32Fixture struct {
+}
+
+func (p *XTU32Fixture) Setup() error {
+	return nil
```
SchSeba, line 143: why we don't create the namespace here?

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (p *XTU32Fixture) Cleanup() error {
+	return namespaces.Delete(xtu32NS, testclient.Client)
+}
+
+type VRFFixture struct {
+}
+
+func (p *VRFFixture) Setup() error {
+	return nil
```
SchSeba, line 154: why we don't create the namespace here?

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (p *VRFFixture) Cleanup() error {
+	return namespaces.Delete(vrfNS, testclient.Client)
+}
+
+type OVSQOSFixture struct {
+}
+
+func (p *OVSQOSFixture) Setup() error {
+	return nil
```
SchSeba, line 165: why we don't create the namespace here?

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+func (p *NumaresourcesFixture) Cleanup() error {
+	numaserialconf.Teardown()
+	return nil
+}
+
+type FECFixture struct {
+}
+
+func (p *FECFixture) Setup() error {
+	return nil
```
SchSeba, line 194: why we don't create the namespace here?

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+func (p *FECFixture) Cleanup() error {
+	fec.Clean()
+	return nil
+}
+
+type TuningcniFixture struct {
+}
+
+func (p *TuningcniFixture) Setup() error {
+	return nil
```
SchSeba, line 206: why we don't create the namespace here?

cnf-tests/testsuites/pkg/features/features.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (p *TuningcniFixture) Cleanup() error {
+	return namespaces.Delete(tuningcniNS, testclient.Client)
+}
+
+type BondcniFixture struct {
+}
+
+func (p *BondcniFixture) Setup() error {
+	return nil
```
SchSeba, line 217: why we don't create the namespace here?

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: we only have here the features that we set/delete something for them before running the tests. is there a setup for s2i?

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: these tests are filtered by sriov https://github.com/openshift-kni/cnf-features-deploy/blob/57a05afbe836e9210ad638f6abe65f72b569148a/cnf-tests/testsuites/e2esuite/security/tuning_sriov.go#L34

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: we weren't creating it before. i didn't want to change the existing setup https://github.com/openshift-kni/cnf-features-deploy/blob/57a05afbe836e9210ad638f6abe65f72b569148a/cnf-tests/testsuites/e2esuite/test_suite_test.go#L103

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: we weren't creating it before. i didn't want to change the existing setup https://github.com/openshift-kni/cnf-features-deploy/blob/57a05afbe836e9210ad638f6abe65f72b569148a/cnf-tests/testsuites/e2esuite/test_suite_test.go#L103

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: we weren't creating it before. i didn't want to change the existing setup https://github.com/openshift-kni/cnf-features-deploy/blob/57a05afbe836e9210ad638f6abe65f72b569148a/cnf-tests/testsuites/e2esuite/test_suite_test.go#L103

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: we weren't creating it before. i didn't want to change the existing setup https://github.com/openshift-kni/cnf-features-deploy/blob/57a05afbe836e9210ad638f6abe65f72b569148a/cnf-tests/testsuites/e2esuite/test_suite_test.go#L103

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: we weren't creating it before. i didn't want to change the existing setup https://github.com/openshift-kni/cnf-features-deploy/blob/57a05afbe836e9210ad638f6abe65f72b569148a/cnf-tests/testsuites/e2esuite/test_suite_test.go#L103

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: we weren't creating it before. i didn't want to change the existing setup https://github.com/openshift-kni/cnf-features-deploy/blob/57a05afbe836e9210ad638f6abe65f72b569148a/cnf-tests/testsuites/e2esuite/test_suite_test.go#L103

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: that's what we currently have, i don't see anything for s2i https://github.com/openshift-kni/cnf-features-deploy/blob/57a05afbe836e9210ad638f6abe65f72b569148a/cnf-tests/testsuites/e2esuite/test_suite_test.go#L103

cnf-tests/testsuites/pkg/features/features.go
SchSeba: ok we can do it in a following PR I think

sabinaaledort: do we need to create it? i guess it is currently done somewhere in the test suite

cnf-tests/testsuites/pkg/features/features.go
sabinaaledort: i can't do that for the ones that are taken from the suites, because we get an import cycle

sabinaaledort: i changed it so we can just use the original vars

cnf-tests/testsuites/e2esuite/test_suite_test.go
sabinaaledort: these are not the features we are going to run, these are only the features (from the features we are going to run) that need a configuration.

hack/run-functests.sh
SchSeba: can you please extend a bit more the doc to explain the difference between the FEATURES and the FOCUS_TESTS.

correct me if I am wrong here but if I understand this right if I want to focus on a specific sriov test I will use both configs right?

```
FEATURES=sriov FOCUS_TESTS="sriov resource injector...."
```

sabinaaledort: right

SchSeba:
/approve
/lgtm

giving time to @cgoncalves for more comments
/hold

PR #1208, opened 2022-08-11
ztp: Prevent changing immutable objects on sno expansion

vitus133:
This change allows deploying additional SNO workers with ZTP.
When adding a worker to an existing SNO, user
adds an additional node to the existing SiteConfig.
The Generator is processing the modified siteconfig
and adds an additional line in AgentClusterInstall:
workerAgents: 1
However, agentclusterinstall is immutable.There is an
admission webhook that declines the change,
and the sync fails:
Attempted to change AgentClusterInstall.Spec which is
immutable after install started, except for ClusterMetadata
fields. Unsupported change: ProvisionRequirements.WorkerAgents: (0 => 1)

/cc @lack @serngawy

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/siteConfig.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfig.go
@@ -209,9 +209,11 @@ func (rv *Clusters) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if rv.NumMasters != 1 && rv.NumMasters != 3 {
 		return fmt.Errorf("Number of masters (counted %d) must be exactly 1 or 3", rv.NumMasters)
 	}
-	// Autodetect ClusterType based on the node counts
+	// Autodetect ClusterType based on the node counts and fix number of workers to 0 for sno.
+	// The latter prevents AgentClusterInstall from being mutated upon SNO expansion
 	if rv.NumMasters == 1 {
 		rv.ClusterType = SNO
+		rv.NumWorkers = 0
 	} else {
 		rv.ClusterType = Standard
 	}


ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -209,9 +209,11 @@ func (rv *Clusters) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if rv.NumMasters != 1 && rv.NumMasters != 3 {
 		return fmt.Errorf("Number of masters (counted %d) must be exactly 1 or 3", rv.NumMasters)
 	}
-	// Autodetect ClusterType based on the node counts
+	// Autodetect ClusterType based on the node counts and fix number of workers to 0 for sno.
+	// The latter prevents AgentClusterInstall from being mutated upon SNO expansion
 	if rv.NumMasters == 1 {
 		rv.ClusterType = SNO
+		rv.NumWorkers = 0
```
serngawy, line 216: rv.NumWorkers is set to 0 at [L200](https://github.com/openshift-kni/cnf-features-deploy/pull/1208/files#diff-5a7eeb6977f4204a7d5c1322a7e1fe7957d13801d22c91348770ef8a71c22466R200)

vitus133: But then at line 206 it is incremented if a worker is added in siteconfig (SNO expansion case).
This exactly leads to my problem of AgentClusterInstall mutation

serngawy: okay, got it

PR #1211, opened 2022-08-16
CNF-3568: ztp: TALM subscription in argocd deployment

pixelsoccupied:
Add new Subscription CR in during hub cluster setup

### DIFF
--- a/ztp/gitops-subscriptions/argocd/deployment/kustomization.yaml
+++ b/ztp/gitops-subscriptions/argocd/deployment/kustomization.yaml
@@ -8,3 +8,4 @@ resources:
   - gitops-cluster-rolebinding.yaml
   - clusters-app.yaml
   - policies-app.yaml
+  - talm-operator-subscription.yaml
--- a/ztp/gitops-subscriptions/argocd/deployment/talm-operator-subscription.yaml
+++ b/ztp/gitops-subscriptions/argocd/deployment/talm-operator-subscription.yaml
@@ -1,0 +1,14 @@
+# Subscription CR to install Topology Aware Lifecycle Manager (TALM)
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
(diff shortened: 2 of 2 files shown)


pixelsoccupied:
/hold 
testing with a new hub cluster

ztp/gitops-subscriptions/argocd/deployment/talm-operator-subscription.yaml
```diff
@@ -0,0 +1,14 @@
+# Subscription CR to install Topology Aware Lifecycle Manager (TALM)
```
imiller0, line 1: I think this needs to be included in the kustomization.yaml or it won't get installed based on the user documentation (says to run oc apply -k <this directory>)

ztp/gitops-subscriptions/argocd/deployment/kustomization.yaml
```diff
@@ -8,3 +8,4 @@ resources:
   - gitops-cluster-rolebinding.yaml
   - clusters-app.yaml
   - policies-app.yaml
+  - talm-operator-subscription.yaml
```
imiller0, line 11: Based on feedback from QE, because the environment is likely disconnected and the catalog source name will vary) it may be best to not automatically include the Subscription in the deployment (so remove this line from the kustomization.yaml). Instead, can you work with the documentation team to update [this section](https://docs.openshift.com/container-platform/4.10/scalability_and_performance/ztp-deploying-disconnected.html#ztp-preparing-the-hub-cluster-for-ztp_ztp-deploying-disconnected) of the docs to refer the user to this sample Subscription as a way to install the TALM operator (pointing out the need to update the catalogsource as needed)

pixelsoccupied:
Closing the PR. Instead this will be documented using https://issues.redhat.com/browse/TELCODOCS-928

PR #1215, opened 2022-08-18
cnf-tests: Fix PSA and S2I filter

SchSeba:

### DIFF
--- a/cnf-tests/testsuites/e2esuite/gatekeeper/gatekeeper.go
+++ b/cnf-tests/testsuites/e2esuite/gatekeeper/gatekeeper.go
@@ -163,7 +163,7 @@ var _ = Describe("gatekeeper", func() {
 					err := client.Get(context.Background(), podKey, pod)
 					Expect(err).ToNot(HaveOccurred())
 					return pod.GetLabels()["mutated"] == "true" && pod.GetAnnotations()["mutated"] == "true"
-				}, 10*time.Second, 2*time.Second).Should(Equal(true), "Mutations are not applied")
+				}, 20*time.Second, 2*time.Second).Should(Equal(true), "Mutations are not applied")
 			},
 		)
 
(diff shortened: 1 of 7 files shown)


cnf-tests/testsuites/e2esuite/gatekeeper/gatekeeper.go
```diff
@@ -52,7 +52,7 @@ var _ = Describe("gatekeeper", func() {
 
 		for _, namespace := range namespacesUsed {
 			if namespaces.Exists(namespace, client) {
-				err := client.Namespaces().Delete(context.Background(), namespace, metav1.DeleteOptions{})
+				err := namespaces.Create(namespace, client.CoreV1Interface)
```
cgoncalves, line 55: Why create instead of delete if the namespace already exists?
In line 71, we wait until the namespaces are deleted so with this change that check would fail, no?

cnf-tests/testsuites/e2esuite/gatekeeper/gatekeeper.go
```diff
@@ -884,20 +882,21 @@ var _ = Describe("gatekeeper", func() {
...
 				},
 			}
 			err = client.Create(context.Background(), mutationEnabledNamespace)
 			Expect(err).NotTo(HaveOccurred())
 
 			mutationDisabledNamespace := &corev1.Namespace{
 				ObjectMeta: metav1.ObjectMeta{
-					Name: mutationDisabledNamespace,
+					Name:   mutationDisabledNamespace,
+					Labels: namespaces.GetPsaLabels(),
```
sabinaaledort, line 899: nit: can we use here the labels var?

cnf-tests/testsuites/pkg/namespaces/namespaces.go
```diff
@@ -89,6 +90,15 @@ func init() {
 	}
 }
 
+func GetPsaLabels() map[string]string {
```
sabinaaledort, line 93: nit: maybe GetPsaLabels ->  GetPSALabels

PR #1216, opened 2022-08-21
ztp: add SNO expansion readme

vitus133:
SNO expansion is a scenario where initially an SNO
is deployed into production, and at a later date
the cluster size is increased by addition of one
or more worker nodes.
The resulting cluster will contain two nodes:
    - The original SNO serving as the control plane for itself
    and the additional node(s)
    - Worker node(s) without any control plane components

/cc @imiller0  @browsell  @lack @serngawy


ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+The resulting cluster will contain two nodes:
+- Original SNO serving as the control plane for itself and the additional node
+- One additional node without any control plane components (controlled by the SNO).
+
+This transition incurs zero downtime on the original SNO node.
+Addition of more than one worker node to the SNO cluster is possible, but has not been verified in this version.
+
+## Prerequisites ##
+1. A cluster installed and configured using the GitOps ZTP fow, as described in [GitOps ZTP pipeline](README.md)
+1. ACM 2.6 (or above) with MultiClusterHub created and configured, running on OCP 4.11 (or above) bare metal cluster
```
browsell, line 13: Why wouldn't this work with 2.5 ?

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+1. Central Infrastructure Management configured as described in [Advanced Cluster Management documentation](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/clusters/managing-your-clusters#enable-cim)
+1. TALM and openshift-gitops operators are installed and configured on the hub cluster
+1. The DNS serving the cluster is configured to properly resolve internal API endpoint `api-int.<cluster_name>.<base_domain>`. Please check [Requirements for installing OpenShift on a single node](https://docs.openshift.com/container-platform/4.11/installing/installing_sno/install-sno-preparing-to-install-sno.html#install-sno-requirements-for-installing-on-a-single-node_install-sno-preparing) for more details.
+1. ZTP container version is 4.12 or above
+1. Initial state:
+   1. the spoke cluster is managed by the hub.
```
browsell, line 19: nit, The

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Writing image to disk
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Waiting for control plane
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Waiting for control plane
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Rebooting
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Done
+
+```
+#### __ManagedClusterInfo__
+
+When the worker node installation completes, its certificates are approved automatically. At his point the worker will appear in ManagedClusterInfo status:
```
browsell, line 145: nit, this

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+For example, for SNO, the git repository linked to the `policies` ArgoCD application will include:
+- [common-ranGen.yaml](example/policygentemplates/common-ranGen.yaml). This template usually contains a set of operator subscriptions and it's unlikely that it should be modified for a worker addition.
+- [group-du-sno-ranGen.yaml](example/policygentemplates/group-du-sno-ranGen.yaml)
+- [example-sno-site.yaml](example/policygentemplates/example-sno-site.yaml)
+- [ns.yaml](example/policygentemplates/ns.yaml)
+- [kustomization.yaml](example/policygentemplates/kustomization.yaml)
+
+The procedure of configuring the DU profile on the worker node is considered as an upgrade. To initiate the upgrade flow, user must update the existing policies, or create an additional ones, and then create a ClusterGroupUpgrade to reconcile the policies in the group of clusters.
+
+#### __Preparations (optional - for ZTP DU profile version 4.11 or below)__
```
browsell, line 166: or we create another manifest for worker mcp

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: According to my understanding of [this](https://coreos.slack.com/archives/CUPJTHQ5P/p1653476931714459) this code of central infra management is going to be included only in 2.6.

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: Fixed

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: fixed

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: This wouldn't help if the SNO was already provisioned with OperatorConfig setting the daemonSelector labels to `master`.  In this case changing the operatorConfig policy to 'worker' would only add `worker` label to the node selector, but it wouldn't remove the `master` label. The worker would remain unselected.
For the future I think we need to change the way we do policies for OperatorConfig - either stop populating daemonset node selector, or set it fixed to `worker`. 
https://github.com/openshift-kni/cnf-features-deploy/pull/1214
https://github.com/openshift-kni/cnf-features-deploy/pull/1213

filanov:
cc: @omertuc @avishayt @paul-maidment

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+SNO expansion is a scenario where initially a SNO is deployed into production, and at a later date the cluster size is increased by the addition of a worker node.
+The resulting cluster will contain two nodes:
+- Original SNO serving as the control plane for itself and the additional node
+- One additional node without any control plane components (controlled by the SNO).
+
+This transition incurs zero downtime on the original SNO node.
+Addition of more than one worker node to the SNO cluster is possible, but has not been verified in this version.
+
+## Prerequisites ##
+1. A cluster installed and configured using the GitOps ZTP fow, as described in [GitOps ZTP pipeline](README.md)
```
serngawy, line 12: GitOps ZTP flow

ztp/gitops-subscriptions/argocd/SNOExpansion.md
lack: It's tough on SNO, since the `worker` pool is present but doesn't affect the node at all.

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: fixed

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
+# SNO expansion #
+
+SNO expansion is a scenario where initially a SNO is deployed into production, and at a later date the cluster size is increased by the addition of a worker node.
+The resulting cluster will contain two nodes:
+- Original SNO serving as the control plane for itself and the additional node
+- One additional node without any control plane components (controlled by the SNO).
+
+This transition incurs zero downtime on the original SNO node.
+Addition of more than one worker node to the SNO cluster is possible, but has not been verified in this version.
```
filanov, line 9: i don't know of any limitations of adding as many workers as you want @omertuc correct me if i'm wrong

vitus133: My thought here was that adding a worker to SNO would incur some costs in CPU use by the control plane components. We are going to measure this soon and put a "CPU price tag" on each additional worker.  Maybe the wording should reflect that?

filanov: That can help, from the wording it sounds like we support only adding one worker.

vitus133: Fixed

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+        priority: 10
+        resourceName: cplane_worker
+######
+# Comment 4  
+    - fileName: MachineConfigGeneric.yaml
+      policyName: "config-policy"
+      metadata:
+        labels:
+          machineconfiguration.openshift.io/role: worker
+        name: enable-workload-partitioning
```
imiller0, line 368: It may be worth having a section to discuss how workload partitioning works with SNO expansion. Checking my understanding here.... 
The expansion node will be a worker, so only those core openshift pods that are not bound (nodeSelector) to the master will run on this node. For pods which are not bound to the master node I suspect the scheduler won't place them on this expansion node until this workload partitioning config is put in place (since there are no `management.workload.openshift.io/cores` resources advertised).
Is that accurate?

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+kind: PolicyGenTemplate
+metadata:
+  name: "cnfdf15-worker"
+  namespace: "ztp-cnfdf15-policies"
+spec:
+  bindingRules:
+    # These policies will correspond to all clusters with this label:
+    sites: "cnfdf15"
+######
+# Comment 1
```
imiller0, line 294: Took me additional scrolling to realize these comments were below. Any reason to not include them inline?

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+              service.stalld=start,enable
+              service.chronyd=stop,disable
+```
+##### __Comment 1__
+Set MCP to "worker"
+##### __Comment 2__
+This PTP configuration is added to demonstrate that in case ptp configuration on the worker is different from the master, a separate `PtpConfig` object can be added to match the worker configuration. In this example the additional `PtpConfig` is matched against a different label selector.
+##### __Comment 3__
+Please note that SR-IOV related configurations have a different policy name than the other objects in this file. It's because they have a different ZTP wave. and must be applied separately.
+This example also demonstrates a different node selector label.
```
imiller0, line 427: Is this required or optional?

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+1. TALM and openshift-gitops operators are installed and configured on the hub cluster
+1. The DNS serving the cluster is configured to properly resolve internal API endpoint `api-int.<cluster_name>.<base_domain>`. Please check [Requirements for installing OpenShift on a single node](https://docs.openshift.com/container-platform/4.11/installing/installing_sno/install-sno-preparing-to-install-sno.html#install-sno-requirements-for-installing-on-a-single-node_install-sno-preparing) for more details.
+1. ZTP container version is 4.12 or above
+1. Initial state:
+   1. the spoke cluster is managed by the hub.
+   2. ArgoCD applications are synchronized and all policies are compliant on the spoke cluster.
+
+## Procedure ##
+### Deploying a worker node ###
+1. Assuming your cluster was deployed using [this SiteConfig manifest](example/siteconfig/example-sno.yaml), add your new worker node to `spec.clusters['example-sno'].nodes` list, for example:
```
imiller0, line 24: Can this also be added "day-0" if the user wants to initially deploy SNO+worker?

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
+# SNO expansion #
+
+SNO expansion is a scenario where initially a SNO is deployed into production, and at a later date the cluster size is increased by the addition of a worker node.
+The resulting cluster will contain two nodes:
+- Original SNO serving as the control plane for itself and the additional node
+- One additional node without any control plane components (controlled by the SNO).
```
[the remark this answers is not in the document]
vitus133: Rephrased

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+
+This transition incurs zero downtime on the original SNO node.
+Addition of workers to an SNO cluster will increase CPU resources used by the original (master) node for control plane functions it performs. Although there is no hard limit on amount of the additional workers that can be added, worker addition must come with re-evaluation of reserved CPU allocation on the master. 
+
+## Prerequisites ##
+1. A cluster installed and configured using the GitOps ZTP fow, as described in [GitOps ZTP flow](README.md)
+1. ACM 2.6 (or above) with MultiClusterHub created and configured, running on OCP 4.11 (or above) bare metal cluster
```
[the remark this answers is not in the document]
vitus133: We do have a specific dependence on 4.11. IIUC, ironic is supplied as a part of the OCP version, and we rely on new features introduced in 4.11 for SNO expansion.

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Writing image to disk
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Waiting for control plane
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Waiting for control plane
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Rebooting
+14fd821b-a35d-9cba-7978-00ddf535ff37   cnfdf15   true       worker   Done
+
+```
+#### __ManagedClusterInfo__
+
+When the worker node installation completes, its certificates are approved automatically. At his point the worker will appear in ManagedClusterInfo status:
```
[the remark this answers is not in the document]
vitus133: Fixed

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+{"daemonNodeSelector":{}}
+```
+
+If the `daemonNodeSelector` is set to `node-role.kubernetes.io/master: ""`, it must be cleared.
+
+**Important: changing the `daemonNodeSelector` will incur temporary PTP synchronization loss**
+
+More information on PTP operator `daemonNodeSelector` is available [in the PTP operator README file](https://github.com/openshift/ptp-operator#ptpoperatorconfig). 
+
+To clear the PTP operator `daemonNodeSelector`, deploy the following manifest on your hub cluster:
```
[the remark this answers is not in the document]
vitus133: The changes needed have been already merged in It's[1214](https://github.com/openshift-kni/cnf-features-deploy/pull/1214) and [1213](https://github.com/openshift-kni/cnf-features-deploy/pull/1213). The change was in the go code setting the daemon nede selectors to workers unconditionally. This works for all use cases known so far, because masters have the worker label as well.

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,483 @@
...
+      policyName: "config-policy-net"
+      metadata:
+        name: "f1u-network-worker"
+      spec:
+        resourceName: uplane_worker
+        vlan: 140
+    - fileName: SriovNetworkNodePolicy.yaml
+      policyName: "config-policy-net"
+      metadata:
+        name: "sriov-nnp-uplane-worker"
```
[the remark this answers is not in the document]
vitus133: Since this use case assumes workers are added at a later point in time, we have already deployed DU profile to the SNO with all the selectors set to "master". We are not going to reconfigure the master, as this will incur downtime (and we are required to avoid downtime). So we can only talk about equality of nodes between the additional workers. These could have been sharing a common configuration, but we don't have any label applied to workers that makes them unique (because worker label is applied to the original SNO too).
I think the right solution would be to implement node labeling API as a part of siteconfig. This will help us in other places as well (such as ODF storage nodes in CU deployments for example). It will also allow us to stop using master and worker labels everywhere as we do now.

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: Same problem as above - we have already deployed everything with "master" selector, so now we have to use different selectors

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: I was trying to do something similar to openshift documentation. I'll put them inline, I realize now that it is misleading

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: No, this is not a supported flow.

ztp/gitops-subscriptions/argocd/SNOExpansion.md
vitus133: Not sure then how any of the daemonset pods would be running. But they are... Let me check this

vitus133: Unless they are daemonset pods. It seems to be enough to apply the machineconfig defining the wl partitioning and reboot the node. 
I will have a deeper look into this. And for sure it should be documented somewhere. Is here the right place for it?

vitus133: I think I got it. Without the workload partitioning configured on the kubelet side, the pods would not be mutated to the new resource type. The resource they request will remain "cpu", and not management.workload.openshift.io/cores.
For example, this is the worker without WL partitioning enabled:
```bash
oc -n openshift-machine-config-operator describe po/machine-config-daemon-lp2js
...
    Requests:
      cpu:     20m
      memory:  50Mi
```
This is its "brother" on the master with WL partitioning enabled:
```bash
oc -n openshift-machine-config-operator describe po/machine-config-daemon-rsw44
...
Requests:
      management.workload.openshift.io/cores:  20
```
How the API server knows which pods to mutate and which not to? I guess it checks the kubelet capabilities  when the node boots and registers with the API.

browsell: The admission hook, checks for the following: 
1) Checks that it is a SNO
2) Checks that all nodes in the cluster are advertising the management.workload.openshift.io/cores
3) namespace has the following annotation: workload.openshift.io/allowed: management
4) pod has the following annotation: target.workload.openshift.io/management: {"effect":"PreferredDuringScheduling"}

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,374 @@
...
+SNO expansion is a scenario where initially a SNO is deployed into production, and at a later date the cluster size is increased by the addition of one or more worker nodes.
+The resulting cluster will contain at least two nodes:
+- Original SNO serving as the cluster control plane
+- One or more worker nodes without any control plane components (controlled by the SNO).
+
+This transition incurs zero downtime on the original SNO node.
+Addition of workers to an SNO cluster will increase CPU resources used by the original (master) node for control plane functions it performs. Although there is no hard limit on amount of the additional workers that can be added, worker addition must come with re-evaluation of reserved CPU allocation on the master. 
+
+## Prerequisites ##
+1. A cluster installed and configured using the GitOps ZTP fow, as described in [GitOps ZTP flow](README.md)
```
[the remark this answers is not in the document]
vitus133: fixed

ztp/gitops-subscriptions/argocd/SNOExpansion.md
```diff
@@ -0,0 +1,374 @@
...
+  - cnfdf15
+  enable: true
+  managedPolicies:
+  - cnfdf15-worker-config-policy
+  preCaching: false
+  remediationStrategy:
+    maxConcurrency: 1
+EOF
+```
+Please mote that the `ClusterGroupUpgrade` CR can be applied to the hub simultaneously with the worker node policies. In this case the remediation will start immediately after the worker node is approved by the cluster.
```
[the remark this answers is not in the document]
vitus133: fixed

serngawy (approved):

PR #1222, opened 2022-08-22
Bug 2049306: ztp: Fix the user overrides for a list

Missxiaoguo:
This PR fixes the issue where the generated CR has missing content when user provided overrides for a list have more elements than the ones in the source-cr.


imiller0 (commented):
/lgtm

PR #1239, opened 2022-09-06
s2i: switch to generic platform dpdk

SchSeba:
the following error:

```
 ERROR: This system does not support "VPCLMULQDQ".
          Please check that RTE_MACHINE is set correctly.
          EAL: FATAL: unsupported cpu type.
          EAL: unsupported cpu type.
          EAL: Error - exiting with code: 1
            Cause: Cannot init EAL: Operation not supported
```

switching to the general platform in dpdk


zeeke:
/lgtm

Here the docs: https://doc.dpdk.org/guides-21.11/linux_gsg/build_dpdk.html#adjusting-build-options

PR #1252, opened 2022-09-13
[release-4.11] ztp: OCPBUGS-496: run PTP daemons on worker nodes only - Revert

vitus133:
Reverts openshift-kni/cnf-features-deploy#1226
The reason for creating the original PR was to make the PTP
daemon selector suitable for deployments besides SNO. This PR is
being reverted because the change to the source CRs would cause a
potentially disruptive update to the customer's cluster upon upgrade.
A better way to tackle the original problem would be updating
the reference PolicyGenTemplate.
/cc @imiller0


imiller0 (approved):
/lgtm

PR #1258, opened 2022-09-13
CNF-5287: ztp: add doc for using ACM policy generator

Missxiaoguo:


ztp/gitops-subscriptions/ACMPolicyGeneratorIntergration.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+The annotation "ran.openshift.io/ztp-deploy-wave" is required to be added in the policy for [TALM](https://github.com/openshift-kni/cluster-group-upgrades-operator) to know the order of policies to be applied/checked. This can be done via `policyDefaults.policyAnnotations` as shown above.
+
+* Add `extended-du-validator.yaml` in the `kustomization.yaml` file as a generator.
+
+
+See [ACM PolicyGenerator reference YAML](https://github.com/stolostron/policy-generator-plugin/blob/main/docs/policygenerator-reference.yaml) for the full structure of `PolicyGenerator`. 
+
+### Integrating with Openshift GitOps(ArgoCD)
```
imiller0, line 67: This section and the following section detail two different methods of integrating ACM PolicyGenerator into the GitOps ZTP flow that we define. I think it would be helpful to put both into one section with a short introduction paragraph that explains that there are two methods, both will work equivalently, and if there is a recommendation of which one is preferred we should give some rationale for that recommendation.

ztp/gitops-subscriptions/ACMPolicyGeneratorIntergration.md
```diff
@@ -0,0 +1,183 @@
...
+As ACM Policy Generator is not preinstalled in the Openshift GitOps container image, these steps describes how to do integrate it with Openshift GitOps.
+
+**Requirements**:
+1. Advanced Cluster Management(ACM) operator v2.4+ installed on the hub cluster
+1. Redhat Openshift GitOps operator v1.4/1.5 installed and patched with [ZTP container](argocd/deployment/argocd-openshift-gitops-patch.json) on the hub cluster
+
+**Steps**:
```
imiller0, line 74: Are these steps also documented by ACM and/or upstream OCM project? If so, can we link to their documentation steps?

ztp/gitops-subscriptions/ACMPolicyGeneratorIntergration.md
```diff
@@ -0,0 +1,183 @@
...
+oc edit appprojects.argoproj.io -n openshift-gitops policy-app-project
+
+spec:
+  namespaceResourceWhitelist:
+  - group: policy.open-cluster-management.io/v1
+    kind: PolicyGenerator
+```
+
+3. Push the prepared custom source files and `PolicyGenerator` yamls to the Git respository, ArgoCD will detect the changes and generate the policies.
+Note: you could push to the same Git directory as the current Policy app references to or a new directory and create a new Argo application with that directory to roll out the changes.
```
imiller0, line 126: What is being referred to by "current Policy app references". Does this mean the current PolicyGenTemplate yaml files handled by the ztp-site-generator version of the plugin? If so, maybe clarify?

ztp/gitops-subscriptions/ACMPolicyGeneratorIntergration.md
```diff
@@ -0,0 +1,183 @@
...
+kind: LocalVolume
+metadata:
+  name: fs
+  namespace: openshift-local-storage
+status:
+  conditions:
+  - status: "True"
+    type: Available
+```
+Note: the validation CR can only contain `status`.
```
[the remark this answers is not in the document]
Missxiaoguo: Yes, you are right

ztp/gitops-subscriptions/ACMPolicyGeneratorIntergration.md
```diff
@@ -0,0 +1,183 @@
...
+* Create a ACM Policy Generator configuration file called `extended-du-validator.yaml`
+```
+apiVersion: policy.open-cluster-management.io/v1
+kind: PolicyGenerator
+metadata:
+  name: extended-du-validator
+policyDefaults:
+  policyAnnotations: {"ran.openshift.io/ztp-deploy-wave": "10000"}
+  evaluationInterval:
+    compliant: 24h
```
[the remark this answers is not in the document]
Missxiaoguo: Sure.

ztp/gitops-subscriptions/ACMPolicyGeneratorIntergration.md
```diff
@@ -0,0 +1,183 @@
...
+    placementRulePath: validator-placementrule.yaml
+  remediationAction: inform
+  complianceType: musthave
+policies:
+- name: "extended-du-validator-policy"
+  manifests:
+    - path: custom_source_crs/validationCRs/
+```
+
+The annotation "ran.openshift.io/ztp-deploy-wave" is required to be added in the policy for [TALM](https://github.com/openshift-kni/cluster-group-upgrades-operator) to know the order of policies to be applied/checked. This can be done via `policyDefaults.policyAnnotations` as shown above.
```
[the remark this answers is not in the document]
Missxiaoguo: Good point

ztp/gitops-subscriptions/ACMPolicyGeneratorIntergration.md
Missxiaoguo: The ACM official doc has this mentioned but doesn't have a full list of clear steps for all the resources need to be updated and the user can not directly use the patch provided in their example. The patch example I put down below is compability with the argocd setting we did for ztp policygenerator.

Missxiaoguo:
Have addressed the comments

imiller0 (approved):
/lgtm

PR #1261, opened 2022-09-15
ztp: OCPBUGS-217: set PTP daemonsetNodeSelector

vitus133:
This commit provides reference for overriding PTP daemonsetNodeSelector to "worker" instead of the default "master" predefined in the source-crs. The change is recommended on SNOs configured with PTP event notification for forward compatibility with possible SNO expansion. When the default setting is left intact, then in case of an SNO expansion with one or more workers, PTP operator
would not create linuxptp-daemon containers on the worker node(s). Any attempt to change the daemonsetNodeSelector will result in ptp daemon restart and time synchronization loss.

/assign @imiller0 
/cc @serngawy


imiller0 (approved):
/lgtm

PR #1262, opened 2022-09-15
ztp: OCPBUGS-239: set SR-IOV node selector

vitus133:
This commit provides reference for overriding SR-IOV configDaemonNodeSelector to "worker" instead of the default "master" predefined in the source-crs. The change is recommended on SNOs for forward compatibility with possible SNO expansion. When the default setting is left intact, then in case of an SNO expansion with one or more workers, SR-IOV operator would not create daemon containers on the worker node(s). Any attempt to change the configDaemonNodeSelector will result in daemon restart and possible downtime.

/assign @imiller0 
/cc @serngawy

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
@@ -56,7 +56,14 @@ spec:
           phc2sysOpts: "-a -r -n 24"
     - fileName: SriovOperatorConfig.yaml
       policyName: "config-policy"
+      # For existing clusters with node selector set as "master", 
+      # change the complianceType to "mustonlyhave".
+      # After complying with the policy, the complianceType can 
+      # be reverted to "musthave"
+      complianceType: musthave
       spec:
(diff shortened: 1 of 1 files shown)


ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -56,7 +56,10 @@ spec:
           phc2sysOpts: "-a -r -n 24"
     - fileName: SriovOperatorConfig.yaml
       policyName: "config-policy"
+      complianceType: mustonlyhave
```
imiller0, line 59: A question on the "mustonlyhave" part of this. Is it required to set the proper role for new clusters, or was it added to handle existing clusters which may have a conflicting role (eg master)? 
In the past some configuration (I believe OperatorGroup) caused issues with high CPU use when we used "mustonlyhave" as the compliance type due to a battle between multiple controllers setting annotations on the CR. Can we leave this as "complianceType: musthave" and add a note that existing clusters need to use mustonlyhave to remove the old annotation and then switch to musthave?

imiller0 (approved):
/lgtm

PR #1264, opened 2022-09-16
ztp: Add an option to enable/disable inspection

tliu2021:


ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -158,6 +163,10 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
 						clusterCRs = append(clusterCRs, instantiatedCR)
 					}
 				} else {
+					if kind == "BareMetalHost" {
+						// Ironic inspection is enabled by default, delete the inspect annotation if it is not disabled
```
imiller0, line 167: Is Ironic inspection disabled by default in all versions of ACM/BMO that we expect to run on (2.5, 2.6, 2.7+)? Or does the new "inspect" option only apply to ACM 2.6+ or 2.7+?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -158,6 +163,10 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
 						clusterCRs = append(clusterCRs, instantiatedCR)
 					}
 				} else {
+					if kind == "BareMetalHost" {
```
imiller0, line 166: Why is this if inside the else instead of else if kind == "BareMetalHost" like HostFirmwareSettings and NMStateConfig?

ztp/ran-crd/site-config-crd.yaml
```diff
@@ -393,3 +393,10 @@ spec:
                                       name:
                                         description: nic name used in the yaml, which relates 1:1 to the mac address. Name in REST API logicalNICName
                                         type: string
+                            inspect:
```
imiller0, line 396: Can we use something more descriptive than "inspect". For someone who is learning about deploying sites they will ask the question "what am I inspecting on the node?". Maybe bmcInspect, baremetalInspect, ???

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
tliu2021: The Ironic inspection is enabled by default in BMO, and it can be disabled by adding "inspect.metal3.io: disabled" annotation to BMH.  This was introduced for ZTP/AI deployment (i.e. ACM2.5, ACM2.6). In ACM 2.7 + and ocp version 4.11+, the converged flow will be the default behavior and the disable annotation should not be present.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
tliu2021: Because 'clusterCRs = append(clusterCRs, instantiatedCR)' is common under else, otherwise it is called under else if and else.

ztp/ran-crd/site-config-crd.yaml
tliu2021: It refers to the Ironic 'inspection' process. I will change it to  'ironicInspect'.

tliu2021: Done

imiller0 (approved):
/lgtm

PR #1267, opened 2022-09-19
ztp: OCPBUGS-1365: Update kargs in reference performance profile CR

browsell:


imiller0 (approved):
/lgtm
This will update the generated Policies for users when they upgrade to a container version with this change. Users who want these settings will need to add them to their PolicyGenTemplate as override content.

PR #1271, opened 2022-09-22
Bug 2116547: Create new ptpconfig for Dual CVL

josephdrichard:


nishant-parekh:
@josephdrichard add `ztp:` prefix to your commit message for the prow/ci failure

PR #1276, opened 2022-09-26
Enable adding additional network in network.operator config

karampok:
By default all telco clusters (SNO/MNO) will have to create additional networks, ideally through the network.operator config (CNO).

### DIFF
--- a/ztp/source-crs/AddAdditionalNetworks.yaml
+++ b/ztp/source-crs/AddAdditionalNetworks.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: operator.openshift.io/v1
+kind: Network
+metadata:
+  name: cluster
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "10"
+spec:
+  additionalNetworks:
+  # - name: ran-mh-macvlan
+  #   namespace: default
+  #   rawCNIConfig: '{ "cniVersion": "0.3.1", "name": "macvlan", "plugins": [ {"type": "macvlan","master": "vlan223", "mtu": 1500, "mode": "bridge", "ipam": { "type": "whereabouts", "range": "2520:41:0:2106::/64","gateway": "2520:41:0:2106::1" } }, {"type":"tuning","sysctl":{"net.ipv6.conf.IFNAME.accept_ra":"0"}} ]}'
+  #   type: Raw


ztp/source-crs/AddAdditionalNetworks.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: operator.openshift.io/v1
+kind: Network
+metadata:
+  name: cluster
```
imiller0, line 5: Hi @karampok Thanks for this PR.
The Network CR is already included in the source-crs ([DisableSnoNetworkDiag.yaml](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/source-crs/DisableSnoNetworkDiag.yaml#L2-L4)). Although the name is not generic you can add any content you want by using the overlay feature of the PolicyGenTemplate (PGT). 

Updating the reference group [PGT entry](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml#L97-L98) for DisableSnoNetworkDiag.yaml like this should be what you need:
```
    - fileName: DisableSnoNetworkDiag.yaml
      policyName: "config-policy"
      spec:
        additionalNetworks:
        - name: ran-mh-macvlan
          namespace: default
          rawCNIConfig: '{....}'
          type: Raw
```
This will add your additionalNetworks entry to the default content from the source-cr (you can also overwrite the `disableNetworkDiagnostics` entry if desired/necessary).
Hope this helps!

djuran:
Slightly off-topic but can we even have two source-crs referring to the same CR?

If that is the case, that would put a requirement on the end user to never include those two for the same cluster, wouldn't it?
Which would be a bit tricky to explain from a UX point-of-view

imiller0:
> Slightly off-topic but can we even have two source-crs referring to the same CR?
Short answer, yes, but it is a nuanced answer...
The source-crs is essentially a library of CRs that can be referenced in the PolicyGenTemplate. If you have 2 source-crs with the same CR in them you need to manage that potential conflict at the PolicyGenTemplate level. You can either:
- Use only one of them in your PolicyGenTemplate (in which case the other is simply unused/ignored)
- Use both but change the `metadata.name` so that they actually manage two different CRs on the target cluster(s)

> If that is the case, that would put a requirement on the end user to never include those two for the same cluster, wouldn't it? Which would be a bit tricky to explain from a UX point-of-view

PR #1293, opened 2022-10-11
OCPBUGS-1594: Fec tests: do not clean in discovery mode

fedepaol:
Discovery mode skips the tests, so we should not try to delete any config in case discovery mode is enabled.


fedepaol:
Failure in mutli network policy, not related to this fix.
/override ci/prow/e2e-gcp-ovn

PR #1298, opened 2022-10-19
Bug 2094046: cnf-tests: verify minimal CPUs requirement for cyclictest test

Tal-or:
We should handle the amount of minimal CPUs for cyclictest test more carefully. When HT disabled cyclictest needs 2 CPUs.

But when HT enabled oslat needs at least 4 CPUs:
One CPU should be used to run cyclictest's main thread. The second is the sibling of the first one, which should be excluded from the list of the tested CPUs, Because it might cause to false spikes (noisy-neighbor issue). The third one is the actual CPU to be tested, but due to SMT alignment restriction we need its sibling too. Four in total.

### DIFF
--- a/cnf-tests/pod-utils/cyclictest-runner/main.go
+++ b/cnf-tests/pod-utils/cyclictest-runner/main.go
@@ -31,9 +31,22 @@ func main() {
 	}
 
 	mainThreadCPUs := selfCPUs.ToSlice()[0]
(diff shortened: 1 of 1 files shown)


Tal-or:
@fromanirh Could you please review

fromanirh (commented):
/lgtm

Tal-or:
/test e2e-gcp-ovn
--


<br class="Apple-interchange-newline">

PR #1301, opened 2022-10-23
OCPBUGS-2618: cnf-tests: run cnf-tests on properly isolated cpus

Tal-or:
We need to replace the `exec.Command` with `execve` syscall. `exec.Command` is using fork, which means that it leaves the parent process alive, and the parent process affined to the entire CPU set which might cause latency spikes. `execve` OTOH is replacing the parent (the runner process) with the actual latency test process (and essentially removing the parent process). This way we guarantee that no other process is running on the tested CPU set beside the test's process itself.


Tal-or:
/hold
Need some testing

Tal-or:
/cc
@MarSik @browsell

cnf-tests/pod-utils/hwlatdetect-runner/main.go
```diff
@@ -36,7 +36,7 @@ func main() {
 	}
 
 	hwlatdetectArgs := []string{
-		hwlatdetectBinary,
+		"hwlatdetect",
```
fromanirh, line 39: why this replace?

fromanirh (commented):
/approve

fromanirh (commented):
didn't review the code carefully, but the issue is real and we want to address it. Replacing with exec seems like a good direction. self-pinning the parent before the test start could have been another option.

cnf-tests/pod-utils/hwlatdetect-runner/main.go
Tal-or: We pass the binary path to the `exec` system call https://github.com/Tal-or/cnf-features-deploy/blob/cnftests_properly_isolate_cpus/cnf-tests/pod-utils/hwlatdetect-runner/main.go#L51 

It will detect automatically that it's a python program due to the shebang

fromanirh: uhm, looking around it seems we're doing like others args are constructed - this instance used to be the odd one.
This comments applies to all the instances then. Why are we setting `argv[0] != argv0` all around?
Should not be a big deal 99% of the cases, but when causes bugs it will be quite hard to figure out.

Tal-or:
/hold
depends on: https://github.com/openshift/cluster-node-tuning-operator/pull/499

cnf-tests/pod-utils/hwlatdetect-runner/main.go
Tal-or: According to your comment `argv[0]` should contain the filename,  and this is exactly what we're doing here.
Unless filename implicitly means full path of filename

fromanirh: yes, I meant the full absolute path. It's probably OK (or at very least not more broken then before) but AFAICT we're still drifting out of the expected behavior.

Tal-or:
> the failure in the `image` lane seems real (e.g. not a flake)

Looking good now - CI happy

fromanirh:
> > the failure in the `image` lane seems real (e.g. not a flake)
> 
> Looking good now - CI happy

weird, but I'm not complaining!

fromanirh:
/approve
I'm still unconvinced that using the unqualified (non absolute path) as first argument of `Exec` is the right thing to do, but OTOH doesn't look bad enough to block this PR any further.

PR #1302, opened 2022-10-24
vendor: bump NTO dependencies 20221024

Tal-or:
Signed-off-by: Talor Itzhak <titzhak@redhat.com>

PR #1305, opened 2022-10-27
infra: Pin packages for go version dependency

imiller0:
Working to fix CI failures. Many thanks to @fromanirh for the fix.  

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -28,8 +28,8 @@ RUN mkdir ~/bin && \
     eval "$(gimme $GOVERSION)" && \
     cat $GIMME_ENV >> $HOME/.bashrc && \
     # get required golang tools and OC client
-    go get github.com/onsi/ginkgo/ginkgo && \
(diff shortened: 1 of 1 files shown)


fromanirh:
/lgtm
/approve
We need to review the prow config setting. It seems we're NOT testing HEAD+PR applied, but current HEAD, hence lanes are still failing even if this PR works (tested locally). This behavior could be something we want or something that sneaked in unintentionally.

fromanirh:
I recommend to merge this PR even with lanes failing and monitor right after with any other test PR. The next one after this merged should work. If it is not the case, send another fix or revert this one.

fromanirh:
> I recommend to merge this PR even with lanes failing and monitor right after with any other test PR. The next one after this merged should work. If it is not the case, send another fix or revert this one.

This PR is a good testbed: https://github.com/openshift-kni/cnf-features-deploy/pull/1302

imiller0:
To unblock current CI we are going to try pinning ginkgo and gomega (this PR) in the short term while looking at the bump to go 1.18.

PR #1312, opened 2022-11-04
Add a holdInstallation flag to SiteConfig

leo8a:
Since for Telco ZTP we're targeting e2e deployments with no user action, I just set the default value to false.

Signed-off-by: Leo Ochoa [lochoa@redhat.com](mailto:lochoa@redhat.com)

### DIFF
--- a/ztp/ran-crd/site-config-crd.yaml
+++ b/ztp/ran-crd/site-config-crd.yaml
@@ -95,6 +95,13 @@ spec:
                           for subdomains, some resource tagging, and other instances
                           where a friendly name for the cluster is useful.
                         type: string
+                      holdInstallation:
+                        description: | 
+                          HoldInstallation will prevent installation from happening when true. Inspection and 
+                          validation will proceed as usual, but once the RequirementsMet condition is true, 
+                          installation will not begin until this field is set to false.
+                        type: boolean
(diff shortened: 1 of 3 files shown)


leo8a:
pushed again with some fixes to other nits on the spec

imiller0:
/hold
Adding a hold until the 4.12 release branch is created.

imiller0 (commented):
@leo8a Thanks for adding this functionality! Is there a particular use case where this is needed?

Can you add to the PR CI validation tests (here is [one example](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go#L214))

leo8a:
> Thanks for adding this functionality! Is there a particular use case where this is needed?

Thanks for taking a look. Although Telco ZTP expects this holdInstallation as false (hence the default assignment), there might be some times that the admin/user might want to pause the cluster installation for troubleshooting any issue with the deployment before continuing till the end.

If holdInstallation is set to true, the day1 installation flow is paused and can be resumed at any moment. More info on [PR#2293](https://github.com/openshift/assisted-service/pull/2293).

serngawy (approved):

PR #1314, opened 2022-11-07
vendor: Using ptp reporter in ginkgo's TestTest function.

greyerof:
The PTP reporter is the default ginkgo v1's junit reporter plus a backported version of the v2's AddReportEntry() function. When ginkgo is finally migrated to v2 in this project, just remove it. More info here: https://github.com/openshift/ptp-operator/pull/253

Also, upgraded ptp-operator dependencies.

Signed-off-by: Gonzalo Reyero Ferreras <greyerof@redhat.com>


edcdavid (approved):

fedepaol:
If I get this right, it will affect the generated report only if the tests opt-in right?

greyerof:
@fedepaol correct. There's already one [TC from the ptp-operator](https://github.com/openshift/ptp-operator/blob/30605b753ac2da9013c0650eeae12aa4980902a6/test/conformance/ptp/ptp.go#L218)'s conformance test suite that's using the "backported" `AddReportEntry()` : the system-out xml node for that TC will have those entries appended, but the report for the other TCs should remain exactly the same.

josephdrichard (approved):
/lgtm

greyerof:
I've rebased this PR. Maybe the dpdk failures have been fixed by #1320

SchSeba:
/lgtm
/approve

I check the reporter and there is no change in the output with this new feature introduced so we should be good.

SchSeba:
failed test is not related to this change.

PR #1317, opened 2022-11-09
[release-4.11] cnf-tests: Filter tests by features suites

sabinaaledort:
Adding brackets to focus on tests by feature. If FEATURES var is used for ginkgo focus we should do the filtering by suites ([feature]). Without the brackets we may run tests that have a feature name in their description while they belong to a feature suite that we don't want to currently test.


SchSeba:
we are not going to use the to select features any more. instead the user will use a combination of focus + skip in ginkgo

PR #1320, opened 2022-11-14
s2i: update python version to support meson

SchSeba:


SchSeba:
the failing test is not related to this change

/override ci/prow/e2e-gcp-ovn

PR #1322, opened 2022-11-16
cnf-tests: avoid run multinetpol tests with sriov

zeeke:
If a user install `[sriov]` feature and run related tests, suite should not fail for a not deployed feature. Removing `[sriov]` label from multinetworkpolicy tests should do the job.

Signed-off-by: Andrea Panattoni <apanatto@redhat.com>


SchSeba:
/hold

I think we should close this one if the use want to run only sriov tests it should also add skip=multinetworkpolicy

zeeke:
@SchSeba If we force the user to add the skip parameter, we should at least document it, but it would be an intricate paragraph.

With this PR, the following scenarios can happen, and cells contain the multinetworkpolicy behaviors.

| ↓`sriov`\\`multinetworkpolicy`→| deployed|NOT deployed|
|---|---|---|
|**deployed**|tests run and must pass|tests are skipped by the runner, as multinetworpolicy feature is not in FEATURE env var|
|**NOT deployed**|tests are skipped (via [this](https://github.com/openshift-kni/cnf-features-deploy/blob/a9a14378e412a1da6c6bc519856841aa541a31f2/cnf-tests/testsuites/e2esuite/multinetworkpolicy/multinetworkpolicy_sriov.go#L144)). BTW, if we will add some macvlan test on this feature, we will make them run in cases like this|tests don't run, as they are not selected|

I think this is more or less what happens with SCTP tests suite. right?

The main risk is having a green CI with skipped tests, and I prefer to tackle that problem in another way (e.g. having a test skip list). 

Does it make sense?

SchSeba:
sounds good :)

/hold cancel
/lgtm
/approve

SchSeba:
failed test is not related to this change

/override ci/prow/e2e-gcp-ovn

PR #1335, opened 2022-11-29
Update topologyPolicy, delete userLevelNetworking

borball:
To align with what customer is using


dav1x:
Another customer I'm working with is also using
```
topologyPolicy: single-numa-node
```

PR #1337, opened 2022-11-30
fix for ptp interfaces test failure

edcdavid:
This is to fix a failure with the ptp node device interface test due to Virtual functions not removed from the list of potential PTP interfaces


SchSeba:
error is not related to this PR

/override ci/prow/e2e-gcp-ovn

PR #1341, opened 2022-12-01
OCPBUGS-4378: Fix the hwlatdetect default window duration

MarSik:
The window length and width determine how often will hwlat detector check for cpu time unaccounted for.

Our opinionated default should be 95% of every second.

There was a typo in the code that reduced that to 9.5% as it was scanning for 0.95 s every 10 s. This patch fixes that.

### DIFF
--- a/cnf-tests/pod-utils/hwlatdetect-runner/main.go
+++ b/cnf-tests/pod-utils/hwlatdetect-runner/main.go
@@ -20,8 +20,8 @@ func main() {
 	threshold := flag.Int("threshold", 20, " value above which is considered an hardware latency")
 	hardlimit := flag.Int("hardlimit", 20, " value above which the test is considered to fail")
 	duration := flag.String("duration", "15s", "total time to test for hardware latency: <n>{smdw}")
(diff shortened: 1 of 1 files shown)


Tal-or (commented):
/lgtm
Out of curiosity, what is the underline in between every three zeros?

MarSik:
> /lgtm Out of curiosity, what is the underline in between every three zeros?

Separator to make the number more readable.

It is allowed by the golang syntax here https://go.dev/ref/spec#Integer_literals

"For readability, an underscore character _ may appear after a base prefix or between successive digits; such underscores do not change the literal's value."

MarSik:
@fromanirh @SchSeba @yanirq Hey, it seems 4.13 CI is broken atm, this is a small change and could be forced, however I am not too happy about that. On the other hand I would like to get the backports running. Any idea?

yanirq:
> @fromanirh @SchSeba @yanirq Hey, it seems 4.13 CI is broken atm, this is a small change and could be forced, however I am not too happy about that. On the other hand I would like to get the backports running. Any idea?

@MarSik  I would examine ci/prow/ci  failure which is complaining about a failing script (maybe try running it locally: verify-commits.sh)

PR #1356, opened 2022-12-13
infra: Add feature-deploy-on-ci task to Makefile for U/S CI

sshnaidm:
One of jobs will need to deploy features, but not to run functest. It will run origin tests probably. Add a task that prepare for it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,9 +44,11 @@ wait-and-validate:
 	@echo "Validating"
 	SKIP_TESTS="$(SKIP_TESTS)" DONT_FOCUS=true TEST_SUITES="validationsuite" hack/run-functests.sh
 
-functests-on-ci: setup-test-cluster setup-build-index-image feature-deploy feature-wait functests
+functests-on-ci: feature-deploy-on-ci functests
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -48,6 +48,8 @@ functests-on-ci: setup-test-cluster setup-build-index-image feature-deploy featu
 
 functests-on-ci-no-index-build: setup-test-cluster feature-deploy feature-wait functests
 
+feature-deploy-on-ci: setup-test-cluster setup-build-index-image feature-deploy feature-wait
```
SchSeba, line 51: we have `functests-on-ci` please change that one to 

functests-on-ci: feature-deploy-on-ci functests

so we can reuse

nishant-parekh (changes requested):
@sshnaidm take a look at this script.. https://github.com/openshift-kni/cnf-features-deploy/blob/master/.githooks/commit-msg
I think you need to update the commit message to get past the prow/ci issue you are seeing..

SchSeba (approved):
/lgtm
/approve

PR #1357, opened 2022-12-19
[release-4.11] [manual] OCPBUGS-5020: cnf-tests: run cnf-tests on properly isolated cpus

Tal-or:
We need to replace the `exec.Command` with `execve` syscall. `exec.Command` is using fork, which means that it leaves the parent process alive, and the parent process affined to the entire CPU set which might cause latency spikes. `execve` OTOH is replacing the parent (the runner process) with the actual latency test process (and essentially removing the parent process). This way we guarantee that no other process is running on the tested CPU set beside the test's process itself.


ffromani:

failures looks unrelated, and the lane was broken since a lot of time already

cnf-tests/TESTLIST.md
```diff
@@ -134,7 +134,12 @@ The cnf tests instrument each different feature required by CNF. Following, a de
 
 | Test Name | Description |
 | -- | ----------- |
+| [performance]  Should have kernel param rcutree.kthread | XXXXXXXX |
```
ffromani, line 137: these XXXXXX (there's few more below) don't look right, please check

Tal-or: This generated by default when I'm running the relevant make target.
Changing the XXX cause the CI to fail

ffromani: @SchSeba any insights here?

yuvalk: it's because the tests are missing a description 

AFAIK, just fill in the description in the json file.
and you'll also get the new tests documented :-)

SchSeba: that is right you need to update the json here for example
https://github.com/openshift-kni/cnf-features-deploy/pull/1357/files#diff-cfad2c15f9822954c50b9f3f08ce93166ba405eb510950e86e4c045a4eb22fa9R37
and generate the docs

Tal-or: Maybe I did it the other way around and updated the docs themselves instead of the json. I'll give it a try

Tal-or:
@SchSeba, @ffromani  gentle reminder, please review when possible

ffromani:
/approve
/lgtm

updates looks good, thanks!

ffromani:
right, I'm not an approver in 4.11 (alias not updated :) )

PR #1360, opened 2022-12-22
cnf-tests: psa: don't bail out if labels are missing

fromanirh:
If the namespace we want to add PSA labels to has no labels at all, the current code adds the PSA labels and drops out, without Update()ing back the namespace object in the apiserver. This seems wrong, but probably never triggers because k8s namespace already have labels provided by the system.

Nevertheless, let's make sure we update back also in this flow.


SchSeba:
Thanks!

/lgtm
/approve

PR #1368, opened 2023-01-02
diskPartition feature not implemented on 4.10

jgato:
If I am not wrong diskPartition is only available from 4.10. So, we should not have test code about it.


lack:
Strictly speaking this code doesn't ever actually RUN anything; it's just example YAML to check that the SiteConfig engine converts the input into the right output.

It's also fine to remove, but since this is in an older branch that's pretty much locked-down, I'm not sure it's worth going through a bug report and cycling through to get this cleaned up.

That said, the change looks
/ok-to-test

Issue #1343, opened 2022-12-02
When you generate the example SNO SiteConfig CR, the sample cpuset is "0-1,52-53". This is misleading


aireilly:
When you generate `out/argocd/example/siteconfig/example-sno.yaml` using the ztp-site-generate` container, the example cpuset is `cpuset: "0-1,52-53"` This is misleading. best practice suggests the CPU set should be contiguous.

cc @pixelsoccupied

lack:
I think they should be contiguous, but it should also not split up NUMA pairs.  So in this example, if `0` and `52` are pairs and `1` and `53` are pairs, `0-1,52-53` is right?

aireilly:
Ah ok. This came from a conversation with @pixelsoccupied if this is correct let's close the issue.

aireilly:
We had discussed something like 0-3 as being a better example.

pixelsoccupied:
@lack I did not think of NUMA aspect actually. But strictly speaking in-terms of Workload partitioning -- we need to match the `cpuset` and `reserved` to get systems and apps pin to the same set. But if we spread it around 0-1,52-53...crio shows correct pinning up but PerformanceProfile confirmed (and experiments on my side) that this is an undefined behaviour.

e.g `pgrep ovn | while read i; do taskset -cp $i; done` good -- but `pgrep systemd | while read i; do taskset -cp $i; done` is undefined.

pixelsoccupied:
Oops I think I got it mixed up -- 

Just to clear up "undefined" part --if PP CR would looks something like this
```
        cpu:
          # These must be tailored for the specific hardware platform
          isolated: "7-103"
          reserved: "0-3"
``` 
Here leaving out 4,5,6 cpu is considered an unspecified behavior (in my systemd.cpu_affinity was saying 0-6 even though CR is 0-3)

Official docs [ref](https://docs.openshift.com/container-platform/4.10/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html#cnf-cpu-infra-container_cnf-master)
```
The reserved and isolated CPU pools must not overlap and together must span all available cores in the worker node.
```

@aireilly we can keep `0-1,52-53` (of course matching both `reserved` and `cpuset` is still applicable).

I'll make some changes in our repo doc to document this before we close this.

lack:
Ah yes!  Leaving a gap is bad.

PR #1376, opened 2023-01-09
ztp: document core reduction configs

pixelsoccupied:
document core reduction configs and close [issue](https://github.com/openshift-kni/cnf-features-deploy/issues/1343) 
/cc @lack  
/cc @imiller0

### DIFF
--- a/ztp/gitops-subscriptions/argocd/WorkloadPartitioning.md
+++ b/ztp/gitops-subscriptions/argocd/WorkloadPartitioning.md
@@ -1,6 +1,6 @@
 # Workload Partitioning Feature
 
-Workload partitioning is the feature that allows the platform to dedicate specific CPUSets to customer workloads and platform workloads (~2 cores for platform and rest for customers). This feature **must** be activated at install time of a cluster, once active that cluster is locked into functioning with this feature on.
+Workload partitioning allows the platform to dedicate specific CPUSets to customer workloads and platform workloads (~2 cores for platform and the rest for customers). This feature **must** be activated at install time of a cluster, once active that cluster is locked into functioning with this feature on.
 
 ## Part 1 - Configure PolicyGenTemplate
 - This will pin all the host level services such as systemd and crio.


ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,32 @@
...
+  2. pinning done with Workload Partitioning and output must match `cpuset`
+     - `pgrep ovn | while read i; do taskset -cp $i; done`
+  3.  Values after `..current affinity list:` for both `systemd` and `ovn` must the match
+- Look for the correct config under `cat /proc/cmdline` updated with new info such `systemd.cpu_affinity=xx`. `xx` should match the value in `reserved` from `PolicyGenTemplate`
+- Look for the correct config under `cat /etc/crio/crio.conf.d/01-workload-partitioning` and make sure it matches `cpuset` in SiteConfig.
+
+## Additional readings
+- [Management Workload Partitioning](https://github.com/openshift/enhancements/blob/5c92a52b27580c96eaf7ea3af79fef35463b3e2a/enhancements/workload-partitioning/management-workload-partitioning.md)
+- [Performance Profile](https://docs.openshift.com/container-platform/4.11/scalability_and_performance/cnf-low-latency-tuning.html)
+- [Workload Partitioning in  SNO](https://docs.openshift.com/container-platform/4.11/scalability_and_performance/sno-du-enabling-workload-partitioning-on-single-node-openshift.html)
```
lack, line 32:
```suggestion
- [Workload Partitioning in SNO](https://docs.openshift.com/container-platform/4.11/scalability_and_performance/sno-du-enabling-workload-partitioning-on-single-node-openshift.html)
```

lack (commented):
Is it worth adding some note that the workload partitioning is normally done at install time, whereas the performance profile is a day-2 operation?  They're mentioned in the opposite order since it makes sense to write the performance profile first, but the installation order is reversed...

pixelsoccupied:
> Is it worth adding some note that the workload partitioning is normally done at install time, whereas the performance profile is a day-2 operation? They're mentioned in the opposite order since it makes sense to write the performance profile first, but the installation order is reversed...

Slight reworded and added a new step to clarify that the `apply` of CRs are done in reverse order

bartwensley:
Just wondering what the purpose of this document is vs. the existing user documentation here:
https://access.redhat.com/documentation/en-us/openshift_container_platform/4.11/html/scalability_and_performance/sno-du-enabling-workload-partitioning-on-single-node-openshift
and here:
https://access.redhat.com/documentation/en-us/openshift_container_platform/4.11/html/scalability_and_performance/cnf-low-latency-tuning

Seems like duplication?

pixelsoccupied:
> Just wondering what the purpose of this document is vs. the existing user documentation here: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.11/html/scalability_and_performance/sno-du-enabling-workload-partitioning-on-single-node-openshift and here: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.11/html/scalability_and_performance/cnf-low-latency-tuning
> 
> Seems like duplication?

It's meant to be an upstream version of the how to configure for core reduction in ZTP context with additional verification steps/tools and basically just consolidating multiple docs about this topic in one place. We have similar docs for e.g Wave, SecureBoot, ImageRegistry and so on.

bartwensley:
> It's meant to be a downstream version of the how to configure for core reduction in ZTP context with additional verification steps/tools and basically just consolidating multiple docs about this topic in one place. We have similar docs for e.g Wave, SecureBoot, ImageRegistry and so on.

IMO this is duplicating info already available at docs.openshift.com and should be avoided. Long term, it will lead to one or the other getting out of date. If we feel the PGTs are clear enough, then I'd prefer to see that addressed by adding more comments to the PGTs.

lack:
I generally agree that duplication is bad. But in this case, we should "normally" have upstream docs to start with, then the official openshift docwriters can look at our upstream docs and craft the downstream docs based on the content.

We're playing catch-up right now because we didn't always do a good job of doing upstream docs when we did these features originally, but once they're there the normal steps will be easier: (1) make our changes upstream and then (2) downstream docs are based on those changes

ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,37 @@
+# Core Reduction with Workload Partitioning and Performance Profile
```
eggfoobar, line 1: Should we mention 'core reduction' at all here? I think we've historically used that term to mean the effort of reducing the CPU usage for all platform services, workload partitioning is a feature that can be used with out an intent to reduce actual CPU usage. Maybe we just call this `# Workload Partitioning` to keep it simple on the intent of this section?

ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,37 @@
...
+
+Maximum number of cores can be allocated for customer workload only by pinning host and cluster workloads to a small number of cores. This is known as "core reduction" and be achieved by correctly configuring Workload Partitioning and Performance Profile.
```
eggfoobar, line 3: Same note here, technically the act of segregating platform workloads from customer workloads on different CPUSets is workload partitioning not core reduction. So something like this might be a good summary of what we want to cover in this section, thoughts?
```
Workload partitioning is the feature that allows the platform to dedicate specific CPUSets to customer workloads and platform workloads. This feature **must** be activated at install time of a cluster, once active that cluster is locked into functioning with this feature on.
```

ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Part 1 - Configure Performance Profile
```
eggfoobar, line 5: Maybe we should add a little intro blurb here that says the Performance Profile is the one that handles setting up services that run outside of the cluster for workload partitioning, things like `crio, systemd, kubelet`.

ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,37 @@
...
+- Configure this using `SiteConfig`, See example [here](https://github.com/openshift-kni/cnf-features-deploy/blob/82ff3617a5e69b47b1f8d8b5d4a8db7719ab4bb4/ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml#L59)
+- Set the value of `cpuset` the as the exact one used for `reserved` from `Part 1`
+
+## Part 3 - Apply configured SiteConfig and PolicyGenTemplate with GitOps
+The CRs are applied in reverse order. Note that PerformanceProfile is configured first (`Part 1`) for correctness and convenience, but it's expected to be applied as part of day-2 operations. To maximize core reduction benefits, both (SC and PGT) must be configured and applied eventually. 
+1. Apply SiteConfig
+2. Apply PolicyGenTemplate
+
+## Verification
+ssh into the node (e.g `oc debug node/myhost.rh.com`) and try:
```
eggfoobar, line 24: small nitpick, people tend to just copy paste when reading docs, maybe change this example from `oc debug node/myhost.rh.com` to `oc debug node/<NODE_NAME>`

ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,37 @@
...
+## Part 3 - Apply configured SiteConfig and PolicyGenTemplate with GitOps
+The CRs are applied in reverse order. Note that PerformanceProfile is configured first (`Part 1`) for correctness and convenience, but it's expected to be applied as part of day-2 operations. To maximize core reduction benefits, both (SC and PGT) must be configured and applied eventually. 
+1. Apply SiteConfig
+2. Apply PolicyGenTemplate
+
+## Verification
+ssh into the node (e.g `oc debug node/myhost.rh.com`) and try: 
+- Look for cpu pinning with `taskset`
+  1. pinning done with PerformanceProfile and output must match `reserved`
+     - `pgrep systemd | while read i; do taskset -cp $i; done`
```
eggfoobar, line 27: What do you think of the idea of being more elaborate with the example here, something like this should allow a one command printout of the outside process checks, not the cleanest regex but just a thought.
```
pgrep "systemd|crio|kubelet" | while read i; do echo "CPUSet $(taskset -cp $i | grep -Po '[0-9]+[-,]+[0-9]+') for process $(ps -p $i -o comm=)"; done
```

ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,37 @@
...
+  2. pinning done with Workload Partitioning and output must match `cpuset`
+     - `pgrep ovn | while read i; do taskset -cp $i; done`
```
eggfoobar, line 29: Same suggestion on this example as above, but with `ovn`.

ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,37 @@
...
+  3.  Values after `..current affinity list:` for both `systemd` and `ovn` must the match
```
eggfoobar, line 30: If we choose to add the more explicit command, we can update the wording here to reflect the print out customers would see.

ztp/gitops-subscriptions/argocd/CoreReduction.md
```diff
@@ -0,0 +1,37 @@
...
+- Look for the correct config under `cat /proc/cmdline` updated with new info such as `systemd.cpu_affinity=xx`. `xx` should match the value in `reserved` from `PolicyGenTemplate`
```
eggfoobar, line 31: Make sure we remind them that for these commands to work they must have run `chroot /host` prior in the debug container.

eggfoobar (commented):
Just some quick suggestions and nitpicks for this PR. 

In regard to the duplicate documentation conversation, I tend to agree that we should just redirect people to the official docs instead of adding another location with possible out of date information. However, if this is supposed to go upstream then I'm on the same mindset as @lack that we should have accurate information or at least make sure we point the upstream reader to the official docs to avoid confusion.

ztp/gitops-subscriptions/argocd/CoreReduction.md
pixelsoccupied: Yeah I have it down below `This will pin all the host level services such as systemd and crio.` but can move it up

ztp/gitops-subscriptions/argocd/CoreReduction.md
eggfoobar: Yeah, maybe we say the whole section is workload partitioning, and we make a change of `Part 1 - Configure Node Processes` and `Part 2 - Configure Cluster Processes`, something to that effect.

bartwensley:
> I generally agree that duplication is bad. But in this case, we should "normally" have upstream docs to start with, then the official openshift docwriters can look at our upstream docs and craft the downstream docs based on the content.
> 
> We're playing catch-up right now because we didn't always do a good job of doing upstream docs when we did these features originally, but once they're there the normal steps will be easier: (1) make our changes upstream and then (2) downstream docs are based on those changes

I'm still not understanding. Take workload partitioning for example - there are detailed docs here:
https://docs.openshift.com/container-platform/4.11/scalability_and_performance/ztp_far_edge/ztp-reference-cluster-configuration-for-vdu.html#ztp-sno-du-enabling-workload-partitioning_sno-configure-for-vdu

Aren't these public docs? What we are adding here is no where near as complete as the existing info in the above doc, so it isn't going to help our doc writers.

Also, if we were going to create new upstream docs for workload partitioning (for example), why would it be documented in our argocd folder?

aireilly:
Hi :wave:

I think this topic of upstream/downstream docs would be a good topic for the Platform Lifecycle working group meeting. I'll add it as an agenda item. 

FWIW, A number of RH teams are already working between Dev/QE and Docs teams to write/review/QE docs upstream first. It has a lot of benefits, but requires some effort up front.

Question: do we have users actually using the upstream software in this instance? If the primary user here is the internal RH developer we could maybe add a preface to these docs stating this and pointing users to the downstream docs where appropriate.

imiller0 (commented):
A couple notes, but this is good helpful information for users on being successful with deploying clusters using this methodology.

My take on the question of documentation here vs the official Red Hat documentation is that upstream (here) documentation of features/content that is relevant to this repo is a good thing. Any feature development/improvements should be accompanied by updates to these docs. That way the docs here reflect the latest feature set. I share the concern about duplicate content but hope that generally the upstream documentation is more feature/functionality focused and gets pulled into downstream documentation which includes additional solution/context content.

ztp/gitops-subscriptions/argocd/CoreReduction.md
pixelsoccupied: used your suggestion

ztp/gitops-subscriptions/argocd/WorkloadPartitioning.md
```diff
@@ -0,0 +1,37 @@
...
+  - look for `--reserved-cpu-count` when using Performance Profile Creator cli.
+- WARNING: PerformanceProfile's `reserved` and `isolated` must span ALL AVAILABLE CORES and not doing so will result in an undefined behaviour (generating it with tool above should take care of this behind-the-scenes)
+
+## Part 2 - Configure SiteConfig
+- Configure workload partitioning in `SiteConfig`, See example [here](https://github.com/openshift-kni/cnf-features-deploy/blob/82ff3617a5e69b47b1f8d8b5d4a8db7719ab4bb4/ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml#L59)
+- Set the value of `cpuset` the as the exact one used for `reserved` from `Part 1`
+
+## Part 3 - Apply configured SiteConfig and PolicyGenTemplate with GitOps
+The CRs are applied in reverse order. Note that PerformanceProfile is configured first (`Part 1`) for correctness and convenience, but it's expected to be applied as part of day-2 operations. To maximize core reduction benefits, both (SC and PGT) must be configured and applied eventually.
```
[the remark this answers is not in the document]
pixelsoccupied: I was to make it clear that although it's more convenient to configure first and get the right CPU numbers...SiteConfig is applied first. But yeah it's implied. Removing the this sentence

ztp/gitops-subscriptions/argocd/WorkloadPartitioning.md
```diff
@@ -0,0 +1,37 @@
+# Workload Partitioning Feature
+
+Maximum number of cores can be allocated for customer workload only by pinning host and cluster workloads to a small number of cores. This is known as "core reduction" and be achieved by correctly configuring Workload Partitioning and Performance Profile.
+
+## Part 1 - Configure PolicyGenTemplate
+- Update PerformanceProfile in `PolicyGenTemplate`. See example [here](https://github.com/openshift-kni/cnf-features-deploy/blob/82ff3617a5e69b47b1f8d8b5d4a8db7719ab4bb4/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml#L99).
+- In this context we will look at [`reserved`](https://github.com/openshift-kni/cnf-features-deploy/blob/82ff3617a5e69b47b1f8d8b5d4a8db7719ab4bb4/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml#L105)
+  - The value here will CPU core numbers and are dependent on the hardware.
```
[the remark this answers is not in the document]
pixelsoccupied: updated

ztp/gitops-subscriptions/argocd/WorkloadPartitioning.md
```diff
@@ -0,0 +1,38 @@
+# Workload Partitioning Feature
+
+Workload partitioning is the feature that allows the platform to dedicate specific CPUSets to customer workloads and platform workloads (~2 cores for platform and rest for customers). This feature **must** be activated at install time of a cluster, once active that cluster is locked into functioning with this feature on.
```
lack, line 3:
```suggestion
Workload partitioning allows the platform to dedicate specific CPUSets to customer workloads and platform workloads (~2 cores for platform and the rest for customers). This feature **must** be activated at install time of a cluster, once active that cluster is locked into functioning with this feature on.
```

I don't remember if this "must be activated at install time" is still true, though?  We assumed it was, but maybe it's not?  @imiller0 do you know more about this?

imiller0: My recollection is that workload partitioning must be set during initial install to ensure that pods are correctly mutated during admission. So long as it is enabled at initial install the cpuset can be modified at day-N. @eggfoobar has been working on this most recently and may know better if it is still required to be enabled at cluster install.

pixelsoccupied: Yep Egli actually [suggested](https://github.com/openshift-kni/cnf-features-deploy/pull/1376#discussion_r1065922360) this change

PR #1377, opened 2023-01-14
ran-profile CI and ZTP policygenerator intergration

Missxiaoguo:
This PR aims to replace the old DU RAN profile with the ZTP policy generator generated artifacts used in the ran-profile CI. 
- The ran-profile directory is now pointed to the true DU profile which is a set of PGTs located in the ztp directory. The ran-profile-std-cloud contains the overlay for the ran profile deployed on cloud environment. 
- The generate-ran-artifacts.sh script is added to generate the DU artifacts based on the true DU profile prior to feature deploy.
- ran-profile-ci entry is added in Makefile for the ran-profile CI job.
 
Jira: [CNF-2983](https://issues.redhat.com//browse/CNF-2983)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,12 @@ origin-tests-disconnected-environment:
 
 validate-on-ci: setup-test-cluster setup-build-index-image feature-deploy wait-and-validate
 
+ran-profile-ci: generate-ran-artifacts validate-on-ci
+
+generate-ran-artifacts:
+	@echo "Generating the RAN DU features"
+	hack/generate-ran-artifacts.sh
+
 gofmt:
 	@echo "Running gofmt"
 	hack/gofmt.sh
--- a/feature-configs/cn-ran-overlays/README.md
+++ b/feature-configs/cn-ran-overlays/README.md
@@ -31,73 +31,32 @@ Current example is focused on a DU. A CU profile can be built by
 - Using stock kernel instead of RT kernel
 - Skipping FEC (Work in progress)
 
-The [`ran-profile`](ran-profile) directory contains the Kustomize profile for deployment of DU integration features, namely:
-- SCTP MachineConfig patch
-- DU performance profile
-- PTP operator and slave profile
-- SR-IOV operator and associated profiles
-
-## The manifest structure
-
-The profile is built from one cluster specific folder and one or more site-specific folders. This is done to address a deployment that includes remote worker nodes (several sites belonging to the same cluster).
-The [`cluster-config`](ran-profile/cluster-config) directory contains performance and PTP customizations based upon operator deployments in [`deploy`](../feature-configs/deploy) folder.
-The [`site.1.fqdn`](site.1.fqdn) folder contains site-specific network customizations.
-
-
-## Prerequisites
-
-1. Create a machine config pool for the RAN worker nodes. For example:
-
-```
-cat <<EOF | oc apply -f -
-apiVersion: machineconfiguration.openshift.io/v1
-kind: MachineConfigPool
-metadata:
-  name: worker-cnf
-  labels:
-    machineconfiguration.openshift.io/role: worker-cnf
-spec:
-  machineConfigSelector:
-    matchExpressions:
-      - {
-          key: machineconfiguration.openshift.io/role,
-          operator: In,
-          values: [worker-cnf, worker],
-        }
-  paused: false
-  nodeSelector:
-    matchLabels:
-      node-role.kubernetes.io/worker-cnf: ""
----
-EOF
-```
-
-2. Include the worker node in the above machine config pool by labelling it with `node-role.kubernetes.io/worker-cnf` label:
-
-```bash
-oc label --overwrite node/{your node name} node-role.kubernetes.io/worker-cnf=""
-```
-3. Label the node as ptp slave (DU only):
-
-```bash
-oc label --overwrite node/{your node name} ptp/slave=""
-```
-
-An example of labelling the nodes used in CI/CD can be seen in `cnf-features-deploy/hack/setup-test-cluster.sh`
-
+The [`ran-profile`](ran-profile) directory contains the PolicyGenTemplates for deployment of DU integration features for three types of clusters, those PolicyGenTemplates are the symbolic links from the [`recommended DU profile`](../../ztp/gitops-subscriptions/argocd/example/policygentemplates/), that are:
+- A single `common-ranGen.yaml` that should apply to all types of sites
+- A set of shared `group-du-*-ranGen.yaml`, each of which should be common across a set of similar clusters
+- An example `example-*-site.yaml` which will normally be copied and updated for each individual site
 
 ## Deployment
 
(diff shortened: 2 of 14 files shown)


Missxiaoguo:
/hold

working in progress

feature-configs/cn-ran-overlays/ran-profile-gcp/pre_deploy.sh
```diff
@@ -0,0 +1,63 @@
...
+fi
+
+cd ${obsolute_feature_dir}
+cat > kustomization.yaml << EOF
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+EOF
+
+find cluster-config/ -type f | while read config_yaml; do
```
lack, line 56: Maybe we should add `-name='*.yaml'` to the find command to filter out potential other files like READMEs or editor swapfiles?

lack:
Last failures were due to cloud resource constraints.
/retest

feature-configs/cn-ran-overlays/ran-profile-gcp/pre_deploy.sh
```diff
@@ -0,0 +1,63 @@
...
+export YQ_VERSION=v4.18.1
+
+set -e
+
+echo "****** ran-profile-gcp pre_deploy.sh ******"
+echo "FEATURES_ENVIRONMENT: ${FEATURES_ENVIRONMENT}"
+
+echo "PRINT current location: $(pwd)"
+echo "PRINT script: $0"
+obsolute_feature_dir=$(dirname $(realpath "$0"))
```
lack, line 14: Minor typo `s/obsolute/absolute/`

cnf-tests/testsuites/validationsuite/cluster/validation.go
```diff
@@ -502,7 +502,7 @@ var _ = Describe("validation", func() {
 		})
 	})
 
-	Context("[multineworkpolicy]", func() {
+	Context("[multinetworkpolicy]", func() {
```
lack, line 505: Nice catch!

imiller0:
Compared to the existing CI infrastructure how long does the new CI take to run?

Missxiaoguo:
> Compared to the existing CI infrastructure how long does the new CI take to run?

The new CI takes 20mins longer than the old one mainly because the new profile applies sctp MCs for both master and worker nodes, also performanceprofile to update the kernel.

hack/feature-deploy.sh
```diff
@@ -53,6 +53,16 @@ do
...
+    pre_deploy=feature-configs/${FEATURES_ENVIRONMENT}/${feature}/pre_deploy.sh
+    if [[ -f $pre_deploy ]]; then
+        echo "[INFO] Pre-deploying feature '$feature' for environment '$FEATURES_ENVIRONMENT'"
+        if ! $pre_deploy; then
+          echo "[WARN] Pre-deploy of feature '$feature' failed."
+          feature_failed=1
+          continue
+      fi
+    fi
+
```
imiller0, line 65: Does this change affect only the CI runs on changes under cnf-features-deploy/ztp, or are the CI runs for cnf-tests also affected by this change?

feature-configs/cn-ran-overlays/hacks/pre_deploy.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+set -e
+
+absolute_feature_dir="$(pwd)/$(dirname $0)"
+feature=$(basename ${absolute_feature_dir})
+cluster_type="${CLUSTER_TYPE:-standard}"
```
imiller0, line 7: For future code maintenance it would be helpful to list at the top the set of environment variables that are used as inputs to this script.

feature-configs/cn-ran-overlays/hacks/pre_deploy.sh
```diff
@@ -0,0 +1,82 @@
...
+
+    # call ztp generator cli to generate the configuration CRs for CI environment
+    if [ $feature != $du_source_profile ]; then
+        ci_config_CRs="${ztp_generator_cli} config -N ${absolute_feature_dir}/policygentemplates ${absolute_feature_dir}/cluster-config"
+        if ! $ci_config_CRs; then
+            echo "[ERROR] Failed to generate the configuration CRs based on the overrided PGTs for CI environment: ${absolute_feature_dir}/policygentemplates"
+            exit 1
+        fi
+    fi
+fi
```
imiller0, line 55: If I understand correctly this block is generating configuration CRs that are specific to the CI environment and the block above (lines 39-45) are generating configuration CRs from the reference. It is a bit confusing that both contain "common-rangen.yaml" and "group-du-standard-ranGen.yaml". Is the intent that this later set is overwriting the reference versions? Or are they overwriting just parts of the reference versions? A note to clarify how these two blocks relate would be helpful.

imiller0 (commented):
Are the two commits in this PR independent, or can they be collapsed into a single one? The goal is to keep the history clean with "atomic" changes.

hack/feature-deploy.sh
Missxiaoguo: It only affects the CI on changes under cnf-features-deploy/ztp right now, but the change is generic which can be leveraged by other CIs for cnf-tests who want to run something prior to feature deploy.

feature-configs/cn-ran-overlays/hacks/pre_deploy.sh
Missxiaoguo: The PGTs in the ran-profile-gcp repo override the parts of the reference versions that are for CI env.  Will update.

imiller0:
/lgtm
Waiting on review by someone from cnf-tests before approving.

yuvalk:
commits need to be atomic, where each commit is doing a single change.
for the very least, please separate the feature-deploy change into it's own commit.

yuvalk:
I wonder how did this even pass the ci job, I guess we have a bug there..

yuvalk:
> commits need to be atomic, where each commit is doing a single change. for the very least, please separate the feature-deploy change into it's own commit.
I meant for the pre_deploy addition

please also make sure to explain the WHY of a change in the commit msg
so history is relevant and people can use it to understand why certain changes happend

ie, in this case, why do you need the pre_deploy is needed

Missxiaoguo:
> > commits need to be atomic, where each commit is doing a single change. for the very least, please separate the feature-deploy change into it's own commit.
> > I meant for the pre_deploy addition
> 
> please also make sure to explain the WHY of a change in the commit msg so history is relevant and people can use it to understand why certain changes happend
> 
> ie, in this case, why do you need the pre_deploy is needed

Updated.  Yeah, I initially had different commits for different components but I found out those can be in one commit without CI compliant. Looks like the CI can pass as long as you have a valid component added in the commit title and no matter changes are checked in which repo.

Missxiaoguo:
Closed since it's obselete

PR #1381, opened 2023-01-20
ztp: OCPBUGS-4449: Disable ice module when doing a kdump

bartwensley:
Adding a MachineConfig to remove the ice module from kdump to prevent kdump failures on certain servers. This is a temporary workaround for RHELPLAN-138236 and can be removed when that issue is fixed.

Note: This is done by editing the kdump config file. When this MachineConfig is removed, the changes to the kdump config file will not be undone. This is OK because the ice module is not used when doing a kdump.

Signed-off-by: Bart Wensley <bwensley@redhat.com>


ztp/extra-manifests-builder/05-kdump-config/kdump-remove-ice-module.sh
```diff
@@ -0,0 +1,25 @@
...
+
+SED="/usr/bin/sed"
+GREP="/usr/bin/grep"
+
+# override for testing purposes
+KDUMP_CONF="${1:-/etc/sysconfig/kdump}"
+REMOVE_ICE_STR="module_blacklist=ice"
+
+# exit if file doesn't exist
+[ ! -f ${KDUMP_CONF} ] && exit 0
```
lack, line 17: Normally we want to `exit 1` on error? Not sure how important that is in this case, though?

ztp/extra-manifests-builder/05-kdump-config/kdump-remove-ice-module.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# exit if file already updated
+${GREP} -Fq ${REMOVE_ICE_STR} ${KDUMP_CONF} && exit 0
+
+# Target line looks something like this:
+# KDUMP_COMMANDLINE_APPEND="irqpoll nr_cpus=1 ... hest_disable"
+# Use sed to match everything between the quotes and append the REMOVE_ICE_STR to it
+${SED} -i 's/^KDUMP_COMMANDLINE_APPEND="[^"]*/& '${REMOVE_ICE_STR}'/' ${KDUMP_CONF} || exit 0
```
lack, line 25: Is there a reason we are squashing sed's exit code here? Wouldn't we want an editing error to be reported/recorded?

ztp/extra-manifests-builder/05-kdump-config/kdump-remove-ice-module.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# override for testing purposes
+KDUMP_CONF="${1:-/etc/sysconfig/kdump}"
+REMOVE_ICE_STR="module_blacklist=ice"
+
+# exit if file doesn't exist
+[ ! -f ${KDUMP_CONF} ] && exit 0
+
+# exit if file already updated
+${GREP} -Fq ${REMOVE_ICE_STR} ${KDUMP_CONF} && exit 0
```
lack, line 20: This exit is perfect. No-op = success.

ztp/extra-manifests-builder/05-kdump-config/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+kdump_remove_ice_module_base64=$(base64 -w0 kdump-remove-ice-module.sh)
+
+echo "\
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: ${MCPROLE}
+  name: 05-kdump-config-${MCPROLE}
```
lack, line 17: Would we want a more specific name like `05-kdump-config-icelake-fix`?

Probably not that important, and I'm fine with it as-is.

lack (approved):
Looks great!

All my comments are suggestions only, and not really needed for this change.

/ok-to-test

lack:
@bartwensley Looks like we must still have a unit test that records yet another copy of these Machine configs that also needs updating :(

ztp/extra-manifests-builder/05-kdump-config/build.sh
bartwensley: I was thinking that it is possible there would be other issues in this area in the future where we might have to blacklist another driver (for example), so I wanted to use a more generic name for the machineconfig to allow for that.

ztp/extra-manifests-builder/05-kdump-config/kdump-remove-ice-module.sh
bartwensley: My thinking (could be wrong) was that this is a non-essential change and that I didn't want to cause the machineconfig to fail any case. I didn't want to affect the health of the system in any way, regardless of what happens in this machineconfig.

lack (approved):

PR #1383, opened 2023-01-24
ztp: configure composable Openshift with installConfigOverrides

daosman:
installConfigOverrides is the generic way of passing install-config parameters through the siteConfig.
This change adds a sample installConfigOverrides that configures the composable openshift feature with the 'capabilities' field.  It removes all but the marketplace component from the optional set of components.  The resulting 'capabilities' in the install-config will be:

capabilities:
  baselineCapabilitySet: None
    additionalEnabledCapabilities:
      - marketplace

### DIFF
--- a/ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
+++ b/ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
@@ -16,6 +16,7 @@ spec:
     biosConfigRef:
       filePath: "testSiteConfig/testHW.profile"
     networkType: OVNKubernetes
+    installConfigOverrides:  "{\"capabilities\":{\"baselineCapabilitySet\": \"None\", \"additionalEnabledCapabilities\": [ \"marketplace\" ] }}"
     extraManifestPath: testSiteConfig/testUserExtraManifest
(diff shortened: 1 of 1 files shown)


ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
```diff
@@ -16,6 +16,7 @@ spec:
     biosConfigRef:
       filePath: "testSiteConfig/testHW.profile"
     networkType: OVNKubernetes
+    installConfigOverrides:  "{\"capabilities\":{\"baselineCapabilitySet\": \"None\", \"additionalEnabledCapabilities\": [ \"marketplace\" ] }}"
```
imiller0, line 19: This file is purely for internal unit testing and is not user visible, documented or delivered to users. Is the intent to provide an example/sample for users? If so it should be put in https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/example/siteconfig/example-standard.yaml (or one of the other siteconfigs there). However, those examples are our baseline reference configuration and should reflect the mainline use case. One option would be to add this but leave the line commented out to show that it is an optional field. 
This should also be added to documentation.

PR #1386, opened 2023-01-25
ztp: example config of composable Openshift

daosman:
installConfigOverrides is the generic way of passing install-config parameters through the siteConfig.
This change adds a sample installConfigOverrides that configures the composable openshift feature with the 'capabilities' field.  It removes all but the marketplace component from the optional set of components.  The resulting 'capabilities' in the install-config will be:

capabilities:
  baselineCapabilitySet: None
    additionalEnabledCapabilities:
      - marketplace

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml
@@ -14,6 +14,12 @@ spec:
   clusters:
   - clusterName: "example-sno"
     networkType: "OVNKubernetes"
+    # installConfigOverrides is a generic way of passing install-config
+    # parameters through the siteConfig.  The 'capabilities' field configures
+    # the composable openshift feature.  In this 'capabilities' setting, we
+    # remove all but the marketplace component from the optional set of
+    # components.
+    installConfigOverrides:  "{\"capabilities\":{\"baselineCapabilitySet\": \"None\", \"additionalEnabledCapabilities\": [ \"marketplace\" ] }}"
     clusterLabels:
       # These example cluster labels correspond to the bindingRules in the PolicyGenTemplate examples in ../policygentemplates:
       # ../policygentemplates/common-ranGen.yaml will apply to all clusters with 'common: true'


ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml
```diff
@@ -14,6 +14,11 @@ spec:
   clusters:
   - clusterName: "example-sno"
     networkType: "OVNKubernetes"
+    # Optionally; installConfigOverrides is a generic way of passing install-config
+    # parameters through the siteConfig.  The capabilities field configures the
+    # composable openshift feature .  In this example, we remove all but the
+    # marketplace component from the optional set of components.
+    #installConfigOverrides:  "{\"capabilities\":{\"baselineCapabilitySet\": \"None\", \"additionalEnabledCapabilities\": [ \"marketplace\" ] }}"
```
imiller0, line 21: For the primary use case is this optional, or do we expect new clusters to include this setting? If this is intended as part of the reference configuration we should uncomment it here. @browsell what do you think?

browsell: This needs to be part of the reference config, not optional

daosman: Removed optional tag and comment

ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml
```diff
@@ -14,6 +14,11 @@ spec:
   clusters:
   - clusterName: "example-sno"
     networkType: "OVNKubernetes"
+    # installConfigOverrides is a generic way of passing install-config
+    # parameters through the siteConfig.  The capabilities field configures the
+    # composable openshift feature.  In this example, we remove all but the
+    # marketplace component from the optional set of components.
```
imiller0, line 20: Minor nit, this is no longer an example. Maybe something like "This default value removes all but the ...."

imiller0 (approved):
LGTM

PR #1388, opened 2023-01-27
Bug 2111108: ztp: Remove container-mount-namespace.service installation warnings

lack:
The unit doesn't need to be enabled as it's set up as a dependency of
crio and kubelet.  Removing the 'enabled' flag from the MachineConfig
will remove an ignition warning message that could be concerning to
users.

/hold until testing is completed


lack:
Testing complete, looks good!  Spun up a clusterbot with the latest 4.12 and applied this config.  End result is a cluster with mount namespace separation enabled:
```
sh-4.4# readlink /proc/$(pgrep kubelet)/ns/mnt
mnt:[4026532171]
sh-4.4# readlink /proc/$(pgrep crio)/ns/mnt
mnt:[4026532171]
sh-4.4# readlink /proc/1/ns/mnt 
mnt:[4026531840]
```

/unhold

PR #1391, opened 2023-02-07
ztp: Add workloadHints to source-crs/PerformanceProfile

sakhoury:
This commit updates source-crs/PerformanceProfile with workloadHints for powerstate configurations. The default configuration is set to performance mode.

Tests have been performed to validate that no reboot on SNO upgrade will be triggered as a result of this change.

Signed-off-by: Sharat Akhoury <sakhoury@redhat.com>

/assign @bartwensley @browsell @imiller0

### DIFF
--- a/ztp/source-crs/PerformanceProfile.yaml
+++ b/ztp/source-crs/PerformanceProfile.yaml
@@ -30,3 +30,11 @@ spec:
   # To use the standard (non-realtime) kernel, set enabled to false
   realTimeKernel:
     enabled: true
+  workloadHints:
+    # WorkloadHints defines the set of upper level flags for different type of workloads.
+    # See https://github.com/openshift/cluster-node-tuning-operator/blob/master/docs/performanceprofile/performance_profile.md#workloadhints
+    # for detailed descriptions of each item.
+    # The configuration below is set for a low latency, performance mode.
+    realTime: true
+    highPowerConsumption: false
+    perPodPowerManagement: false


ztp/source-crs/PerformanceProfile.yaml
```diff
@@ -30,3 +30,7 @@ spec:
   # To use the standard (non-realtime) kernel, set enabled to false
   realTimeKernel:
     enabled: true
+  workloadHints:
+      realTime: true
+      highPowerConsumption: false
+      perPodPowerManagement: false
```
bartwensley, line 36: Looks like the rest of the file is using 2 spaces for the indent - should probably be consistent.

bartwensley (approved):
Looks good to me

ztp/source-crs/PerformanceProfile.yaml
```diff
@@ -30,3 +30,7 @@ spec:
   # To use the standard (non-realtime) kernel, set enabled to false
   realTimeKernel:
     enabled: true
+  workloadHints:
```
imiller0, line 33: Can you add a comment here to describe what this section is for. Does not need to be comprehensive documentation but a minimal indication of what workloadHints are for and what behavior these default values result in.

browsell: We should add a link to the official documentation vs re-documenting.

ztp/source-crs/PerformanceProfile.yaml
```diff
@@ -30,3 +30,14 @@ spec:
...
+    # WorkloadHints defines the set of upper level flags for different type of workloads.
+    # See https://github.com/openshift/cluster-node-tuning-operator/blob/master/docs/performanceprofile/performance_profile.md#workloadhints
+    # for detailed descriptions of each item.
+    # The configuration below is set for a low latency, performance mode (i.e. default DU profile).
+    # Set `highPowerConsumption: true` for ultra-low latency, high performance mode.
+    # Set `perPodPowerManagement: true` for power saving mode.
+    # Note that perPodPowerManagement and highPowerConsumption hints can not be enabled together.
```
imiller0, line 40:
```suggestion
    # workloadHints defines the set of upper level flags for different type of workloads.
    # See https://github.com/openshift/cluster-node-tuning-operator/blob/master/docs/performanceprofile/performance_profile.md#workloadhints
    # for detailed descriptions of each item.
    # The configuration below is set for a low latency, high performance mode.
```

imiller0 (approved):
/lgtm

PR #1402, opened 2023-02-14
ztp: remove systemd from accelerated-container-startup

daosman:
OCPBUGS-4146


liornoy:
@daosman I see that the CI failed on the ztp component. 
I want to raise the fact that we bumped the cnf-tests to go v1.19 
and the openshift_ci_image also to v1.19 as part of this pr: [link](https://github.com/openshift-kni/cnf-features-deploy/commit/7eac88479fb666ebc6d4e89dcdbcbd72590c0dab)
Where it looks like it accidentally skipped the ztp [Containerfile](https://github.com/openshift-kni/cnf-features-deploy/blob/8459bd30ad97f1c7b0133ef6a92162faaaa469ef/ztp/resource-generator/Containerfile#L2)
What do you think? is it the cause of this error?

daosman:
@liornoy, there was a previous failure for gofmt error that was fixed in the [PR](https://github.com/openshift-kni/cnf-features-deploy/pull/1403).
Not sure what the current failure is

lack (approved):

PR #1406, opened 2023-02-20
infra: Hide podman login in logs to prevent password leaks

sshnaidm:
Set +x before login so it's not leaked in logs, set -x back after a login.

### DIFF
--- a/hack/setup-build-index-image.sh
+++ b/hack/setup-build-index-image.sh
@@ -54,8 +54,10 @@ spec:
           wget https://github.com/operator-framework/operator-registry/releases/download/v1.23.0/linux-amd64-opm
           mv linux-amd64-opm opm
           chmod +x ./opm
-          export pass=$( jq .\"image-registry.openshift-image-registry.svc:5000\".password /var/run/secrets/openshift.io/push/.dockercfg )
+          set +x
+          pass=$( jq .\"image-registry.openshift-image-registry.svc:5000\".password /var/run/secrets/openshift.io/push/.dockercfg )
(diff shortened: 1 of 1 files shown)


ffromani (commented):
/lgtm

hack/setup-build-index-image.sh
```diff
@@ -54,8 +54,10 @@ spec:
           wget https://github.com/operator-framework/operator-registry/releases/download/v1.23.0/linux-amd64-opm
           mv linux-amd64-opm opm
           chmod +x ./opm
+          set +x
           export pass=$( jq .\"image-registry.openshift-image-registry.svc:5000\".password /var/run/secrets/openshift.io/push/.dockercfg )
```
lack, line 58: Should this also not be 'export'?  If the only use is the next line, a non-exported variable would be safer, otherwise every command after this point may have access to it and accidentally log it if they log their environment.

sshnaidm: Yeah, good catch. Let me do it and test.

sshnaidm:
Tested in  https://github.com/openshift/release/pull/36581 and no password leaks: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/36581/rehearse-36581-periodic-ci-openshift-release-master-nightly-4.12-e2e-telco5g-cnftests/1627762726417731584/artifacts/e2e-telco5g-cnftests/telco5g-cnf-tests/build-log.txt

PR #1408, opened 2023-02-21
ztp: Add ts2phc configuration to new ptpconfig 

aneeshkp:
ptp config to support ts2phc implementation

### DIFF
--- a/ztp/source-crs/PtpConfigGmWpc.yaml
+++ b/ztp/source-crs/PtpConfigGmWpc.yaml
@@ -1,0 +1,149 @@
+# The grandmaster profile is provided for testing only
+# It is not installed on production clusters
+apiVersion: ptp.openshift.io/v1
+kind: PtpConfig
+metadata:
+  name: grandmaster
+  namespace: openshift-ptp
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "10"
+spec:
+  profile:
+  - name: "grandmaster"
+    ptp4lOpts: "-2 --summary_interval -4"
+    phc2sysOpts: -r -u 0 -m -O -37 -N 8 -R 16 -s $iface_master -n 24
+    ptpSchedulingPolicy: SCHED_FIFO
+    ptpSchedulingPriority: 10
+    plugins:
+      e810:
+        enableDefaultConfig: true
+    ts2phcOpts: " "
+    ts2phcConf: |
+      [nmea]
+      ts2phc.master 1
+      [global]
+      use_syslog  0
+      verbose 1
+      logging_level 7
+      ts2phc.pulsewidth 100000000
+      #GNSS module s /dev/ttyGNSS* -al use _0
+      #cat /dev/ttyGNSS_1700_0 to find available serial port
+      #example value of gnss_serialport is /dev/ttyGNSS_1700_0
+      ts2phc.nmea_serialport $gnss_serialport
+      leapfile  /usr/share/zoneinfo/leap-seconds.list
+      [$iface_master]
+      ts2phc.extts_polarity rising
+      ts2phc.extts_correction 0
+    ptp4lConf: |
+      [$iface_master]
+      masterOnly 1
+      [$iface_master_1]
+      masterOnly 1
+      [$iface_master_2]
+      masterOnly 1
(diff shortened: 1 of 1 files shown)


yliu127:
lgtm

josephdrichard:
Also needs the operatorconfig change to enable this.
```
apiVersion: ptp.openshift.io/v1
kind: PtpOperatorConfig
metadata:
  name: default
  namespace: openshift-ptp
  annotations:
spec:
  daemonNodeSelector:
    node-role.kubernetes.io/$mcp: ""
  plugins:
  - e810
```

yliu127:
@josephdrichard is the new  PtpOperatorConfig plugins spec  compatible with previous ptp operator versions? If not, we need to discuss with @imiller0  on the best approach. May add a new source-cr or add a commented spec as an example but let user to actually add it in their PGT.

josephdrichard:
We should probably have a new PtpOperatorConfig for this, similar to what this commit is doing to create a new PtpConfig

ztp/source-crs/PtpConfigGmWpc.yaml
```diff
@@ -0,0 +1,150 @@
...
+      transportSpecific 0x0
+      ptp_dst_mac 01:1B:19:00:00:00
+      p2p_dst_mac 01:80:C2:00:00:0E
+      udp_ttl 1
+      udp6_scope 0x0E
+      uds_address /var/run/ptp4l
+      #
+      # Default interface options
+      #
+      clock_type OC
```
josephdrichard, line 127: Should this be BC?  I don't see any option for specifying GM here.

ztp/source-crs/PtpConfigGmWpc.yaml
```diff
@@ -0,0 +1,150 @@
...
+      # Default Data Set
+      #
+      twoStepFlag 1
+      priority1 128
+      priority2 128
+      domainNumber 24
+      #utc_offset 37
+      clock_class_threshold 165
+      clockClass 248
+      clockAccuracy 0xFE
```
josephdrichard, line 57: For E810, I believe this is within 50 μs, which would gfall in this category
0x27:  The time is accurate to within 100 μs

ztp/source-crs/PtpConfigGmWpc.yaml
```diff
@@ -0,0 +1,150 @@
...
+      [global]
+      #
+      # Default Data Set
+      #
+      twoStepFlag 1
+      priority1 128
+      priority2 128
+      domainNumber 24
+      #utc_offset 37
+      clock_class_threshold 165
```
josephdrichard, line 55: Drop.  Not applicable for GM.

aneeshkp: DOne

ztp/source-crs/PtpConfigGmWpc.yaml
```diff
@@ -0,0 +1,150 @@
...
+      clockClass 248
```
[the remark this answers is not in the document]
aneeshkp: Done

ztp/source-crs/PtpConfigGmWpc.yaml
aneeshkp: done

ztp/source-crs/PtpConfigGmWpc.yaml
aneeshkp: done

ztp/source-crs/PtpConfigGmWpc.yaml
```diff
@@ -0,0 +1,149 @@
...
+      ingressLatency 0
+      boundary_clock_jbod 0
+      #
+      # Clock description
+      #
+      productDescription ;;
+      revisionData ;;
+      manufacturerIdentity 00:00:00
+      userDescription ;
+      timeSource 0xA0
```
josephdrichard, line 143: timeSource should be 0x20 for GNSS

aneeshkp: Fixed

josephdrichard (approved):
/lgtm

PR #1414, opened 2023-02-28
ztp: Only provide composable openshift examples for SNO clusters

lack:


imiller0 (approved):
/LGTM

PR #1415, opened 2023-02-28
[release-4.12] OCPBUGS-5380: ztp: remove systemd from accelerated-container-startup

daosman:
OCPBUGS-5380


imiller0 (approved):
lgtm

PR #1419, opened 2023-03-04
OCPBUGS-8313: Add support for apiVIPs and ingressVIPs

loganmc10:
As of OCP 4.12, the `apiVIP` and `ingressVIP` install parameters are deprecated[1]

This change adds support for the new array-based `apiVIPs` and `ingressVIPs`. The examples were updated to use the new syntax, though the old syntax is still supported.

[1]https://docs.openshift.com/container-platform/4.12/release_notes/ocp-4-12-release-notes.html#ocp-4-12-nw-api-ingress-ipv6-support

### DIFF
--- a/ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site2-sno-du.yaml
+++ b/ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site2-sno-du.yaml
@@ -19,7 +19,7 @@ spec:
       common: true
       sites : "site-sno-du-2"
     apiVIP: "10.16.231.101"
-    ingressVIP: "10.16.231.102"       
+    ingressVIP: "10.16.231.102"
     clusterNetwork:
       - cidr: 10.128.0.0/14
         hostPrefix: 23
--- a/ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
@@ -162,6 +162,8 @@ spec:
   sshPublicKey: "ssh-rsa "
   clusters:
   - clusterName: "cluster1"
+    apiVIP: 10.16.231.2
+    ingressVIP: 10.16.231.3
     clusterNetwork:
       - cidr: 10.128.0.0/14
         hostPrefix: 23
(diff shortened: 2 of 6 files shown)


serngawy:
the commit message need to start with **ztp** to pass the ci job for example "ztp: Add support for apiVIPs and ingressVIPs"

loganmc10:
Thanks @serngawy. Is there something I should do about the other 2 jobs? It's unclear to me why they are failing

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site2-sno-du.yaml
```diff
@@ -18,8 +18,10 @@ spec:
       group-du-sno: ""
       common: true
       sites : "site-sno-du-2"
-    apiVIP: "10.16.231.101"
```
serngawy, line 21: please don't change old unit test data, we have to keep back compatibility

ztp/siteconfig-generator/siteConfig/testdata/siteConfigStandardClusterTestOutput.yaml
```diff
@@ -19,10 +19,12 @@ metadata:
     name: cluster1
     namespace: cluster1
 spec:
+    apiVIP: 10.16.231.2
```
serngawy, line 22: same don't change old unit test data

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -162,6 +162,8 @@ spec:
   sshPublicKey: "ssh-rsa "
   clusters:
   - clusterName: "cluster1"
+    apiVIP: 10.16.231.2
```
serngawy, line 165: same don't change old unit test data

serngawy (changes requested):
would you add validation in this [function](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go#L69) That user cannot set apiVIP and apiVIPs at the same time. similar ingressVIP and ingressVIPs

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
loganmc10: I added this to ensure that both the `apiVIP` and `apiVIPs` syntax are tested. Is this unit test even valid without `apiVIP`? It's a unit test for a standard (multinode) cluster, I believe that apiVIP would be a required field.

Without this, there isn't a test that includes `apiVIP` (that I saw)

serngawy: okay, that is strange it should complain if there is no api or ingress IPs set

loganmc10:
Actually, I need to re-work this a bit, I just tested, and strangely, both `apiVIP` and `apiVIPs` need to be set. The example also shows it this way: https://github.com/openshift/assisted-service/blob/5dcb76d0b06479fa8852ccd2082227eceea51826/docs/hive-integration/crds/agentClusterInstall-dualstack-vips.yaml#L7-L10

If you just set `apiVIPs`, you get a `api-vips-defined` validation failure from the AgentClusterInstall.

I'm not sure why it was designed this way, but I'll update the PR to match the way that AgentClusterInstall expects it.

loganmc10:
Ok @serngawy I think this is good now. Actually, when you specify `apiVIPs` and `ingressVIPs`, you still need to specify `apiVIP` and `ingressVIP` (and they need to match `apiVIPs[0]` and `ingressVIPs[0]`). I'm not really sure why it's that way, but I've updated the PR to match that expectation

serngawy:
> Ok @serngawy I think this is good now. Actually, when you specify `apiVIPs` and `ingressVIPs`, you still need to specify `apiVIP` and `ingressVIP` (and they need to match `apiVIPs[0]` and `ingressVIPs[0]`). I'm not really sure why it's that way, but I've updated the PR to match that expectation

yes, it made this way to handle back compatibility check [here](https://github.com/openshift/assisted-service/blob/master/internal/cluster/validations/validations.go#L363)  , sounds good will merge it

serngawy (approved):

loganmc10:
Thanks @serngawy , when I opened this PR, the release-4.13 branch didn't exist. It looks like it exists now, so I assume this will also need to be cherry-picked into that branch?

PR #1425, opened 2023-03-07
OCPBUGS-7540: ztp: Add `soakSeconds` annotation to informDUValidator

sudomakeinstall2:
first time TALM observes a policy with this annotation as compliant, it won't move on. In future reconcilations if the specified number of seconds has passed since the first compliancy time and policy is still compliant, then TALM will move on from this policy.

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-validator-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-validator-ranGen.yaml
@@ -20,3 +20,8 @@ spec:
     - fileName: validatorCRs/informDuValidator.yaml
       remediationAction: inform
       policyName: "du-policy"
+      
+      # This low setting is only valid if the validation policy is disconnected from the cluster at steady-state
+      # using a bindingExcludeRules entry with ztp-done
+      evaluationInterval:
+        compliant: 5s
--- a/ztp/policygenerator-kustomize-plugin/kustomization.yaml
+++ b/ztp/policygenerator-kustomize-plugin/kustomization.yaml
@@ -4,3 +4,25 @@ generators:
 - testPolicyGenTemplate/group-du-sno-ranGen.yaml
 - testPolicyGenTemplate/site-du-sno-1-ranGen.yaml
 - testPolicyGenTemplate/site-du-sno-2-ranGen.yaml
+- testPolicyGenTemplate/example-group-du-sno-validator-ranGen.yaml
+
+# Updates to the PerformanceProfile or MachineConfigurations may take a short period of time before 
+# causing the MachineConfigPool to move to degraded. This patch adds an annotation to validator policy.
+# This soak time ensures that the upgrading status can be set prior to considering the validation policy
+# in compliance
+patches:
+- patch: |-
+    apiVersion: policy.open-cluster-management.io/v1
+    kind: Policy
+    metadata:
+      annotations:
+        # first time TALM observes this policy as compliant, it won't move on.
+        # In future reconcilations if the specified number of seconds has passed 
+        # since the first compliancy time and policy is still compliant, then TALM
+        # will move on from this policy
+        ran.openshift.io/soak-seconds: "30"
+
+      # The 'name' an 'namespace' values must match the values in the validator policy
+      # the policy name format is as {name}-{spec.sourceFiles[x].policyName}
+      name: example-group-du-sno-validator-du-policy
+      namespace: ztp-group
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-validator-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-validator-ranGen.yaml
@@ -20,3 +20,8 @@ spec:
     - fileName: validatorCRs/informDuValidator.yaml
       remediationAction: inform
       policyName: "du-policy"
+
+      # This low setting is only valid if the validation policy is disconnected from the cluster at steady-state
+      # using a bindingExcludeRules entry with ztp-done
+      evaluationInterval:
+        compliant: 5s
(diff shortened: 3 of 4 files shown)


ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-validator-ranGen.yaml
```diff
@@ -20,3 +20,4 @@ spec:
     - fileName: validatorCRs/informDuValidator.yaml
       remediationAction: inform
       policyName: "du-policy"
+      compliant: 5s
```
imiller0, line 23: This low setting is only valid if the validation policy is disconnected from the cluster at steady-state using a bindingExcludeRules entry with ztp-done. Please add a comment to help document this.

ztp/source-crs/validatorCRs/informDuValidator.yaml
```diff
@@ -4,6 +4,11 @@ metadata:
   name: $mcp
   annotations:
     ran.openshift.io/ztp-deploy-wave: "10000"
+    # first time TALM observes this policy as compliant, it won't move on.
+    # In future reconcilations if the specified number of seconds has passed 
+    # since the first compliancy time and policy is still compliant, then TALM
+    # will move on from this policy
+    soakSeconds: "30"
```
imiller0, line 11: Can we make this annotation more TALM specific? If, for example, the Policy team implements a similar annotation we don't want to have a name collision. eg ran.openshift.io/soakSeconds

imiller0 (commented):
Is the 3node cluster validator missing?

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-validator-ranGen.yaml
jc-rh, line 23: This field should be under `evaluationInterval`

ztp/source-crs/validatorCRs/informDuValidator.yaml
jc-rh: Is this the right place for the annotation though? @serngawy Policy gen copies it from this MCP CR template to the policy?

serngawy: this annotation `soakSecond` will be in the MachineConfigPool that exist in the policy not in the policy annotation. Is that what you are looking to do ?

jc-rh: In the end, it needs to be in policy annotation for TALM to recognize. However, from a modelling perspective, it does make some sense to attach it to the actual CR template at the source. How about policy gen checking this value in CRs and use the greatest at the policy level? Sort of similar to the checks on the waves.

ztp/policygenerator-kustomize-plugin/kustomization.yaml
```diff
@@ -4,3 +4,18 @@ generators:
...
+    kind: Policy
+    metadata:
+      annotations:
+        # first time TALM observes this policy as compliant, it won't move on.
+        # In future reconcilations if the specified number of seconds has passed 
+        # since the first compliancy time and policy is still compliant, then TALM
+        # will move on from this policy
+        ran.openshift.io/soak-seconds: "30"
+      name: group-du-sno-validator-du-policy
+      namespace: ztp-group
```
sudomakeinstall2, line 21: does this look good?
@jc-rh @imiller0 @serngawy

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-validator-ranGen.yaml
```diff
@@ -20,3 +20,5 @@ spec:
     - fileName: validatorCRs/informDuValidator.yaml
       remediationAction: inform
       policyName: "du-policy"
+      evaluationInterval:
+        compliant: 5s
```
jc-rh, line 24: As Ian mentioned, can we add some comments here to explain why this short interval value is ok with this policy.

ztp/policygenerator-kustomize-plugin/kustomization.yaml
imiller0: I think that the comment should discuss why adding this is necessary. Something along the lines of "updates to the PerformanceProfile or MachineConfigurations may take a short period of time before causing the MachineConfigPool to move to degraded. This soak time ensures that the upgrading status can be set prior to considering the validation policy in compliance.

ztp/policygenerator-kustomize-plugin/testPolicyGenTemplate/group-du-sno-validator-ranGen.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: ran.openshift.io/v1
+kind: PolicyGenTemplate
+metadata:
+  # The name will be used to generate the placementBinding name as {name}-placementBinding, the placementRule name as {name}-placementRule,
+  # and the policy name as {name}-{spec.sourceFiles[x].policyName}
+  name: "group-du-sno-validator"
+  namespace: "ztp-group"
```
imiller0, line 7: Can you make this look more like test code? When the test versions are nearly identical to the reference configuration it causes confusion on what the active version of code is.

ztp/policygenerator-kustomize-plugin/kustomization.yaml
imiller0, line 21: We need a comment that the name/namespace here **must match** the values in the validator policy.

imiller0 (approved):
/lgtm

As a short-term fix/workaround this is good, however we need to continue working on the underlying issue. Customers will need to adapt the patch in kustomization.yaml to their deployment/environment which requires that we communicate it to them and that they correctly implement it (any error will simply miss adding the annotation which may not be noticed until upgrading at scale).

PR #1426, opened 2023-03-07
[release-4.10] OCPBUGS-8527: ztp: Remove ovs from accelerated-container-startup

daosman:
OCPBUGS-4146

ovs remove the ability to dynamically set the number of handler threads since 2.16.  This removed our abitlity to increase the number of handler threads early at boot and reduce it when Openshift reaches steady state to speed up the boot process.

This fix is removing the affinity manipulation for the ovs process from the accelerated-container-startup.sh script.

Tests showed that we do not lose much from removing ovs from the set of processes we manipulate the affinity in the
accelerated-container-startup.sh script.


daosman:
/retitle [release-4.10] OCPBUGS-8527: ztp: Remove ovs from accelerated-container-startup

imiller0 (approved):
/lgtm

PR #1427, opened 2023-03-08
ztp: remove systemd from accelerated-container-startup

daosman:
OCPBUGS-8527


daosman:
/retitle [release-4.10] OCPBUGS-8527: ztp: Remove ovs from accelerated-container-startup

imiller0 (approved):
/lgtm

PR #1431, opened 2023-03-09
OCPBUGS-8238: ztp: Remove incorrect macAddress field

imiller0:
The macAddress field under
spec.clusters[].nodes[].nodeNetwork.config.interfaces[] is not needed. The network configuration is matched to interfaces based on the macAddress under the spec.clusters[].nodes[].nodeNetwork.interfaces[]. In addition, if the config.interfaces field is included it needs to be mac-addresses. Since it is not necessary this PR simply removes it.

### DIFF
--- a/ztp/gitops-subscriptions/argocd/SNOExpansion.md
+++ b/ztp/gitops-subscriptions/argocd/SNOExpansion.md
@@ -274,7 +274,6 @@ EOF
               - name: eno1
                 type: ethernet
                 state: up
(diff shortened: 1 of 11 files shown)


tliu2021:
should we also remove spec.clusters[].nodes[].nodeNetwork.config.interfaces[].macAddress from cnf-features-deploy/ztp/siteconfig-generator-kustomize-plugin/testSiteConfig?

nishant-parekh (approved):

PR #1433, opened 2023-03-09
Update examples to use apiVIPs and ingressVIPs

serngawy:


loganmc10:
I don't think this will work, you still have to specify both (this is a requirement of the `AgentClusterInstall`), like:
```
    apiVIP: 192.168.2.2
    apiVIPs:
    - 192.168.2.2
    - 1111:2222:3333:4444::1:1
    ingressVIP: 192.168.2.3
    ingressVIPs:
    - 192.168.2.3
    - 1111:2222:3333:4444::1:2
```
It really only makes sense to do so when you are setting up dual-stack VIPs

serngawy:
> I don't think this will work, you still have to specify both (this is a requirement of the `AgentClusterInstall`), like:
> 
> ```
>     apiVIP: 192.168.2.2
>     apiVIPs:
>     - 192.168.2.2
>     - 1111:2222:3333:4444::1:1
>     ingressVIP: 192.168.2.3
>     ingressVIPs:
>     - 192.168.2.3
>     - 1111:2222:3333:4444::1:2
> ```
> 
> It really only makes sense to do so when you are setting up dual-stack VIPs

ya you are correct, but it is strange that they mentioned apiVIP & ingressVIP are deprecated. How it is deprecated and then it is required.

PR #1436, opened 2023-03-13
infra: added github configuration file for dependabot with commit prefix

nishant-parekh:


nishant-parekh:
> hey, this is very nice indeed! but I'd prefer if you add a config file https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message and set the prefix to `depbot`

I made the changes per your suggestions..

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
yuvalk, line 11: why not daily?

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+    commit-message:
+      prefix: "depbot"
+  - package-ecosystem: "pip"
+      directory: "/ztp/resource-generator/tools/cachito"
+      schedule:
+        interval: "weekly"
```
yuvalk, line 17: why not daily?

nishant-parekh: Thought there might be too many PRs but a daily check on keeping dependencies up to date seems like a good idea none the less.. updated..

.github/dependabot.yml
nishant-parekh: updated

yuvalk:
can you plz make commit msg explain the why rather than the what?

SchSeba:
/override ci/prow/e2e-gcp-ovn

nice work!

PR #1438, opened 2023-03-15
ztp: Add one-shot time sync on boot

vitus133:
Add a one-shot boot-time attempt to synchronize the system clock. This will prevent nodes with a depleted RTC battery from becoming unusable and needing manual recovery.

/cc @browsell @imiller0 @lack


ztp/extra-manifests-builder/99-one-shot-time-sync/sync-time-once.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+SYNC_ATTEMPT_TIMEOUT_SEC=300
+
+/usr/sbin/chronyd -n -f /etc/chrony.conf -q -t $SYNC_ATTEMPT_TIMEOUT_SEC
+if [[ $? -ne 0 ]]; then
+  echo "failed to achieve time sync in ${SYNC_ATTEMPT_TIMEOUT_SEC}s"
```
lack, line 6: Does the chronyd executable not already say something like this if it fails?

ztp/extra-manifests-builder/99-one-shot-time-sync/sync-time-once.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+SYNC_ATTEMPT_TIMEOUT_SEC=300
```
lack, line 2: Instead of hard-coding this, we could easily allow environment overrides, which would let customers configure this timeout via a simple systemd drop-in, if 300s was not right for their environment:
```suggestion
SYNC_ATTEMPT_TIMEOUT_SEC=${SYNC_ATTEMPT_TIMEOUT_SEC:-300}
```

ztp/extra-manifests-builder/99-one-shot-time-sync/build.sh
```diff
@@ -0,0 +1,38 @@
...
+            Description=Sync time once
+            After=network.service
+            [Service]
+            Type=oneshot
+            TimeoutStartSec=infinity
+            ExecStart=/usr/local/bin/sync-time-once.sh
+            [Install]
+            WantedBy=multi-user.target
+          enabled: true
+          name: sync-time.service"
```
lack, line 38: Maybe let's name it `sync-time-once` so it's consistent with the script?

ztp/extra-manifests-builder/99-one-shot-time-sync/sync-time-once.sh
lack: Actually, if chronyd has good enough output on its own, we could eliminate this script entirely and just call chronyd directly in the systemd unit, which would make this a lot simpler!

ztp/extra-manifests-builder/99-one-shot-time-sync/build.sh
```diff
@@ -0,0 +1,38 @@
...
+    systemd:
+      units:
+        - contents: |
+            [Unit]
+            Description=Sync time once
+            After=network.service
+            [Service]
+            Type=oneshot
+            TimeoutStartSec=infinity
+            ExecStart=/usr/local/bin/sync-time-once.sh
```
lack, line 34: Can we change this to 
```suggestion
            Environment=SYNC_ATTEMPT_TIMEOUT_SEC=300
            ExecStart=/usr/sbin/chronyd -n -f /etc/chrony.conf -q -t $SYNC_ATTEMPT_TIMEOUT_SEC
```
and eliminate the script?

ztp/extra-manifests-builder/99-one-shot-time-sync/build.sh
```diff
@@ -0,0 +1,38 @@
...
+          overwrite: true
+          path: /usr/local/bin/sync-time-once.sh
+    systemd:
+      units:
+        - contents: |
+            [Unit]
+            Description=Sync time once
+            After=network.service
+            [Service]
+            Type=oneshot
```
lack, line 32: Unless we also add
```
RemainAfterExit=yes
```
This service will be listed as "dead" for the life of the system, which may be misleading.

ztp/extra-manifests-builder/99-one-shot-time-sync/build.sh
lack, line 38:
```suggestion
          name: sync-time-once.service"
```

ztp/extra-manifests-builder/99-one-shot-time-sync/sync-time-once.sh
vitus133: No, unfortunately not. It just says it exited.
For example:
```bash
# /usr/sbin/chronyd -n -f /etc/chrony.conf -q -t 1
2023-03-15T12:46:14Z chronyd version 4.1 starting (+CMDMON +NTP +REFCLOCK +RTC +PRIVDROP +SCFILTER +SIGND +ASYNCDNS +NTS +SECHASH +IPV6 +DEBUG)
2023-03-15T12:46:14Z Frequency 23.213 +/- 0.005 ppm read from /var/lib/chrony/drift
2023-03-15T12:46:15Z chronyd exiting
```
Same string is printed when it succeeds and exits due to -q flag

ztp/extra-manifests-builder/99-one-shot-time-sync/sync-time-once.sh
vitus133: Right, fixing...

ztp/extra-manifests-builder/99-one-shot-time-sync/build.sh
vitus133: +1

ztp/extra-manifests-builder/99-one-shot-time-sync/build.sh
vitus133: +1

ztp/extra-manifests-builder/99-one-shot-time-sync/build.sh
vitus133: Done

ztp/extra-manifests-builder/99-one-shot-time-sync/sync-time-once.sh
vitus133: done

ztp/source-crs/extra-manifest/99-sync-time-once-master.yaml
```diff
@@ -0,0 +1,28 @@
...
+      units:
+        - contents: |
+            [Unit]
+            Description=Sync time once
+            After=network.service
+            [Service]
+            Type=oneshot
+            TimeoutStartSec=infinity
+            Environment=SYNC_ATTEMPT_TIMEOUT_SEC=300
+            ExecStart=/usr/sbin/chronyd -n -f /etc/chrony.conf -q -t
```
lack, line 23: The variable didn't come through; probably needs to be escaped in build.sh

ztp/extra-manifests-builder/99-sync-time-once/build.sh
```diff
@@ -0,0 +1,31 @@
...
+      units:
+        - contents: |
+            [Unit]
+            Description=Sync time once
+            After=network.service
+            [Service]
+            Type=oneshot
+            TimeoutStartSec=infinity
+            Environment=SYNC_ATTEMPT_TIMEOUT_SEC=300
+            ExecStart=/usr/sbin/chronyd -n -f /etc/chrony.conf -q -t $SYNC_ATTEMPT_TIMEOUT_SEC
```
lack, line 26:
```suggestion
            ExecStart=/usr/sbin/chronyd -n -f /etc/chrony.conf -q -t \$SYNC_ATTEMPT_TIMEOUT_SEC
```

vitus133: This didn't help...

vitus133:
/unhold
(tested)

ztp/extra-manifests-builder/99-sync-time-once/build.sh
```diff
@@ -0,0 +1,31 @@
...
+      version: 3.2.0
+    systemd:
+      units:
+        - contents: |
+            [Unit]
+            Description=Sync time once
+            After=network.service
+            [Service]
+            Type=oneshot
+            TimeoutStartSec=${SYNC_ATTEMPT_TIMEOUT_SEC}
```
lack, line 25: This is even cleaner - Let chronyd run as long as it likes, and leave the termination up to systemd.  I like it!

ztp/source-crs/extra-manifest/99-sync-time-once-master.yaml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    systemd:
+      units:
+        - contents: |
+            [Unit]
+            Description=Sync time once
+            After=network.service
```
midu16, line 18: @vitus133 should we run this `Before=chronyd.service` ?

vitus133: We only run it if chronyd service is disabled:
https://github.com/openshift-kni/cnf-features-deploy/blob/f8ba7b1b77f0924ef87e371b2f428b971db696d1/ztp/extra-manifests-builder/99-sync-time-once/build.sh#L27

PR #1439, opened 2023-03-15
Fix go install ginkgo command

liornoy:
This commit changes the go install command, and removes the suffix "@v2.8.4" in order to make go install from the vendor and not hit the golang proxy.

This commit is required because of failures in building the cnf-tests image where the installer is in disconnected env.


cnf-tests/hack/build-test-bin.sh
```diff
@@ -18,7 +18,7 @@ DONT_REBUILD_TEST_BINS="${DONT_REBUILD_TEST_BINS:-false}"
 
 if ! which ginkgo; then
 	echo "Downloading ginkgo tool"
-	go install -mod=readonly github.com/onsi/ginkgo/v2/ginkgo@v2.8.4
+	go install github.com/onsi/ginkgo/v2/ginkgo
```
ffromani, line 21: we pinned it to a reason though, because we were bitten in the past by compatibility changes. Building from vendor is good enough - we pin packages here using go.mod. But what guarantees we build from vendor?
perhaps we need `-mod=vendor` or something like it?

openshift-ci/Dockerfile.tools
```diff
@@ -28,7 +28,7 @@ RUN mkdir ~/bin && \
     eval "$(gimme $GOVERSION)" && \
     cat $GIMME_ENV >> $HOME/.bashrc && \
     # get required golang tools and OC client
-    go install github.com/onsi/ginkgo/v2/ginkgo@v2.8.4 && \
+    go install github.com/onsi/ginkgo/v2/ginkgo && \
```
ffromani, line 31: same

cnf-tests/hack/build-test-bin.sh
liornoy: This will not download latest, but the vendor version.
see this from https://go.dev/ref/mod#go-install:
`At go 1.14 or higher, automatic [vendoring](https://go.dev/ref/mod#vendoring) may be enabled. If the file vendor/modules.txt is present and consistent with go.mod, there is no need to explicitly use the -mod=vendor flag.`

ffromani: ok, good enough then.

ffromani: nit: the "downloading" comment is thus now misleading, because we are building from `vendor/`

liornoy: right, good point. let me reword

fedepaol: I'd keep the mod=vendor too, to be on the safe side

liornoy: we don't need this on the `openshift-ci/Dockerfile.tools` I guess right?

openshift-ci/Dockerfile.tools
ffromani: this needs a second look

liornoy: it was before the ginkgo/v2 changes
`go install github.com/onsi/ginkgo/ginkgo@latest && \`

liornoy: so maybe we do want here to have :
`go install github.com/onsi/ginkgo/v2/ginkgo@v2.8.4` ?
or even the latest?

sshnaidm:
Please run cnf tests 4.13 job with this patch to check: https://github.com/openshift/release/pull/37475

liornoy:
@sshnaidm Thanks for the reply Sagi,
Are you asking me to update this PR and change the oc_client version in the openshift-ci dockerfile? [this](https://github.com/openshift-kni/cnf-features-deploy/blob/3a14be183b7982feb41473793c6afeb60989fbd8/openshift-ci/Dockerfile.tools#L34)

openshift-ci/Dockerfile.tools
liornoy: reverted back

ffromani:
/retest

still looks like infra issue

SchSeba:
looks good

/override ci/prow/e2e-gcp-ovn

PR #1444, opened 2023-03-23
cnf-tests: Make junit report readable by Prow

zeeke:
CI Operator needs junit output files to be in the
form "junit*.xml" to be shown on the job page.


evgenLevin:
lgtm

zeeke:
@SchSeba @ffromani @fedepaol please take a look.

ffromani (commented):
/lgtm

CI runs on this PR should be good enough validation for the symlinks we do in the second commit in the PR. Or do we need extra verification?

zeeke:
> CI runs on this PR should be good enough validation for the symlinks we do in the second commit in the PR. Or do we need extra verification?

I think so. @evgenLevin WDYT?

PR #1445, opened 2023-03-24
cnf-tests: Avoid clean pods after test in tuningcni tests

zeeke:
Cleaning resources in AfterEach decorator makes the reporter not collect test resources. It's preferable to do it in the clean section and at the end of the suite (i.e. in *Fixtures in features.go)

### DIFF
--- a/cnf-tests/testsuites/e2esuite/security/tuning_sriov.go
+++ b/cnf-tests/testsuites/e2esuite/security/tuning_sriov.go
@@ -35,7 +35,7 @@ var _ = Describe("[sriov] Tuning CNI integration", func() {
 		Expect(err).ToNot(HaveOccurred())
 	})
 
-	AfterEach(func() {
+	BeforeEach(func() {
 		namespaces.CleanPods(namespaces.SriovTuningTest, apiclient)
 	})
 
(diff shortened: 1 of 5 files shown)


zeeke:
Good point. I can split this between the variable move and the resource cleaning.

cnf-tests/testsuites/e2esuite/security/tuning_sriov.go
```diff
@@ -35,12 +31,10 @@ var _ = Describe("[sriov] Tuning CNI integration", func() {
...
 
 	execute.BeforeAll(func() {
-		err := namespaces.Create(SriovTestNamespace, apiclient)
+		err := namespaces.Create(namespaces.SriovTuningTest, apiclient)
 		Expect(err).ToNot(HaveOccurred())
-	})
 
-	AfterEach(func() {
-		namespaces.CleanPods(SriovTestNamespace, apiclient)
+		namespaces.CleanPods(namespaces.SriovTuningTest, apiclient)
```
SchSeba, line 37: we have this also on line 48 so we maybe we can remove it from here or the other way around?

SchSeba (commented):
just a small comment

SchSeba:
and please rebase :)

cnf-tests/testsuites/e2esuite/security/tuning_sriov.go
zeeke: Good point! As there are two test cases that create pods, I'll move it to the BeforeEach section

cnf-tests/testsuites/pkg/utils/reporter.go
```diff
@@ -95,6 +95,7 @@ func NewReporter(reportPath string) (*k8sreporter.KubernetesReporter, error) {
 		perfUtils.NamespaceTesting:              "performance",
 		namespaces.DpdkTest:                     "dpdk",
 		sriovNamespaces.Test:                    "sriov",
+		namespaces.SriovTuningTest:              "sriov",
```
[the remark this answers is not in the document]
zeeke: I guess so

SchSeba (approved):
/lgtm
/approve

PR #1448, opened 2023-03-28
OCPBUGS-9272: ztp: validate disk partitioning is applied only for SNO

pixelsoccupied:
/cc @imiller0 
/cc @lack


imiller0 (approved):
/lgtm

PR #1452, opened 2023-04-03
Remove the ptp parallel tests

liornoy:
This commit removes the ptp parallel tests from the cnftests
because those are supposed to run with extra ginkgo parameters
to handle the parallel setup.

Updated go.mod /vendor and docs.


SchSeba:
failing tests are not related to the PR

/override ci/prow/e2e-gcp-ovn

PR #1453, opened 2023-04-05
[wip] ztp: some code to get started with diffing

pixelsoccupied:
The idea here similar to how one would test public APIs

1. "record" mode.
    1. Feed PGT binary with all of customers' PGT (a dir with all site+common+group)
    2. run the PGT binary and because it's diff record mode..the file names generated will be generic (depended on the source filename)

2. "diff" mode
    1.  Feed PGT binary with all of [reference](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example/policygentemplates) PGT (common-ranGen.yaml  + example-sno-site.yaml + group-du-sno-ranGen.yaml)
    2. run PGT binary (3 cases to handle) (the filenames should be the same as 1)2) )
        1. Exact match e.g both record and diff mode generated the same file (/v-out/git-out/ClusterLogNS.yaml/Namespace-openshift-logging.yaml). Do a diff.  No new file generated.
        2. Partial match. Sometimes exactly match might not be possible just by looking at file name but the directory name maybe same. e.g `/v-out/git-out/SriovNetworkNodePolicy.yaml/SriovNetworkNodePolicy-sriov-MY-MID-BAND-openshift-sriov-network-operator.yaml` vs `SriovNetworkNodePolicy.yaml/SriovNetworkNodePolicy-sriov-THEIR-MID-BAND-openshift-sriov-network-operator.yaml`. Attempt to diff there. No new file generated
        3. No match (we have it but they are not using it). Create a new file to record that.

/cc @lack

### DIFF
--- a/ztp/policygenerator/policyGen/policyBuilder.go
+++ b/ztp/policygenerator/policyGen/policyBuilder.go
@@ -68,7 +68,7 @@ func (pbuilder *PolicyBuilder) Build(policyGenTemp utils.PolicyGenTemplate) (map
 					}
 					name = strings.Join(nameParts, "-")
 				}
-				output := path.Join(utils.CustomResource, policyGenTemp.Metadata.Name, name)
+				output := path.Join(sFile.FileName, name)
 				policies[output] = resources
 			} else {
 				// Generate a policy-wrapped CR, with the filename based on the policy and source filename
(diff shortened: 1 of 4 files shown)


pixelsoccupied:
Working a tool to do this task. closing this.

PR #1454, opened 2023-04-11
ztp: CNF-7798: git-first-search-path-for-ztp

sabbir-47:
source crs are looked first at git repo, then inside the ztp container

/cc @serngawy @imiller0

### DIFF
--- a/ztp/policygenerator/utils/filesHandler.go
+++ b/ztp/policygenerator/utils/filesHandler.go
@@ -1,7 +1,7 @@
 package utils
 
 import (
-	"io/ioutil"
+	"errors"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


ztp/policygenerator/utils/filesHandler.go
```diff
@@ -55,24 +55,35 @@ func (fHandler *FilesHandler) ReadSourceFile(fileName string) ([]byte, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	dir = filepath.Dir(ex)
-	ret, err = fHandler.ReadFile(dir + "/" + SourceCRsPath + "/" + fileName)
+	sourceCRPathGit := gitDir + "/" + SourceCRsPath + "/" + fileName
 
-	// added fail safe for test runs as `os.Executable()` will fail for tests
+	// path of the local executable
+	localDir, err = os.Executable()
```
serngawy, line 73: I believe we should read first the `os.ReadFile(sourceCRPathGit)`  if it fail to read the file then read local `os.Executable()`

sabbir-47: This is exactly what we are doing at line 80 below.

ztp/policygenerator/utils/filesHandler.go
```diff
@@ -55,24 +55,35 @@ func (fHandler *FilesHandler) ReadSourceFile(fileName string) ([]byte, error) {
...
+	sourceCRPathLocal := dir + "/" + SourceCRsPath + "/" + fileName
+
+	fileByte, err = os.ReadFile(sourceCRPathGit)
+	if errors.Is(err, os.ErrNotExist) {
+		fileByte, err := os.ReadFile(sourceCRPathLocal)
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, errors.New(fileName + " is not found both in Git path: " + sourceCRPathGit + " and in the ztp container path: " + sourceCRPathLocal)
 		}
-		ret, err = fHandler.ReadFile(dir + "/" + SourceCRsPath + "/" + fileName)
+		return fileByte, nil
```
imiller0, line 86: By returning nil here we are masking any other potential error cases (eg file permissions) from the os.ReadFile. Any others are very unlikely but it seems like we should use:
```suggestion
		return fileByte, err
```

imiller0 (commented):
/lgtm

serngawy (approved):

PR #1457, opened 2023-04-12
ztp: config crun as the default container runtime for 4.13+

Missxiaoguo:
Jira: [CNF-7823](https://issues.redhat.com/browse/CNF-7823)

### DIFF
--- a/ztp/source-crs/optional-extra-manifest/enable-crun-master.yaml
+++ b/ztp/source-crs/optional-extra-manifest/enable-crun-master.yaml
@@ -1,0 +1,10 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: ContainerRuntimeConfig
+metadata:
+ name: enable-crun-master
+spec:
+ machineConfigPoolSelector:
+   matchLabels:
+     pools.operator.machineconfiguration.openshift.io/master: ""
+ containerRuntimeConfig:
+   defaultRuntime: crun
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
@@ -61,3 +61,14 @@ spec:
               [service]
               service.stalld=start,enable
               service.chronyd=stop,disable
+    #
+    # These CRs are to enable crun on master and worker nodes for 4.13+ only
+    #
+    # Include these CRs in the group PGT instead of the common PGT to make sure
+    # they are applied after the operators have been successfully installed,
+    # however, it's strongly recommended to include these CRs as day-0 extra manifests
+    # to avoid the risky of an extra reboot.
+    - fileName: optional-extra-manifest/enable-crun-master.yaml
+      policyName: "config-policy"
+    - fileName: optional-extra-manifest/enable-crun-worker.yaml
+      policyName: "config-policy"
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
@@ -143,6 +143,17 @@ spec:
               [service]
               service.stalld=start,enable
               service.chronyd=stop,disable
+    #
+    # These CRs are to enable crun on master and worker nodes for 4.13+ only
+    #
+    # Include these CRs in the group PGT instead of the common PGT to make sure
+    # they are applied after the operators have been successfully installed,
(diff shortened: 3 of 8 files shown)


Missxiaoguo:
open quesiton for the location of the CRs being added for the day-2 configuration.

/cc @lack @imiller0 @serngawy

ztp/source-crs/optional-extra-manifest/enable-crun-master.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: machineconfiguration.openshift.io/v1
```
imiller0, line 1: Does everything under the source-crs directory get included into the generated container automatically? @nishant-parekh

imiller0 (commented):
Perhaps add a single copy of the ContainerRuntimeConfig CR to the source-crs? The user can update the name and matchLabel using the PGT overlay.

ztp/source-crs/optional-extra-manifest/enable-crun-master.yaml
nishant-parekh: yes that is correct:
https://github.com/openshift-kni/cnf-features-deploy/blob/c827f233f55e5c4ab5e3c42c907c6783a3ce00cf/ztp/resource-generator/Containerfile#L32

lack:
@imiller0 Since source-crs will include the subdirectories, users can write a policy to refer to `optional-extra-manifest/enable-crun-master.yaml` (I think) without having another copy in source-crs directly.

Missxiaoguo:
> Perhaps add a single copy of the ContainerRuntimeConfig CR to the source-crs? The user can update the name and matchLabel using the PGT overlay.

If so, maybe move the CR to the source-crs instead of optional-extra-manifest?

imiller0:
> @imiller0 Since source-crs will include the subdirectories, users can write a policy to refer to `optional-extra-manifest/enable-crun-master.yaml` (I think) without having another copy in source-crs directly.

I think that sounds like a good plan. I like that they show up in the extra manifests since it helps clarify that we recommend them for day-0 inclusion. In fact, for 4.13+ crun is the recommended runtime so I would suggest a couple changes here:
1. Add these manifests to the reference common.yaml PolicyGenTemplate (both the master and worker variant)
2. Add a commented entry to the reference SiteConfig recommending inclusion of crun manifests
3. Update the official documentation to explicitly call out that these should also be added as day-0 manifests to avoid an extra reboot.

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
```diff
@@ -67,3 +67,12 @@ spec:
...
           source: registry.redhat.io
+    #
+    # These CRs are to enable crun on master and worker nodes for 4.13+ only
+    #
+    # It's strongly recommended to include these CRs as day-0 extra manifests to
+    # avoid an extra reboot.
+    # - fileName: optional-extra-manifest/enable-crun-master.yaml
+    #   policyName: "config-policy"
+    # - fileName: optional-extra-manifest/enable-crun-worker.yaml
+    #   policyName: "config-policy"
```
imiller0, line 78: Because this is the reference configuration for 4.13 I would uncomment these and update the note above to say that for 4.12 and prior these lines should not be included.

ztp/gitops-subscriptions/argocd/example/policygentemplates/common-ranGen.yaml
```diff
@@ -67,3 +67,12 @@ spec:
...
           source: registry.redhat.io
+    #
+    # These CRs are to enable crun on master and worker nodes for 4.13+ only
+    #
+    # It's strongly recommended to include these CRs as day-0 extra manifests to
+    # avoid an extra reboot.
+    - fileName: optional-extra-manifest/enable-crun-master.yaml
+      policyName: "config-policy"
+    - fileName: optional-extra-manifest/enable-crun-worker.yaml
+      policyName: "config-policy"
```
imiller0, line 78: From a logical standpoint the common configuration is the right place for this. However, this will cause a reboot due to the runtime change. If that reboot occurs while day-2 operators are installing it will create a race condition that can cause those installs to fail. For that reason I think we need to move this to the group config so that it will occur after the operator installation.
If the user follows our advice and puts this in the day0 machine config they won't run into this race at installation. However if they choose not to do it day0, or if these get applied as part of an upgrade where operators are upgraded (and the user keeps our initial "wave" ordering) they will get hit by this. Please add a note to the group indicating why it is there.

Missxiaoguo: Very good point! Will move to group.

imiller0 (approved):
/lgtm

PR #1458, opened 2023-04-12
ztp: OCPBUGS-8718: Unload idrma driver for DU profile

browsell:
Signed-off-by: Brent Rowsell <browsell@redhat.com>


imiller0 (approved):
/lgtm

PR #1463, opened 2023-04-17
[release-4.12] OCPBUGS-14297: ztp: Add soaking annotation to validator PGTs

sudomakeinstall2:
- reduce compliant interval for validators
- patch validator PGTs to add soaking annotation


imiller0 (approved):
/lgtm

PR #1481, opened 2023-04-25
CNF-8518: ztp: Update DefaultCatSrc with Status and add depreciation warning in the rest of the cat sources

pixelsoccupied:
Update DefaultCatSrc with Status and add depreciation warning in rest of the CatalogueSource (we are meant to use just the  one)
/cc @imiller0 
/cc @jc-rh

### DIFF
--- a/ztp/policygenerator-kustomize-plugin/README.md
+++ b/ztp/policygenerator-kustomize-plugin/README.md
@@ -33,7 +33,6 @@ kustomize/
                     ├── AmqSubscriptionNS.yaml
                     ├── AmqSubscriptionOperGroup.yaml
                     ├── AmqSubscription.yaml
-                    ├── ClusterLogCatSource.yaml
                     ├── ClusterLogForwarder.yaml
                     ├── ClusterLogging.yaml
                     ├── ClusterLogNS.yaml
@@ -50,7 +49,6 @@ kustomize/
                     ├── MachineConfigSctp.yaml
                     ├── OperatorHub.yaml
                     ├── PerformanceProfile.yaml
-                    ├── PtpCatSource.yaml
                     ├── PtpConfigMaster.yaml
                     ├── PtpConfigSlaveCvl.yaml
                     ├── PtpConfigSlave.yaml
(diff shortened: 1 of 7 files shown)


imiller0 (commented):
/lgtm
Perhaps we should go ahead and deprecate all but DefaultCatSource now? It is confusing to have only 1 catalogsource in the reference configuration but carry one per subscription.

pixelsoccupied:
> deprecate all but DefaultCatSource

@imiller0 I removed all the CRs that are not in reference. Not sure if there's any way to add a depreciation warning with the current stack before removing though. wdyt?

ztp/policygenerator/utils/utils.go
```diff
@@ -1,21 +1,36 @@
...
+	DefaultCompliantEvaluationInterval    = "10m"
+	DefaultNonCompliantEvaluationInterval = "10s"
+	DisableEvaluationInterval             = "never"
+	// DefaultComplianceType ComplianceType of "mustonlyhave" uses significant CPU to enforce. Default to
+	// "musthave" so that we realize the CPU reductions unless explicitly told otherwise
+	DefaultComplianceType = "musthave"
+	// ZtpDeployDeprecationWarningKey annotation key for when source-cr is set to be deprecated. See ZtpDeployDeprecatedSourceCRFileNames for full list
+	ZtpDeployDeprecationWarningKey = "ran.openshift.io/deprecated"
+	// ZtpDeployDeprecationWarningValue annotation value for when source-cr is set to be deprecated. See ZtpDeployDeprecatedSourceCRFileNames for full list
+	ZtpDeployDeprecationWarningValue = "true"
```
imiller0, line 22: Would it make sense to have a longer message here to indicate why/when it is deprecated. ie, some sort of informational text:
`ran.openshift.io/deprecation-warning: "Source CR StorageCatSource is deprecated. Use DefaultCatSource instead"`

imiller0 (commented):
I think we need to consider what happens when managing multiple versions simultaneously. policyGen may be building policies from 4.n source CRs and 4.n+1 source CRs simultaneously. In that case the n+1 should contain the deprecation warning and the n should not. Maybe we can move the annotation into the source CR and trigger off of that?

imiller0 (approved):
/lgtm

PR #1482, opened 2023-04-25
CNF-7424: Add ACM policy generator examples

serngawy:


ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-common-ranGen.yaml
```diff
@@ -0,0 +1,69 @@
...
+policies:
+  # Create operators policies that will be installed in all clusters
+- name: common-subscriptions-policy-latest
+  policyAnnotations:
+    ran.openshift.io/ztp-deploy-wave: "2"
+  manifests:
+    - path: source-crs/SriovSubscriptionNS.yaml
+    - path: source-crs/PtpSubscriptionNS.yaml
+    - path: source-crs/ClusterLogNS.yaml
+    - path: source-crs/StorageNS.yaml
```
imiller0, line 37: Minor nit: I think it is better to keep the NS and operatorstatus with the subscription. Customers who want to opt out of an operator will need to comment out the ns, subscription, opergroup, and status which is easier to do if they are together. I like the re-ordering though to match dependency

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-common-ranGen.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - path: source-crs/SriovSubscription.yaml
+    - path: source-crs/SriovSubscriptionOperGroup.yaml
```
imiller0, line 39: Switch these two so that the opergroup is created before the subscription (sub needs the opergroup)

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-common-ranGen.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - path: source-crs/PtpSubscriptionOperGroup.yaml
+    - path: source-crs/ClusterLogOperGroup.yaml
+    - path: source-crs/ClusterLogSubscription.yaml
+    - path: source-crs/StorageOperGroup.yaml
+    - path: source-crs/StorageSubscription.yaml
+    - path: source-crs/SriovOperatorStatus.yaml
+    - path: source-crs/PtpOperatorStatus.yaml
+    - path: source-crs/ClusterLogOperatorStatus.yaml
+    - path: source-crs/StorageOperatorStatus.yaml
+- name: common-config-policy-latest
```
imiller0, line 50: If the wave of this section is "1" it would make sense to put it above the wave "2" content above.

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-example-multinode-site.yaml
```diff
@@ -0,0 +1,77 @@
...
+  # categories: []
+  #controls:
+  #  - PR.DS-1 Data-at-rest
+  namespace: ztp-site
+  # Use an existing placement rule so that placement bindings can be consolidated
+  placement:
+    labelSelector:
+      # These policies will correspond to all clusters with this label:
+      site: "example-multinode"
+      # the below label (du-profile: "latest") might not necassery as the site policy applied to a single site.
```
imiller0, line 17: I agree with the comment, but for consistency I like that we have this label here. I would recommend removing the comment.

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-example-multinode-site.yaml
```diff
@@ -0,0 +1,77 @@
...
+    exclude:
+      - kube-*
+    include:
+      - '*'
+  evaluationInterval:
+    compliant: 10m
+    noncompliant: 10s
+policies:
+  # Create operators policies that will be installed in all clusters
+- name: example-multinode-site-config-policy-latest
```
imiller0, line 32: We have run into issues in the past with long policy names. Is this equivalent to the site policy name in the original?

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-group-du-3node-validator-ranGen.yaml
```diff
@@ -0,0 +1,40 @@
...
+  remediationAction: inform
+  severity: low
+  # standards: []
+  namespaceSelector:
+    exclude:
+      - kube-*
+    include:
+      - '*'
+  evaluationInterval:
+    compliant: 5s
```
imiller0, line 32: There is a comment in the original PGT that indicates this low value is only valid (acceptable) if the binding rules disconnect the policy based on ztp-done. Otherwise we end up with high CPU use at steady state.

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-group-du-sno-validator-ranGen.yaml
```diff
@@ -0,0 +1,40 @@
...
+  remediationAction: inform
+  severity: low
+  # standards: []
+  namespaceSelector:
+    exclude:
+      - kube-*
+    include:
+      - '*'
+  evaluationInterval:
+    compliant: 5s
```
imiller0, line 32: Same note on a comment needed here to explain the validity of 5s interval

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-group-du-standard-validator-ranGen.yaml
```diff
@@ -0,0 +1,40 @@
...
+  remediationAction: inform
+  severity: low
+  # standards: []
+  namespaceSelector:
+    exclude:
+      - kube-*
+    include:
+      - '*'
+  evaluationInterval:
+    compliant: 5s
```
imiller0, line 32: Same note on a comment needed here to explain the validity of 5s interval

ztp/source-crs/ClusterLogForwarder.yaml
```diff
@@ -6,11 +6,11 @@ metadata:
   annotations:
     ran.openshift.io/ztp-deploy-wave: "10"
 spec:
-  outputs: $outputs
+  outputs:
```
imiller0, line 9: Is an empty value legal here or should this line simply be removed (or commented out)? The original policyGen would strip this line out if the user doesn't supply a value, so it seems like the equivalent would be to simply omit the line altogether (user can still provide it via overlay). I think this would ensure that the source-cr behaved in the same way for original policyGen and ACM version as well as still providing backwards compatible behavior for the original policyGen.

Same thing below for pipelines

ztp/source-crs/ClusterLogging.yaml
```diff
@@ -10,8 +10,11 @@ spec:
   curation:
     type: "curator"
     curator:
-      schedule: $schedule
-  collection: $collection
+      schedule: ""
+  collection:
+    logs:
+      type: ""
+      fluentd: {}
```
imiller0, line 17: Same comment on perhaps just commenting out the line?

ztp/source-crs/DisconnectedICSP.yaml
```diff
@@ -5,5 +5,6 @@ metadata:
     annotations:
         ran.openshift.io/ztp-deploy-wave: "1"
 spec:
-    repositoryDigestMirrors: $mirrors
+    repositoryDigestMirrors:
+    - mirrors
```
imiller0, line 9: Same note on commenting this out instead.

ztp/source-crs/PerformanceProfile2.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: performance.openshift.io/v2
```
imiller0, line 1: The file should be named after its intended use. Having a "2" version doesn't indicate when this should be used vs the original. Same comment for other "2" versions below.

ztp/source-crs/SriovNetwork.yaml
```diff
@@ -6,14 +6,14 @@ metadata:
   annotations:
     ran.openshift.io/ztp-deploy-wave: "100"
 spec:
-  resourceName: ""
+#  resourceName: ""
```
imiller0, line 9: Why do these need to be commented out?

ztp/source-crs/validatorCRs/informDuValidatorMaster.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfigPool
+metadata:
+  name: master
```
imiller0, line 4: If the reference configuration uses "worker" for mcp, do we need the master version?

imiller0 (commented):
Nice work porting the profile over!

A couple of things I think we should still consider/look at:
- The changes to the source-crs. We need to make sure that the source-crs remain backwards compatible, ideally without having to make copies of them. For the few copies below I think the issue comes down to how the original policyGen made use of $mcp. Is there a way that we can resolve/solve that difference without making a copy of the source-crs?
- I think we need to add a check in the CI for the generated results of these policies compared to the original policyGen versions. This will ensure there is never any drift between the two. Obviously we can't do a simple text/string diff, but the set of generated object-templates within the policies should be identical. Can we write an intelligent diff utility which would compare the two sets of generated policies and use that in CI?

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-common-ranGen.yaml
serngawy: okay, done

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-example-multinode-site.yaml
serngawy: the acm policygen has the check for the policy name length (limit 63 char) no worry about that.

ztp/source-crs/SriovNetwork.yaml
serngawy: cause it will be added as empty value field , no need to do that.

ztp/source-crs/PerformanceProfile2.yaml
serngawy: I rename the files with -SetSelector , open for any  other name that make sense.

ztp/source-crs/DisconnectedICSP.yaml
serngawy: yes, done

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-common-ranGen.yaml
serngawy: good point, done.

ztp/gitops-subscriptions/Makefile
```diff
@@ -0,0 +1,48 @@
...
+$(KUSTOMIZE_BIN):
+	@if [[ $(KUSTOMIZE) == $(KUSTOMIZE_BIN) ]] && [ ! -f $(KUSTOMIZE) ]; then \
+		echo "kustomize not installed getting kustomize v"$(KUSTOMIZE_VERSION) \
+		&& cd $(KUSTOMIZE_DIR) && curl -m 600 -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash -s $(KUSTOMIZE_VERSION); \
+	fi
+
+test: gen-files
+	@echo "ZTP: Test difference between acm and ztp generated policies"
+	# Run the ztp diff tool to compare the policies
+	$(MAKE) -C ../tools/policy-object-template-diff ACM_PATH=../../gitops-subscriptions/$(ACM_POLICYGEN_EX_DIR)/out/ ZTP_PATH=../../gitops-subscriptions/$(POLICYGEN_EX_DIR)/out/ run
```
pixelsoccupied, line 38: I would call tool's [clean](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/tools/policy-object-template-diff/Makefile#LL10C1-L10C6) right after this

pixelsoccupied (commented):
only reviewed the new stuff related to diff. a small comment otherwise lgtm!

ztp/gitops-subscriptions/Makefile
serngawy: I don't think it will make a difference, the ci-job container get destroyed after  the job done. Having less execution time is better I guess.

pixelsoccupied: Yep..it's to avoid accidental /bin commits locally. Maybe an adding a gitignore will work too.

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/README.md
```diff
@@ -0,0 +1,27 @@
+## ACM PolicyGenerator
+The [ACM PolicyGenerator](https://github.com/stolostron/policy-generator-plugin) examples under ztp/gitops-subscriptions/argocd/example/acmpolicygenerator are defining the DU profile policies using [ACM PolicyGenerator reference API](https://github.com/stolostron/policy-generator-plugin/blob/main/docs/policygenerator-reference.yaml). These examples will generate ACM policies similar to the DU profile policies generated from [policygentemplates example](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example/policygentemplates)
```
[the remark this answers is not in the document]
lack: If this is now checked and guaranteed to be identical, we should say so.

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+##### Setup ArgoCD Application
+
+The DU profile ACM PolicyGenerator examples can be used with the same [policies-app](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/deployment/policies-app.yaml) that is used to deploy the DU profile policygentemplates examples. For more info how to setup the ArgoCD policies application follow the [ReadMe](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/README.md) section "Preparation of Hub cluster for ZTP". The Git repo that will be used with the ArgoCD policies application should contain the source-crs directory co-exist with the DU profile ACM PolicyGenerator examples as below.
```
sabbir-47, line 6:
```suggestion
The DU profile ACM PolicyGenerator examples can be used with the same [policies-app](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/deployment/policies-app.yaml) that is used to deploy the DU profile policygentemplates examples. For more info how to setup the ArgoCD policies application follow the [ReadMe](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/README.md) section "Preparation of Hub cluster for ZTP". The Git repo that will be used with the ArgoCD policies application should contain the source-crs directory and must co-exist with the DU profile ACM PolicyGenerator as shown below as example
```

sabbir-47 (commented):
Some minor comments in the documentation

PR #1484, opened 2023-04-28
cnf-tests: debug RPS tests DO NOT MERGE

zeeke:
Temporary adding debug logs when 
```
[FAIL] [rfe_id:27368][performance] RPS configuration [It] [test_id: 59572] Check RPS Mask is applied to atleast one single rx queue on all veth interface
```
fails

cc @jlojosnegros


zeeke:
The test is mixing interface names from different worker nodes. Fixed in:
- https://github.com/openshift/cluster-node-tuning-operator/pull/641

PR #1486, opened 2023-05-02
OCPBUGS-12965: ztp: update isRdma to false to match default value suggested in SriovNetworkNodePolicy doc

pixelsoccupied:
We should use the default config as mentioned in the official doc. See [this](https://docs.openshift.com/container-platform/4.12/networking/hardware_networks/configuring-sriov-device.html) for the latest.


bartwensley (commented):
Could you just add a sentence or two to the commit message explaining the "why" of this change? I looked at the bug and there isn't an explanation there either.

bartwensley (approved):
/lgtm

imiller0 (approved):
/lgtm

PR #1487, opened 2023-05-02
ztp: CNF-7722 Fix ordering of common-subscriptions policies

tliu2021:
This update reorders the subscriptions-policy to reflect the dependency order.


imiller0 (approved):
/lgtm

PR #1490, opened 2023-05-04
ztp: Remove nohz_full in the DU profile

browsell:
Signed-off-by: Brent Rowsell <browsell@redhat.com>


ztp/source-crs/TunedPerformancePatch.yaml
bartwensley, line 1: I think this has moved - I don't see the source-crs directory on the master branch: https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp

I think the new CR is here: https://github.com/openshift-kni/telco-reference/blob/main/telco-ran/configuration/source-crs/node-tuning-operator/TunedPerformancePatch.yaml

I don't see nohz_full in that file, so maybe it has already been removed?

bartwensley (commented):
Maybe this has already been done?

PR #1441, opened 2023-03-21
feat: add support for cpu partitioning on site config

eggfoobar:
add new flag for cpu partitioning mode
add deprecation warning for situations where cpuset is defined for nodes

/hold
/cc @imiller0

PR #1495, opened 2023-05-08
[release-4.13] Bug OCPBUGS-13313: Backport cpu partitioning siteconfig

eggfoobar:
PR back ports https://github.com/openshift-kni/cnf-features-deploy/pull/1441 and https://github.com/openshift-kni/cnf-features-deploy/pull/1497

/assign @imiller0

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/siteConfig.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfig.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/coreos/go-systemd/unit"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 const localExtraManifestPath = "extra-manifest"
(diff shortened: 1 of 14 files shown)


imiller0:
/unhold
4.13 GA artifacts are built. Backport can merge for z-stream.

imiller0:
Just to confirm: If the user continues to use cpuset (and not cpuPartitioningMode) in their SiteConfig the tool will continue to generate the MachineConfigs. In this case workload partitioning continues to be enabled "the old way" and there is no actual impact, correct?

eggfoobar:
That's correct, because of this logic below we still get the old way for SNO and then you can go down the new path with techpreview on 4.13 for non SNO deployments. We also do have some unit tests in place to that will validate that behavior.  https://github.com/openshift-kni/cnf-features-deploy/blob/8d4e1d7b59e35e9bc8792e0440df105d9534b383/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go#L424-L444

imiller0 (commented):
/lgtm

imiller0 (approved):

PR #1498, opened 2023-05-09
OCPBUGS-13981: feat: update example to use new cpu partitioning mode

eggfoobar:


ztp/gitops-subscriptions/argocd/example/siteconfig/example-sno.yaml
```diff
@@ -45,6 +45,7 @@ spec:
       - 1001:2::/112
     additionalNTPSources:
       - 1111:2222:3333:4444::2
+    cpuPartitioningMode: AllNodes # Initiates the cluster for workload partitioning.
```
pixelsoccupied, line 48: Thanks for the example! Couple of Qs

- So to specify the cores we just simply use PerformanceProfile? Can we note that info here? 

- Can we update this doc too? https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/WorkloadPartitioning.md

imiller0 (approved):
/lgtm

PR #1504, opened 2023-05-15
OCPBUGS-13070: set gitOps operator channel

serngawy:


ztp/gitops-subscriptions/argocd/deployment/openshift-gitops-operator.yaml
```diff
@@ -9,9 +9,8 @@ metadata:
   name: openshift-gitops-operator
   namespace: openshift-operators
 spec:
-  channel: stable
+  channel: gitops-1.8
```
yliu127, line 12: shall we consider using "latest" channel instead of versioned channel?

serngawy: Well, the idea to stay with latest-1 version and update the channel with every release.

imiller0 (commented):
/lgtm

PR #1506, opened 2023-05-16
OCPBUGS-6763: ztp: Cleanup some source CRs

Missxiaoguo:
- 'nohz_full=${isolated_cores}' is removed from TunedPerformancePatch.yaml as it's covered in the auto-created tuned patch for the performanceProfile already.

- The reference to ConsoleOperatorDisable.yaml is removed from the PGT example as the console operator is not installed in the first place when disabling baselineCapabilitySet via SiteConfig

- 'disableDrain: false' is removed from SriovOperatorConfig.yaml as it's set automatically by the operator. It's marked to true when the cluster is SNO and false for other types

### DIFF
--- a/ztp/source-crs/SriovOperatorConfig.yaml
+++ b/ztp/source-crs/SriovOperatorConfig.yaml
@@ -8,7 +8,6 @@ metadata:
 spec:
   configDaemonNodeSelector:
     "node-role.kubernetes.io/$mcp": ""
-  disableDrain: false
   # Injector and OperatorWebhook pods can be disabled (set to "false") below
   # to reduce the number of management pods. It is recommended to start with the 
   # webhook and injector pods enabled, and only disable them after verifying the
--- a/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-3node-ranGen.yaml
@@ -47,13 +47,10 @@ spec:
       spec:
         profile:
           - name: performance-patch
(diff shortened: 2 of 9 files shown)


Missxiaoguo:
I am debating if we should just remove the ConsoleDisableOperator.yaml reference from the PGT example, but keep the CR . For the user who forgets to add the baselineCapabilitySet in the siteConfig, it provides the opportunity to disable it on day2.

imiller0:
> I am debating if we should just remove the ConsoleDisableOperator.yaml reference from the PGT example, but keep the CR . For the user who forgets to add the baselineCapabilitySet in the siteConfig, it provides the opportunity to disable it on day2.

I think this makes sense (keeping the CR for now). This also ensures that the current container can be used in deploying clusters on prior releases. Essentially deprecating and removing 2 releases from now.

ztp/source-crs/SriovOperatorConfig.yaml
```diff
@@ -8,7 +8,6 @@ metadata:
 spec:
   configDaemonNodeSelector:
     "node-role.kubernetes.io/$mcp": ""
-  disableDrain: false
```
imiller0, line 11: Did you confirm that for both SNO and multi-node clusters the Operator will set this to the correct value if not supplied by the user?

Missxiaoguo: > Did you confirm that for both SNO and multi-node clusters the Operator will set this to the correct value if not supplied by the user?

Yes, have confirmed. Removing this line keeps the exact the same setting as today. For SNO, disableDrain is auto-set to true. For muli-node clusters, this field will be false(by default)

imiller0 (approved):
/lgtm

PR #1501, opened 2023-05-10
Implement the cnf-tests split

liornoy:
This PR implements the cnf-tests split: https://github.com/openshift-kni/kni-enhancements/blob/main/enhancements/cnf-tests-split.md

A commit updating the READMEs will be created in a different PR for the sake of urgency.

**Suggestion for the kind reviewer:**
Best reviewed by splitting to two rounds (preferably with a break between):
1. Review the five commits without `cnf-tests: Modify the test entrypoint scripts`
those are self-contained, pretty straight forward and come first chronologically.
2. Review the `cnf-tests: Modify the test entrypoint scripts` commit - that includes the new logic
for invoking the tests (bash script).

PR #1510, opened 2023-05-18
cnf-tests: Rewrite the cnf-tests container to contain only the latency tests

jlojosnegros:
The cnf-tests split implementation #1501  breaks the cnf-tests container.
We don't build the test binaries anymore nor run the functional tests in a containerized way.

cnf-tests container is a supported product for running latency tests, 
so we need to rewrite the cnf-tests container Dockerfiles to include only the latency test.

Additionally it would be a god time to improve user interface to allow users 
to easily invoke the tests without providing many different parameters and environment variables, 
thus providing better UX and preventing errors.

### DIFF
--- a/cnf-tests/Dockerfile
+++ b/cnf-tests/Dockerfile
@@ -1,35 +1,3 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder
-WORKDIR /go/src/github.com/openshift-kni/cnf-features-deploy
-COPY . .
-RUN make test-bin
-RUN git rev-list -1 HEAD > ./cnf-tests/bin/cnftests-sha.txt
-
-FROM quay.io/openshift/origin-oc-rpms:4.13 AS oc
-
-FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder-stresser
-ENV PKG_NAME=github.com/openshift-kni/cnf-features-deploy
-ENV PKG_PATH=/go/src/$PKG_NAME
-ENV TESTER_PATH=$PKG_PATH/cnf-tests/pod-utils/stresser
-
-RUN mkdir -p $PKG_PATH
-
-COPY . $PKG_PATH/
-WORKDIR $TESTER_PATH
-
-RUN go build -mod=vendor -o /stresser
-
-FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder-sctptester
-ENV PKG_NAME=github.com/openshift-kni/cnf-features-deploy
-ENV PKG_PATH=/go/src/$PKG_NAME
-ENV TESTER_PATH=$PKG_PATH/cnf-tests/pod-utils/sctptester
-
-RUN mkdir -p $PKG_PATH
-
-COPY . $PKG_PATH/
-WORKDIR $TESTER_PATH
-
-RUN go build -mod=vendor -o /sctptest
-
 # build latency-test's runner binaries
 FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder-latency-test-runners
 
@@ -46,6 +14,13 @@ RUN go build -mod=vendor -o /oslat-runner oslat-runner/main.go && \
     go build -mod=vendor -o /cyclictest-runner cyclictest-runner/main.go && \
     go build -mod=vendor -o /hwlatdetect-runner hwlatdetect-runner/main.go
 
+# build latency testing suite
+FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS go-builder
+WORKDIR /app
+COPY . .
+RUN make test-bin
+
+
 # Build latency-test binaries
 FROM centos:7 as builder-latency-test-tools
 
@@ -61,15 +36,16 @@ RUN yum install -y numactl-devel make gcc && \
     cp hwlatdetect /hwlatdetect && \
     cp cyclictest /cyclictest
 
+FROM quay.io/openshift/origin-oc-rpms:4.13 AS oc
+
+
+# Final image
 FROM centos:7
 
 # python3 is needed for hwlatdetect
 RUN yum install -y lksctp-tools iproute libhugetlbfs-utils libhugetlbfs tmux ethtool ping numactl-libs linuxptp iperf3 python3 nc iptables
 
 RUN mkdir -p /usr/local/etc/cnf
-
-COPY --from=builder-stresser /stresser /usr/bin/stresser
-COPY --from=builder-sctptester /sctptest /usr/bin/sctptest
 
 COPY --from=builder-latency-test-runners /oslat-runner /usr/bin/oslat-runner
 COPY --from=builder-latency-test-tools /oslat /usr/bin/oslat
@@ -81,14 +57,15 @@ COPY --from=builder-latency-test-runners /hwlatdetect-runner /usr/bin/hwlatdetec
 COPY --from=builder-latency-test-tools /hwlatdetect /usr/bin/hwlatdetect
 
 COPY --from=oc /go/src/github.com/openshift/oc/oc /usr/bin/oc
-COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/bin/cnftests /usr/bin/cnftests
-COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/bin/configsuite /usr/bin/configsuite
-COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/bin/validationsuite /usr/bin/validationsuite
-COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/entrypoint/test-run.sh /usr/bin/test-run.sh
-COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/bin/mirror /usr/bin/mirror
-COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/mirror/images.json /usr/local/etc/cnf
-COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/bin/cnftests-sha.txt /usr/local/etc/cnf
+
+COPY --from=go-builder /app/cnf-tests/bin/mirror /usr/bin/mirror
+COPY --from=go-builder /app/cnf-tests/mirror/images.json /usr/local/etc/cnf
+COPY --from=go-builder /app/cnf-tests/entrypoint/test-run.sh /usr/bin/test-run.sh
+
 COPY tools/oot-driver/ /usr/src/oot-driver/
+
+# Copy test suite
+COPY --from=go-builder /app/cnf-tests/submodules/cluster-node-tuning-operator/build/_output/bin/latency-e2e.test /usr/bin/latency-e2e.test
 
 ENV SUITES_PATH=/usr/bin/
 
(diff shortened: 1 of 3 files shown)


liornoy:
Thanks @jlojosnegros. the docker build works on my end.
I'm not qualified to indicate if it contains all the tests required for the latency tests product.
I'll let a compute team member verify it.
Are you planning to work on how a client invokes this test next?

jlojosnegros:
> Thanks @jlojosnegros. the docker build works on my end. I'm not qualified to indicate if it contains all the tests required for the latency tests product. I'll let a compute team member verify it. Are you planning to work on how a client invokes this test next?

About the latency tests, maybe @mrniranjan or @yanirq could take a look and sees if those are all the tests that need to be run.

About the invocation ... I have changed both `build-test-bin.sh` and `entrypoint/test-run.sh` and now the user should be able to use the image just specifying  the KUBECONFIG 
PTAL and tell me WDYT.

cnf-tests/entrypoint/test-run.sh
```diff
@@ -2,20 +2,6 @@
...
-for suite in "${suites[@]}"; do
-    if [ "$DISCOVERY_MODE" == "true" ] &&  [ "$suite" == "configsuite" ]; then
-        echo "Discovery mode enabled, skipping setup"
-        continue
-    fi
-    echo running "$SUITES_PATH/$suite" "$@"
-    "$SUITES_PATH/$suite" "$@"
-done
+pushd /usr/bin/latency_testing/one/two/three/four/five/
+LATENCY_TEST_RUN=true DISCOVERY_MODE=true FEATURES=performance ./5_latency_testing.test --ginkgo.no-color --ginkgo.vv --ginkgo.flake-attempts 2 -test.failfast --ginkgo.focus "\[performance\]\ Latency\ Test" --ginkgo.junit-report ~/tmp/artifacts
```
liornoy, line 6: Don't we want to invoke `latency-e2e.test` as well?

liornoy: In the origin `test-run.sh` we had: `"$SUITES_PATH/$suite" "$@"` which passed to the test binary
all the parameters we invoked the `test-run.sh` with.
This allowed us to pass the --junit and --report flags and chose where to output the reports.
Here you hardwire only the default `ginkgo.junit-report` to ~/tmp/artifacts. 
Should we keep the `$@` so that the user could specify values for the junit and reporter flags?

liornoy: Not only that but after this change, clients should be able to do the same as before, by the [documentation](https://docs.openshift.com/container-platform/4.13/scalability_and_performance/cnf-performing-platform-verification-latency-tests.html):
1. Running hwlatdetect
2. Running cyclictest
3. Running oslat
4. Generating a latency test failure report
5. Generating a JUnit latency test report
6. Running latency tests on an SNO
7. Running latency tests in a disconnected cluster
 
 So I think the `$@` is necessary for more than just the reports.
 And regarding all the requirements (1-7) we should remember to test
 if each one is working as usual after this change.

jlojosnegros: After a talk with @shajmakh I have found I was approaching this the wrong way, so forget about my answer.
I need to rework the PR

liornoy: ok, sounds good! thanks for the update.

jlojosnegros: > 

yeah that would not be a problem

liornoy:
lgtm thanks, I'll wait for another approval from someone else and I'll cherry-pick it to the cnf-tests split PR

ffromani:
if this is ready for review please remove the `[wip]` tag

ffromani:
are the build failures expected? (e.g. `images` and `ci` lanes)

jlojosnegros:
> are the build failures expected? (e.g. `images` and `ci` lanes)

Yes they are expected. 
This change is expecting a totally different folder layout for the tests based on git submodules.
it should work as part of  @liornoy change #1501 
https://github.com/openshift-kni/cnf-features-deploy/pull/1510#issuecomment-1558628631

cnf-tests/entrypoint/test-run.sh
```diff
@@ -3,19 +3,19 @@ set -e
 # Setting -e is fine as we want both config and validiation to succeed
 # before running the "real" tests.
 
-TEST_SUITES=${TEST_SUITES:-"validationsuite configsuite cnftests"}
+LATENCY_TEST_RUN="${LATENCY_TEST_RUN:true}"
```
Tal-or, line 6: Did you try to set a default value here?
Because it should be `"${LATENCY_TEST_RUN:=true}"` note the `=`

cnf-tests/entrypoint/test-run.sh
```diff
@@ -3,19 +3,19 @@ set -e
...
+DISCOVERY_MODE="${DISCOVERY_MODE:true}"
```
Tal-or, line 7: dito

cnf-tests/entrypoint/test-run.sh
```diff
@@ -3,19 +3,19 @@ set -e
...
+FEATURES="${FEATURES:performance}"
```
Tal-or, line 8: dito

cnf-tests/hack/build-test-bin.sh
```diff
@@ -16,27 +16,23 @@ export GOFLAGS="${GOFLAGS:-"-mod=vendor"}"
...
-
-build_and_move_suite "e2esuite" "./bin/cnftests"
-build_and_move_suite "configsuite" "./bin/configsuite"
-build_and_move_suite "validationsuite" "./bin/validationsuite"
-
 if [ "$DONT_REBUILD_TEST_BINS" == "false" ] || [ -f ./cnf-tests/bin/mirror ]; then
+  echo "Building mirror ... "
   go build -o ./bin/mirror mirror/mirror.go
+  retVal=$?
+  if [ $retVal -ne 0 ]; then
```
Tal-or, line 23: those checks aren't necessary because you have `set -e` at the beginning of the file.

cnf-tests/hack/build-test-bin.sh
```diff
@@ -16,27 +16,23 @@ export GOFLAGS="${GOFLAGS:-"-mod=vendor"}"
...
+    echo "Error while building mirror: ${retVal}"
+    exit $retVal
+  fi
+fi
+
+pushd submodules/cluster-node-tuning-operator
+echo "Building latency tools ... "
+make dist-latency-tests
+retVal=$?
+if [ $retVal -ne 0 ]; then
```
Tal-or, line 33: dito

cnf-tests/entrypoint/test-run.sh
```diff
@@ -3,19 +3,19 @@ set -e
...
+TEST_SUITES=${TEST_SUITES:-"latency-e2e.test"}
 SUITES_PATH="${SUITES_PATH:-~/usr/bin}"
 
-suites=( $TEST_SUITES )
+suites=($TEST_SUITES)
 if [ "$IMAGE_REGISTRY" != "" ] && [[ "$IMAGE_REGISTRY" != */ ]]; then
     export IMAGE_REGISTRY="$IMAGE_REGISTRY/"
 fi
 
 for suite in "${suites[@]}"; do
```
Tal-or, line 18: IIUC now we only have the `latency-e2e.test`, so why do we need this loop?

Tal-or (commented):
Thank you for the PR

Please add the PR description as part of the commit.
This would make it easier to track and understand the change

cnf-tests/entrypoint/test-run.sh
jlojosnegros: Because the intention is for the script to be able to run different test-suites if the user wants to.

cnf-tests/entrypoint/test-run.sh
jlojosnegros: Not trying to assign a default value but to get the assigned value or the default (true here) if variable is not defined.
In any case it was wrong.
Thanks for pointing the problem. Nice catch

cnf-tests/hack/build-test-bin.sh
jlojosnegros: you're right.
thanks for the catch.

jlojosnegros:
> @jlojosnegros do we need to also rewrite https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/run-functests.sh latency parts or this will be done in #1501 alone ?

The target for this PR it is just to provide the image for cnf-tests, so we do not need to rewrite anything else here.
AFAIK #1501 aims to rewrite all tests to use submodules instead of vendoring the test dependencies so, I assume we do not need to rewrite them  ... but maybe @liornoy could correct me if I said something wrong here

cnf-tests/entrypoint/test-run.sh
Tal-or: But this whole split purpose is to support a single suite case, so why not simplified as much as possible?

liornoy: @jlojosnegros sorry for the late reply, I was sick last week.
Tal-or's point is valid, i.e. with this split, the test-run.sh script will be used only for the latency test,
therefore we can simplify it and remove the loop, as it only needs to run the latency-e2e.test.

jlojosnegros: Do not worry at all @liornoy 
Change done. 
Thanks

cnf-tests/entrypoint/test-run.sh
```diff
@@ -3,19 +3,13 @@ set -e
...
 
-for suite in "${suites[@]}"; do
-    if [ "$DISCOVERY_MODE" == "true" ] &&  [ "$suite" == "configsuite" ]; then
-        echo "Discovery mode enabled, skipping setup"
-        continue
-    fi
-    echo running "$SUITES_PATH/$suite" "$@"
-    "$SUITES_PATH/$suite" "$@"
-done
+echo running "/usr/bin/latency-e2e.test" "$@"
```
Tal-or, line 14: the `$@` used to specify complete list of variables in the array, but we don't this array anymore so it can be removed as well.

cnf-tests/entrypoint/test-run.sh
```diff
@@ -3,19 +3,13 @@ set -e
...
+LATENCY_TEST_RUN="$LATENCY_TEST_RUN" DISCOVERY_MODE="$DISCOVERY_MODE" FEATURES="$FEATURES" "/usr/bin/latency-e2e.test" "$@"
```
Tal-or, line 15: ditto

cnf-tests/Dockerfile
```diff
@@ -81,15 +57,16 @@ COPY --from=builder-latency-test-runners /hwlatdetect-runner /usr/bin/hwlatdetec
...
-COPY --from=builder /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/bin/cnftests-sha.txt /usr/local/etc/cnf
+
+COPY --from=go-builder /app/cnf-tests/bin/mirror /usr/bin/mirror
+COPY --from=go-builder /app/cnf-tests/mirror/images.json /usr/local/etc/cnf
+COPY --from=go-builder /app/cnf-tests/entrypoint/test-run.sh /usr/bin/test-run.sh
+
 COPY tools/oot-driver/ /usr/src/oot-driver/
 
+# Copy test suite
+COPY --from=go-builder /app/cnf-tests/submodules/cluster-node-tuning-operator/build/_output/bin/latency-e2e.test /usr/bin/latency-e2e.test
```
shajmakh, line 68: why is this needed?

shajmakh: This is a binary load file of the latency tools tests. We use it in https://github.com/openshift/cluster-node-tuning-operator/tree/master/test/e2e/performanceprofile/functests/5_latency_testing which AFAIU don't need to export it in cnf-tests

shajmakh: Thanks for clarifying. the direction then looks good from my end.

PR #1513, opened 2023-05-23
s2i: bump dpdk s2i to centos stream 9 and dpdk version 22.11.2

SchSeba:


cgoncalves:
The base container image is being replaced by an RHEL 8.6 container image.

PR #1514, opened 2023-05-23
OCPBUGS-9413: ztp: remove the deprecated curation section from ClusterLogging CR

Missxiaoguo:
The curation section is now completely deprecated in cluster logging 5.7 which can be safely cleaned up. Removing it shouldn't have any impact on users who currently have it configured.


imiller0:
From your comment it sounds like it is also safe to leave this configuration in place on any existing clusters. Is that right? If so this lgtm

Missxiaoguo:
> From your comment it sounds like it is also safe to leave this configuration in place on any existing clusters. Is that right? If so this lgtm

Yes, exactly.

imiller0 (approved):
/lgtm

PR #1516, opened 2023-05-24
s2i: add capabilities on testpmd and testbbdev bins

mlguerrero12:
This adds the sys_resource, ipc_lock and net_raw (for mellanox) on the testpmd and testbbdev binaries for unpriviledged pods.

### DIFF
--- a/tools/s2i-dpdk/Dockerfile
+++ b/tools/s2i-dpdk/Dockerfile
@@ -67,6 +67,10 @@ COPY ./s2i/bin/ /usr/libexec/s2i
 
 RUN setcap cap_sys_resource,cap_ipc_lock=+ep /usr/libexec/s2i/run
 
+# Allows non-root users to use dpdk-testpmd.
+RUN setcap cap_sys_resource,cap_ipc_lock,cap_net_raw+ep /usr/local/bin/dpdk-testpmd
+RUN setcap cap_sys_resource,cap_ipc_lock,cap_net_raw+ep /usr/local/bin/dpdk-test-bbdev
+
 # This is needed for the s2i to work
 # in the pod yaml we still use the runAsUser:0 we w/a the ulimit issue
 USER 1001


tools/s2i-dpdk/Dockerfile
```diff
@@ -67,6 +67,9 @@ COPY ./s2i/bin/ /usr/libexec/s2i
 
 RUN setcap cap_sys_resource,cap_ipc_lock=+ep /usr/libexec/s2i/run
 
+# Allows non-root users to use dpdk-testpmd.
+RUN setcap cap_sys_resource,cap_ipc_lock,cap_net_raw+ep /usr/local/bin/dpdk-testpmd
```
SchSeba, line 71: should it have `=`

SchSeba: also can you add all the other dpdk-* commands?

mlguerrero12: it's not needed the =
will add it to the other commands

mlguerrero12: Had another thought about this. 

I think it's better to add the caps to the other commands when needed. I was blindly adding it for all of them but then I realized that a command like "dpdk-procinfo" doesn't need to allocate memory but needs to able to set CPU affinitiy. For that, it needs the CAP_SYS_NICE capability. So the requirements are different for some of them.

Note that the downstream image (gitlab) only has dpdk-testpmd and some python scripts. The other commands are not there.

SchSeba: so lets do it for the stuff that we use

```
dpdk-testpmd
dpdk-testbbdev
```

lets also add a comment so we don't forget about it if in the future we need to support other programs

mlguerrero12: where do we use dpdk-testbbdev?

mlguerrero12: @SchSeba, I can't find a test that uses it in this repo

SchSeba: we use it on d/s

mlguerrero12: what do you exactly mean with d/s?

mlguerrero12: (discussed internally with @SchSeba, testbbdev in used in cnf-gotests). Thanks @SchSeba, I added the caps.

SchSeba (approved):
/lgtm
/approve

mlguerrero12:
@SchSeba, job failures seem not to be related to these changes. Could you overwrite them please?

PR #1519, opened 2023-05-25
CNF-8619: ztp: diff acm's polcies with pgt's polcies

pixelsoccupied:
A simple diff tool meant to unblock https://github.com/openshift-kni/cnf-features-deploy/pull/1482 

It takes two sets of Policy CRs and diffs the object-templates. This is strictly a POC and meant to be used by the PR linked above. Do not makes any changes here directly (see the README.md) until we have product roadmap to develop/improve this for general purpose. 

/assign @serngawy 
/cc @imiller0 
/cc @lack


go.mod
```diff
@@ -35,13 +35,16 @@ require (
 	k8s.io/api v0.26.2
 	k8s.io/apiextensions-apiserver v0.26.2
 	k8s.io/apimachinery v0.26.2
-	k8s.io/client-go v1.5.2
+	k8s.io/client-go v12.0.0+incompatible
```
lack, line 38: This feels like a potentially dangerous upgrade?  What else uses client-go and what is incompatible about this version?

ztp/tools/policy-object-template-diff/Makefile
```diff
@@ -0,0 +1,15 @@
+GOCMD=go
+BINARY_NAME=policy-object-template-diff
+ACM_PATH?=""
+ZTP_PATH?=""
+
+build:
+	mkdir -p out/bin
+	GO111MODULE=on $(GOCMD) build -mod vendor -o out/bin/$(BINARY_NAME) .
```
lack, line 8: Do we need to use `-o` or can we rely on $GOBIN being set right?  I think $BINARY_NAME will default to the name of the directory that main.go is in, so it should already be called `policy-object-template-diff`?

ztp/tools/policy-object-template-diff/main.go
```diff
@@ -0,0 +1,447 @@
...
+
+func run() error {
+	setA := make(map[string][]CR)
+
+	setB := make(map[string][]CR)
+
+	/*
+		process set A
+	*/
+	populateMap(setA, getenv("A_PATH", "./example/acmpolicy"))
```
lack, line 54: Why `getenv` and not command line args?

go.mod
pixelsoccupied: I didnt manually change anything mod/vendor....It's auto updated `go mod tidy && go mod vendor && go mod verify` 

that said it's safe since this v0.26.1 is already pinned [here](https://github.com/openshift-kni/cnf-features-deploy/blob/0c55894ab9bedad180a5c347c81cee2252278026/go.mod#L150) (v1.5.2 didnt mean much)

ztp/tools/policy-object-template-diff/Makefile
pixelsoccupied: ` -o` just to make sure we know where the binary lands so that I can be explicit with the `run` and `clean` command below (not relying CI $GOPATH/bin or $GOBIN or $HOME/go/bin)

Made the name (policy-object-template-diff) explicit  at the top just to help with the `run` command.

ztp/tools/policy-object-template-diff/main.go
pixelsoccupied: No reason other than it was just faster to churn it out and lack of attention to UX :/ . If it's ok can we pin this (use CLI as input) as a feature of when this becomes something legit? Also would give me some time to playaround with Cobra and Viper

imiller0 (approved):

PR #1520, opened 2023-05-25
CNF-8413: ztp: add annotation in source-cr pp to assist with search

pixelsoccupied:
should be help with must-gather

/cc @imiller0


imiller0 (approved):
/lgtm

PR #1521, opened 2023-05-29
ztp: Update the reference configuration for LocalVolume

Missxiaoguo:
Update the StorageLV source-cr and the reference example to use persistent device names.


pixelsoccupied:
looking at related tickets...should we also update `rootDeviceHints` in SC to use persistent naming?

PR #1526, opened 2023-06-06
OCPBUGS-3513: use manifestsConfigMapRefs 

serngawy:


lack:
Is there a specific version of ACM required for this to work?  Should we update the docs to say that? Should we mention that in the commit message?

serngawy:
> Is there a specific version of ACM required for this to work? Should we update the docs to say that? Should we mention that in the commit message?

It is available starting from [v2.3](https://github.com/openshift/assisted-service/blob/v2.3.0/api/hiveextension/v1beta1/agentclusterinstall_types.go#L124) I guess they cherry-pick it to old releases. I think we are good, Do you want to mentioned that in the PR desc or doc ?

imiller0 (approved):
/lgtm

PR #1528, opened 2023-06-06
OCPBUGS-14511: ztp: disable OLM pprof collection on SNO

Missxiaoguo:
OLM collect-profiles Cron job is triggered every 15min to collect performance profiles, which causes huge CPU bursts. Add a source cr to disable it via PGT.

/cc @lack @imiller0


imiller0 (approved):
/lgtm

PR #1531, opened 2023-06-08
ztp: copy source-crs in current working directory to mimic git behavior in Makefile

sabbir-47:
- Makefile under policygen plugin earlier copied source-crs in the executable path
- Due to git first search, now the source-crs are copied under current working directory 
  - to mimic source-crs directory being in the git
  
 

/assign @serngawy @imiller0 @lack


natifridman:
`PolicyGenTemplate` is searching in `ran.openshift.io/v1/policygentemplate/source-crs/PerformanceProfile.yaml` path.
After this change I'm getting an error 
`Could not build the entire policy defined by PerformanceProfile.yaml is not found both in Git path:<folder>/source-crs/PerformanceProfile.yaml and in the ztp container path`

Did something else change in the plugin? or should I change something in my `PolicyGenTemplate`?

serngawy:
can you elaborate more on the setup you have ? Do you use ArgoCD ? OR do you run kustomize plugin locally ?

PR #1533, opened 2023-06-09
ztp: Expose disk cleaning option through SiteConfig

irinamihai:
- add the automatedCleaningMode field to the SiteConfig CRD and by default set it to "disabled" (path to the new field is SiteConfig.Spec.Cluster.Node.automatedCleaningMode)
- the automatedCleaningMode field from a SiteConfig CR will be translated into the spec.automatedCleaningMode field in the generated BareMetalHost CR(s)
- update testcases to reflect this new field

### DIFF
--- a/ztp/ran-crd/site-config-crd.yaml
+++ b/ztp/ran-crd/site-config-crd.yaml
@@ -318,6 +318,14 @@ spec:
                                 - UEFI
                                 - UEFISecureBoot
                                 - legacy
+                            automatedCleaningMode:
+                              description: When set to disabled, automated cleaning will be avoided
+                                during provisioning and deprovisioning.
+                              type: string
+                              default: disabled
+                              enum:
+                                - metadata
+                                - disabled
                             role:
                               description: The node role master|worker
                               type: string
--- a/ztp/siteconfig-generator/siteConfig/clusterCRs.go
+++ b/ztp/siteconfig-generator/siteConfig/clusterCRs.go
@@ -119,7 +119,7 @@ spec:
     disableCertificateVerification: true
     credentialsName: "{{ .Node.BmcCredentialsName.Name }}"
   bootMACAddress: "{{ .Node.BootMACAddress }}"
-  automatedCleaningMode: disabled
+  automatedCleaningMode: "{{ .Node.AutomatedCleaningMode }}"
   online: true
   rootDeviceHints: "{{ .Node.RootDeviceHints }}"
   userData:  "{{ .Node.UserData }}"
--- a/ztp/siteconfig-generator/siteConfig/siteConfig.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfig.go
@@ -398,6 +398,7 @@ func (prt *Partitions) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type Nodes struct {
 	BmcAddress             string                 `yaml:"bmcAddress"`
 	BootMACAddress         string                 `yaml:"bootMACAddress"`
+	AutomatedCleaningMode  string                 `yaml:"automatedCleaningMode"`
 	RootDeviceHints        map[string]interface{} `yaml:"rootDeviceHints"`
 	Cpuset                 string                 `yaml:"cpuset"`
 	NodeNetwork            NodeNetwork            `yaml:"nodeNetwork"`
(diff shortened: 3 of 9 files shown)


ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
```diff
@@ -14,7 +14,7 @@ spec:
   clusters:
   - clusterName: "site-sno-du-1"
     biosConfigRef:
-      filePath: "testSiteConfig/testHW.profile"
+      filePath: "../siteconfig-generator-kustomize-plugin/testSiteConfig/testHW.profile"
```
serngawy, line 17: why did you need to change that ?

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site2-standard-du.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   sshPrivateKeySecretRef:
     name: "sshPrvKey"
   biosConfigRef:
-    filePath: "testSiteConfig/testHW.profile"
+    filePath: "../siteconfig-generator-kustomize-plugin/testSiteConfig/testHW.profile"
```
serngawy, line 16: same

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
irinamihai: I saw there was an error while executing 
```
siteconfig-generator]$  ./siteconfig-generator -manifestPath ../source-crs/extra-manifest ../siteconfig-generator-kustomize-plugin/testSiteConfig/site2-standard-du.yaml
2023/06/12 13:56:33 Error: could not build the entire SiteConfig defined by ../siteconfig-generator-kustomize-plugin/testSiteConfig/site2-standard-du.yaml: open testSiteConfig/testHW.profile: no such file or directory
```
because the testHW.profile didn't exist at the specified filePath and thought to address this.


I can remove this change, since it's outside the scope of this PR.
@tliu2021, what do you think?

tliu2021: The file path mentioned here is relative to the kustomization.yaml file, and this test file is intended to be utilized by the Makefile within the siteconfig-generator-kustomize-plugin. If you wish to use this test file outside of the siteconfig-generator-kustomize-plugin,  it may be beneficial to use the proposed path.

irinamihai: My reasoning was that since the [siteconfig-generator README](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/siteconfig-generator) suggests using the SiteConfig files from siteconfig-generator-kustomize-plugin/testSiteConfig, then all of them should successfully run as  siteconfig-generator parameters.

ztp/ran-crd/site-config-crd.yaml
```diff
@@ -318,6 +318,14 @@ spec:
                                 - UEFI
                                 - UEFISecureBoot
                                 - legacy
+                            automatedCleaningMode:
+                              description: When set to disabled, automated cleaning will be avoided
+                                during provisioning and deprovisioning.
+                              type: string
+                              default: disabled
+                              enum:
+                                - metadata
```
serngawy, line 327: what metadata ? I think this should be enable

irinamihai: Hmm - I thought we are supposed to use the same definitions as in the original CRD, in this case BareMetaHost. This is how the BMH defines this field:
- https://github.com/metal3-io/baremetal-operator/blob/main/apis/metal3.io/v1alpha1/baremetalhost_types.go#L418
- https://github.com/metal3-io/baremetal-operator/blob/main/config/crd/bases/metal3.io_baremetalhosts.yaml#L77

Do we want to have the enum contain "enabled" and "disabled" instead of the original BMH definition? @imiller0 & @serngawy, kindly advise.

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site1-sno-du.yaml
tliu2021: I think your proposed change looks good.

irinamihai: Thanks, @tliu2021! We've decided to remove this change from this PR as it's unrelated. Might make a new PR with this small change in the future.

ztp/ran-crd/site-config-crd.yaml
irinamihai: Confirmed with baremetal folks that with the metadata option the partitioning table is removed, but the disk isn't fully wiped. I was wrong above and this is NOT related to this [field](https://github.com/metal3-io/baremetal-operator/blob/main/config/crd/bases/metal3.io_baremetalhosts.yaml#L245).

I updated the testcases accordingly. Thanks, @serngawy

serngawy (approved):

PR #1541, opened 2023-06-21
ztp: enable kubens service from tuned and remove legacy CRs doing the doing same thing

pixelsoccupied:
See CNF-5332 for more details. 

depends on: https://github.com/containers/kubensmnt/pull/11 

/cc @lack


ztp/source-crs/extra-manifest/01-kubelet-conf-master.yaml
```diff
@@ -0,0 +1,41 @@
...
+          mode: 493
+          path: /usr/local/bin/extractExecStart
+        - contents:
+            source: data:text/plain;charset=utf-8;base64,IyEvYmluL2Jhc2gKbnNlbnRlciAtLW1vdW50PS9ydW4vY29udGFpbmVyLW1vdW50LW5hbWVzcGFjZS9tbnQgIiRAIgo=
+          mode: 493
+          path: /usr/local/bin/nsenterCmns
```
lack, line 22: This nsenterCmns is no longer needed!

ztp/extra-manifests-builder/01-kubelet-conf/nsenterCmns
lack, line 1: This file should be removed

ztp/extra-manifests-builder/01-kubelet-conf/90-container-mount-namespace-kubelet.conf
lack, line 1: This file should be renamed, since there's no container-mount-namespace in it any more

ztp/extra-manifests-builder/01-kubelet-conf/build.sh
```diff
@@ -13,10 +13,8 @@ MCPROLE=${MCPROLE:-master}
 #  Housekeeping : 10s
 #  Eviction : 10s
 
-${MCMAKER} -name container-mount-namespace-and-kubelet-conf -mcp ${MCPROLE} -stdout \
+${MCMAKER} -name 01-kubelet-conf -mcp ${MCPROLE} -stdout \
         file -source extractExecStart -path /usr/local/bin/extractExecStart -mode 0755 \
         file -source nsenterCmns -path /usr/local/bin/nsenterCmns -mode 0755 \
```
lack, line 18: Not needed

ztp/extra-manifests-builder/01-kubelet-conf/build.sh
```diff
@@ -13,10 +13,8 @@ MCPROLE=${MCPROLE:-master}
...
-        unit -source container-mount-namespace.service -enable=false \
-        dropin -source 90-container-mount-namespace.conf -for crio.service \
-        dropin -source 90-container-mount-namespace-kubelet.conf -name 90-container-mount-namespace.conf -for kubelet.service \
+        dropin -source 90-container-mount-namespace-kubelet.conf -for kubelet.service \
```
lack, line 19: Should be renamed

ztp/extra-manifests-builder/01-kubelet-conf/90-kubelet-housekeeping-interval.conf
```diff
@@ -0,0 +1,4 @@
+[Service]
+ExecStartPre=/usr/local/bin/extractExecStart %n /%t/%N-execstart.env ORIG_EXECSTART
+EnvironmentFile=-/%t/%N-execstart.env
+ExecStart=bash -c "${ORIG_EXECSTART} --housekeeping-interval=30s"
```
lack, line 4: Because we're editing ExecStart and ExecStart can only have one value, we are forced to manually clear it like this, or Systemd will complain and the service won't start:
```suggestion
ExecStart=
ExecStart=bash -c "${ORIG_EXECSTART} --housekeeping-interval=30s"
```

pixelsoccupied:
combine 30 + 90

pixelsoccupied:
move removed stuff in [optional-extra-manifest](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/source-crs/optional-extra-manifest)

ztp/source-crs/TunedPerformancePatch.yaml
```diff
@@ -25,6 +25,7 @@ spec:
         [service]
         service.stalld=start,enable
         service.chronyd=stop,disable
+        service.kubens=start,enable
```
imiller0, line 28: In the prior implementation the service gets dropped in by MC during installation and would be present from then on. With this implementation the service doesn't get enabled until this tuned patch is applied at day-2 (much later in the process). Perhaps that is OK since this will still be before a workload is applied (and the mount namespace hiding is helping reduce CPU with large numbers of pods) but there is a period of time where the system will be running without it. 
Is there a way to enable it at installation time?

pixelsoccupied: Yep it's possible with a MC

lack: I think we had originally talked about enabling it here because it might have been a better way to have the 2 different implementations coexist - But since we're ripping out the old one entirely, maybe going with the MC-based enable would be better than the NTO for the reason @imiller0 points out.

PR #1542, opened 2023-06-21
ztp: rename DefaultCatsrc to inform that CR is for disconnected env and update all subscriptions

pixelsoccupied:
CNF-8305. Using DefaultCatsrc CR means we are using disconnected env.

/cc @imiller0 
/cc @sabbir-47


imiller0 (commented):
This should drive a corresponding change in the subscriptions which make use of this catalogsource (they need to match the name).

imiller0 (approved):
/lgtm
If users configure a name which overlaps any of the default OLM catalog sources (eg redhat-operators like it was) they can run into an intermittent issue where the catalog source gets overwritten. This ensures that the reference is aligned with the recommendation. Thanks!

PR #1546, opened 2023-06-23
CNF-7818: ztp: Uncomment workloadHints object in PerformanceProfile

sakhoury:
The is PR re-introduces the `workloadHints` object in the PerformanceProfile source-crs. It was initially added in 4.13, however it had to be commented out because at that point QE was experiencing issues deploying tests as ZTP was version unaware.

Now, with ZTP version awareness - it can be safely added back.

/cc @imiller0 @bartwensley  @serngawy @sabbir-47


sakhoury:
ping for a review
/cc @imiller0

imiller0 (commented):
With these in the source-cr it is redundant to have them in the example PolicyGenTemplates as well. Can you remove those entries at the same time?

imiller0 (approved):
/lgtm

PR #1557, opened 2023-07-12
SRIOV ExcludeTopology tests

zeeke:
This PR contains a version bump for the sriov-network-operator to use the new API field `SriovNetworkNodePolicy.Spec.ExcludeTopology`.

Tests leverage two devices on different numa nodes and guaranteed pods.

cc @SchSeba , @gregkopels

### DIFF
--- a/cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
+++ b/cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
@@ -1,0 +1,163 @@
+package dpdk
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	sriovv1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	sriovcluster "github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/cluster"
+	sriovnamespaces "github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/namespaces"
+	sriovnetwork "github.com/k8snetworkplumbingwg/sriov-network-operator/test/util/network"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/client"
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/discovery"
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/namespaces"
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/networks"
+	utilNodes "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/nodes"
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/performanceprofile"
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/pods"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+)
+
+var _ = Describe("[sriov] NUMA node alignment", Ordered, func() {
+
+	BeforeAll(func() {
+		if discovery.Enabled() {
+			Skip("Discovery mode not supported")
+		}
+
+		isSNO, err := utilNodes.IsSingleNodeCluster()
+		Expect(err).ToNot(HaveOccurred())
+		if isSNO {
+			Skip("Single Node openshift not yet supported")
+		}
+
+		perfProfile, err := performanceprofile.FindDefaultPerformanceProfile(performanceProfileName)
+		Expect(err).ToNot(HaveOccurred())
+		if !performanceprofile.IsSingleNUMANode(perfProfile) {
+			Skip("SR-IOV NUMA test suite expects a performance profile with 'single-numa-node' to be present")
+		}
+
+		err = namespaces.Create(sriovnamespaces.Test, client.Client)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Clean SRIOV policies and networks")
+		networks.CleanSriov(sriovclient)
+
+		By("Discover SRIOV devices")
+		sriovCapableNodes, err := sriovcluster.DiscoverSriov(sriovclient, namespaces.SRIOVOperator)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(sriovCapableNodes.Nodes)).To(BeNumerically(">", 0))
+		testingNode, err := nodes.GetByName(sriovCapableNodes.Nodes[0])
+		Expect(err).ToNot(HaveOccurred())
+		By("Using node " + testingNode.Name)
+
+		sriovDevices, err := sriovCapableNodes.FindSriovDevices(testingNode.Name)
+		Expect(err).ToNot(HaveOccurred())
+
+		numa0Device, err := findDeviceOnNUMANode(testingNode, sriovDevices, "0")
+		Expect(err).ToNot(HaveOccurred())
+		By("Using NUMA0 device " + numa0Device.Name)
+
+		numa1Device, err := findDeviceOnNUMANode(testingNode, sriovDevices, "1")
+		Expect(err).ToNot(HaveOccurred())
+		By("Using NUMA1 device " + numa1Device.Name)
+
+		// SriovNetworkNodePolicy
+		// NUMA node0 device excludeTopology = true
+		// NUMA node0 device excludeTopology = false
+		// NUMA node1 device excludeTopology = true
+		// NUMA node1 device excludeTopology = false
+
+		By("Create SRIOV policies and networks")
+
+		ipam := `{ "type": "host-local", "subnet": "192.0.2.0/24" }`
+
+		createSriovNetworkAndPolicyForNumaAffinityTest(8, numa0Device, "#0-3",
+			"test-numa-0-exclude-topology-false-", testingNode.Name,
+			"testNuma0ExcludeTopoplogyFalse", ipam, false)
+
+		createSriovNetworkAndPolicyForNumaAffinityTest(8, numa0Device, "#4-7",
+			"test-numa-0-exclude-topology-true-", testingNode.Name,
+			"testNuma0ExcludeTopoplogyTrue", ipam, true)
+
+		createSriovNetworkAndPolicyForNumaAffinityTest(8, numa1Device, "#0-3",
+			"test-numa-1-exclude-topology-true-", testingNode.Name,
+			"testNuma1ExcludeTopoplogyFalse", ipam, false)
+
+		createSriovNetworkAndPolicyForNumaAffinityTest(8, numa1Device, "#4-7",
+			"test-numa-1-exclude-topology-true-", testingNode.Name,
+			"testNuma1ExcludeTopoplogyTrue", ipam, true)
+
+		By("Waiting for SRIOV devices to get configured")
+		networks.WaitStable(sriovclient)
+	})
+
+	BeforeEach(func() {
+		By("Clean any pods in " + sriovnamespaces.Test + " namespace")
+		namespaces.CleanPods(sriovnamespaces.Test, sriovclient)
+	})
+
+	It("Validate the creation of a pod with excludeTopology set to False and an SRIOV interface in a different NUMA node than the pod", func() {
+		pod := pods.DefinePod(sriovnamespaces.Test)
+		pods.RedefineWithGuaranteedQoS(pod, "1", "100m")
+		pod = pods.RedefinePodWithNetwork(pod, "test-numa-0-exclude-topology-false")
+
+		pod, err := client.Client.Pods(sriovnamespaces.Test).
+			Create(context.Background(), pod, metav1.CreateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func(g Gomega) {
+			actualPod, err := client.Client.Pods(sriovnamespaces.Test).Get(context.Background(), pod.Name, metav1.GetOptions{})
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(actualPod.Status.Phase).To(Equal(corev1.PodFailed))
+			g.Expect(actualPod.Status.Reason).To(Equal("TopologyAffinityError"))
+		}).Should(Succeed())
+	})
+})
+
+func findDeviceOnNUMANode(node *corev1.Node, devices []*sriovv1.InterfaceExt, numaNode string) (*sriovv1.InterfaceExt, error) {
+	for _, device := range devices {
+		out, err := nodes.ExecCommandOnNode([]string{
+			"cat",
+			filepath.Clean(filepath.Join("/sys/class/net/", device.Name, "/device/numa_node")),
+		}, node)
+		if err != nil {
+			klog.Warningf("can't get device [%s] NUMA node: out(%s) err(%s)", device.Name, string(out), err.Error())
+			continue
+		}
+
+		if out == numaNode {
+			return device, nil
+		}
+	}
+
+	return nil, fmt.Errorf("can't find any SR-IOV device on NUMA [%s] for node [%s]. Available devices: %+v", numaNode, node.Name, devices)
+}
+
+func withExcludeTopology(excludeTopology bool) func(*sriovv1.SriovNetworkNodePolicy) {
+	return func(p *sriovv1.SriovNetworkNodePolicy) {
+		p.Spec.ExcludeTopology = excludeTopology
+	}
+}
+
+func createSriovNetworkAndPolicyForNumaAffinityTest(numVFs int, intf *sriovv1.InterfaceExt, vfSelector, policyGeneratedName, nodeName, resourceName, ipam string, excludeTopology bool) {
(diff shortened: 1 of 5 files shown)


cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,226 @@
...
+		Eventually(func(g Gomega) {
+			actualPod, err := client.Client.Pods(sriovnamespaces.Test).Get(context.Background(), pod.Name, metav1.GetOptions{})
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(actualPod.Status.Phase).To(Equal(corev1.PodFailed))
+			g.Expect(actualPod.Status.Message).To(ContainSubstring("Resources cannot be allocated with Topology locality"))
+		}).Should(Succeed())
+	})
+})
+
+func findDeviceOnNUMANode(node *corev1.Node, devices []*sriovv1.InterfaceExt, numaNode string) (*sriovv1.InterfaceExt, error) {
```
fedepaol, line 143: These (and below) are worth being in some pkg

cnf-tests/testsuites/pkg/performanceprofile/performanceprofile.go
```diff
@@ -78,6 +79,50 @@ func FindOrOverridePerformanceProfile(performanceProfileName, machineConfigPoolN
...
+	}
+
+	mcp := &mcv1.MachineConfigPool{}
+	err = client.Client.Get(context.TODO(), goclient.ObjectKey{Name: machineConfigPoolName}, mcp)
+	if err != nil {
+		return err
+	}
+
+	if previousPerfProfile != nil {
+		OriginalPerformanceProfile = previousPerfProfile.DeepCopy()
```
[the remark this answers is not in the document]
zeeke: Good point. What would you use for the map key? previous performance profile name?

I guess this is not going to work in any case when tests run in parallel, as the affected nodes could collide and lead to an unexpected cluster state.

Maybe we can get rid of this save/restore mechanism and make tests clear and create PerformanceProfile. WDYT?

cnf-tests/testsuites/pkg/performanceprofile/performanceprofile.go
```diff
@@ -154,7 +199,30 @@ func DiscoverPerformanceProfiles(enforcedPerformanceProfileName string) (bool, s
 	return false, fmt.Sprintf("Can not run tests in discovery mode. Failed to find a valid perfomance profile. %s", err), nil
 }
 
-func CreatePerformanceProfile(performanceProfileName, machineConfigPoolName string) error {
+func DefineSingleNUMANodePerfomanceProfile(performanceProfileName, machineConfigPoolName string) *performancev2.PerformanceProfile {
```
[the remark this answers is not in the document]
zeeke: good point. done

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,226 @@
...
+		)
+
+		By("Waiting for SRIOV devices to get configured")
+		networks.WaitStable(sriovclient)
+	})
+
+	AfterAll(func() {
+		By("Cleaning performance profiles")
+
+		err := performanceprofile.CleanPerformanceProfiles()
```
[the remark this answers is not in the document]
zeeke: Right, I had to fix the `RestorePerformanceProfile()` to handle the case `OriginalPerformanceProfile == nil`

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,226 @@
...
+	BeforeEach(func() {
+		if discovery.Enabled() {
+			Skip("Discovery mode not supported")
+		}
+
+		isSNO, err := utilNodes.IsSingleNodeCluster()
+		Expect(err).ToNot(HaveOccurred())
+		if isSNO {
+			Skip("Single Node openshift not yet supported")
+		}
```
cgoncalves, line 39: None of the below tests make changes to these conditions. Should these be moved into BeforeAll?

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,226 @@
...
+			withNumVFs(8), withPfNameSelector(numa1Device.Name+"#4-7"),
+			withNetworkNameAndNamespace(sriovnamespaces.Test, "test-numa-1-exclude-topology-true"),
+			withExcludeTopology(true),
+		)
+
+		By("Waiting for SRIOV devices to get configured")
+		networks.WaitStable(sriovclient)
+	})
+
+	AfterAll(func() {
```
cgoncalves, line 117: In BeforeAll, a namespace is created. Should it not be deleted here?

If CI runs test suites in parallel, the same namespace may already exist in which case either
  a) we don't want to delete it here at the cost of leaving it as a leftover or
  b) we create its own namespace and so we can freely delete it here

cnf-tests/testsuites/pkg/performanceprofile/performanceprofile.go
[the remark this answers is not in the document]
fedepaol: The problem is, IIRC in some scenarios the performance profile is created upfront to save some running time.

And my concern is not about running tests in parallel, but this not being able to restore the right one. Until now there was only a test that was doing that in an isolated manner, so the global was serving that purpouse only. Now we are using it for two different scenarios, and looking at the API that is not clear.

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
zeeke: You are right! This comes from a bad habit we developed while working with the custom version of [`BeforeAll()`](https://github.com/openshift-kni/cnf-features-deploy/blob/dad3bccf61b7b33ed1319a69f8810745157af7a4/cnf-tests/testsuites/pkg/execute/ginkgo.go#L7), where skipping in the passed function would skip only the first test in the node.

BTW Fixed

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
zeeke: Good point. The problem here is that `"sriov-conformance-testing"` namespace is a little bit special:

- It must exist to make the DiscoverSriov() function work. (es [s2i.go#L163](https://github.com/openshift-kni/cnf-features-deploy/blob/dad3bccf61b7b33ed1319a69f8810745157af7a4/cnf-tests/testsuites/e2esuite/s2i/s2i.go#L163) ). We need to clean that in the upstream repository.

- It's dumped by the k8sreporter (see [pkg/utils/reporter.go#L97](https://github.com/openshift-kni/cnf-features-deploy/blob/efe8c5ddcad32b5d0b8c58f51d391aafc96fe77d/cnf-tests/testsuites/pkg/utils/reporter.go#L97)), though it wouldn't be a problem to an entry like "numa-dpdk-tests-ns": "sriov"

- it's cleaned by the Fixture mechanism (see [pkg/features/features.go#L103](https://github.com/openshift-kni/cnf-features-deploy/blob/efe8c5ddcad32b5d0b8c58f51d391aafc96fe77d/cnf-tests/testsuites/pkg/features/features.go#L103)), as cleaning it in the AfterAll/AfterEach makes the k8sreporter not to gather it.

None of the above points represents a hard limit for using the own namespace, but they require a reasonable effort.

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,222 @@
...
+
+		pod, err := client.Client.Pods(sriovnamespaces.Test).
+			Create(context.Background(), pod, metav1.CreateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func(g Gomega) {
+			actualPod, err := client.Client.Pods(sriovnamespaces.Test).Get(context.Background(), pod.Name, metav1.GetOptions{})
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(actualPod.Status.Phase).To(Equal(corev1.PodFailed))
+			g.Expect(actualPod.Status.Message).To(ContainSubstring("Resources cannot be allocated with Topology locality"))
```
ffromani, line 135: nonblocking: I think checking `Status.Reason` (rather than the message) is a bit simpler and cleaner. I reckon the way we report errors in kube is not great in general.

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,222 @@
...
+		}).Should(Succeed())
+	})
+})
+
+func findDeviceOnNUMANode(node *corev1.Node, devices []*sriovv1.InterfaceExt, numaNode string) (*sriovv1.InterfaceExt, error) {
+	for _, device := range devices {
+		out, err := nodes.ExecCommandOnNode([]string{"cat", fmt.Sprintf("/sys/class/net/%s/device/numa_node", device.Name)}, node)
```
ffromani, line 142: nonblocking: being biten enough time already to suggest using `filepath.Clean(filepath.Join(....))`.

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
zeeke: The below functions rely on test-specific default values. Moving them to a generic package would make them much less readable. WDYT?

zeeke:
I put the creation of the performance profile out of this PR's scope, so the suite expects the cluster to be configured with a single-numa-node profile.

Creating it dynamic can break the cluster, and I prefer to tackle that problem in a subsequent PR.

@fedepaol , @cgoncalves , @ffromani , @gregkopels PTAL

cnf-tests/testsuites/pkg/performanceprofile/performanceprofile.go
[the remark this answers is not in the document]
fedepaol: An alternative here is to check if `OriginalPerformanceProfile` is filled already and panic in that case, because that would mean we are re-replacing

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
fedepaol: The only one I see test specific is createSriovNetworkAndPolicy.
Also, we already have functions to create sriov networks and sriov policies. I'd try to extend / modify those instead of having another variant scattered across the repo.

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func withNetworkNameAndNamespace(namespace, name string) func(*sriovv1.SriovNetworkNodePolicy, *sriovv1.SriovNetwork) {
+	return func(p *sriovv1.SriovNetworkNodePolicy, n *sriovv1.SriovNetwork) {
+		n.ObjectMeta.Name = name
+		n.Spec.NetworkNamespace = namespace
+	}
+}
+
+func createSriovNetworkAndPolicy(opts ...func(*sriovv1.SriovNetworkNodePolicy, *sriovv1.SriovNetwork)) {
```
fedepaol, line 191: What I don't like of this approach is, the modifier takes both the policy and the network.
We can split this in two (in the calling site) and have a more focused modifier for each side.

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,221 @@
...
+		Eventually(func(g Gomega) {
+			actualPod, err := client.Client.Pods(sriovnamespaces.Test).Get(context.Background(), pod.Name, metav1.GetOptions{})
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(actualPod.Status.Phase).To(Equal(corev1.PodFailed))
+			g.Expect(actualPod.Status.Reason).To(Equal("TopologyAffinityError"))
+		}).Should(Succeed())
+	})
+})
+
+func findDeviceOnNUMANode(node *corev1.Node, devices []*sriovv1.InterfaceExt, numaNode string) (*sriovv1.InterfaceExt, error) {
```
ffromani, line 136: this works. The possible issue I see is this causes quite a lot of `ExecCommandOnNode` which is not too cheap. Could it be worth to discover all the relevant devices (or at least the PFs - I can't imagine a device which does or a setup which wants a VF on a different NUMA node than the one the PF is attached to) and report their NUMA affinity, than make logic on the test side?

cnf-tests/testsuites/pkg/performanceprofile/performanceprofile.go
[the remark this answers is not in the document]
ffromani: would it make sense to return the previous perf profile and set the global/map/whatever in the calling site? this will at least make the flow more explicit instead of buried inside this function, which is not great when we mutate the global state.

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,221 @@
...
+		// NUMA node0 device excludeTopology = true
+		// NUMA node0 device excludeTopology = false
+		// NUMA node1 device excludeTopology = true
+		// NUMA node1 device excludeTopology = false
+
+		By("Create SRIOV policies and networks")
+
+		createSriovNetworkAndPolicy(
+			withNodeSelector(testingNode),
+			withNumVFs(8), withPfNameSelector(numa0Device.Name+"#0-3"),
```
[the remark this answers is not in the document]
zeeke: I had a quick look at supported hardware [1] datasheets, and we can safely assume a NIC supports at least 128 VFs [2] [3] [4] [5] [6] [7] [8], though I didn't find how many VFs Pensando NICs actually supports [9].


[1] https://docs.openshift.com/container-platform/4.13/networking/hardware_networks/about-sriov.html
[2] https://docs.broadcom.com/doc/957414A4142CC-DS
[3] https://docs.broadcom.com/doc/957508-P2100G-DS
[4] https://cdrdv2-public.intel.com/332464/332464_710_Series_Datasheet_v_4_1.pdf
[5] file:///home/apanatto/Downloads/332464_710_Series_Datasheet_v_4_1.pdf
[6] https://www.cisco.com/c/dam/en/us/products/collateral/servers-unified-computing/ucs-c-series-rack-servers/intel-e810-cqda2-ethernet-network-adapter-product-brief.pdf
[7] https://network.nvidia.com/files/doc-2020/pb-connectx-4-lx-en-card.pdf
[8] https://network.nvidia.com/files/doc-2020/pb-connectx-5-en-card.pdf
[9]

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
zeeke: What do you mean by "report their NUMA affinity"? Wouldn't it involve a call to ExecCommandOnNode for each PF?

We can improve this by indexing every device by its NUMA node, but it wouldn't make that difference: suppose we have 4 devices splitter in NUMA nodes 0 and 1

|device|NUMA node|
|---|---|
| ens1f0|0|
| ens1f1|0|
| ens3f0|1|
| ens3f1|1|


With current implementation:
- `findDeviceOnNUMANode(..., 0)` calls 1 `ExecCommandOnNode` and return ens1f0
- `findDeviceOnNUMANode(..., 1)` calls 3 `ExecCommandOnNode` and return ens3f0

For a total of 4 calls, the same as we would have by looping all the devices for indexing

Furthermore, these tests spend a lot of time waiting minutes for SR-IOV devices to get configured. I feel like we are trying to optimize a very small piece of the whole puzzle.

zeeke:
`ci/prow/e2e-gcp-ovn ` test case

```
[rfe_id:27368][performance] Network latency parameters adjusted by the Node Tuning Operator [test_id:28467][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should contain configuration injected through the openshift-node-performance profile
```
is failing in other PRs, hence I assume it's not related to these changes.

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,163 @@
...
+	})
+
+	BeforeEach(func() {
+		By("Clean any pods in " + sriovnamespaces.Test + " namespace")
+		namespaces.CleanPods(sriovnamespaces.Test, sriovclient)
+	})
+
+	It("Validate the creation of a pod with excludeTopology set to False and an SRIOV interface in a different NUMA node than the pod", func() {
+		pod := pods.DefinePod(sriovnamespaces.Test)
+		pods.RedefineWithGuaranteedQoS(pod, "1", "100m")
```
fedepaol, line 109: nit: can you use the same pod = ... pattern here? Or, is there a reason why it was not done?

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,163 @@
...
+		pod = pods.RedefinePodWithNetwork(pod, "test-numa-0-exclude-topology-false")
+
+		pod, err := client.Client.Pods(sriovnamespaces.Test).
```
fedepaol, line 112: can we differentiate here between pod - the template for creation and pod - the created pod?

fedepaol:
left a few *non blocking* nits, to change only if you need to touch the codebase anyway.

/lgtm

ffromani:
/lgtm

there area still some areas on which we can improve/generalize, but not blocking for this work.

PR #1559, opened 2023-07-13
ztp: update the sriov-fec to use vfio-pci for PR also

SchSeba:
Be aware after this change when starting a dpdk application ther is a need to also use the vfio-token that is injected into the pod


imiller0 (approved):
/lgtm

PR #1560, opened 2023-07-13
Bug OCPBUGS-14921: ztp: Use HTTP as default transport for cloud events 

jzding:
Replace PVC with ConfigMap to persist subscriber data for HTTP transport


ztp/source-crs/StoragePV.yaml
```diff
@@ -1,29 +0,0 @@
-apiVersion: v1
```
imiller0, line 1: This CR is part of the 4.13 release and could be used by customers for other PVs as well. I don't think it should be removed.

imiller0 (commented):
/lgtm

PR #1561, opened 2023-07-13
OCPBUGS-13634: ztp: remove accelerated container startup

tliu2021:
A known issue, tracked by OCPBUGS-15102 in OCP 4.13 and 4.14, could lead to the CPU affinities not resetting to unrestricted during startup. Consequently, the accelerated startup script fails to deliver significant improvements in boot time.

Furthermore, it was observed that the accelerated startup script did not yield substantial improvements in boot time in OCP 4.12. Therefore, it might be necessary to explore alternative approaches instead of relying solely on setting CPU affinities to unrestricted cores.

This update addresses the issue by removing the MC that contains the accelerated startup script and services. Moving forward, once the OCPBUGS-15102 is resolved, a unified accelerated startup method will be developed to cater to all cluster topologies.


imiller0 (approved):
/lgtm

imiller0 (commented):
/lgtm

PR #1568, opened 2023-07-17
CNF-8851: ztp: Create warning for deprecated field: extraManifestPath

sabbir-47:
- deprecation warning added for extraManifestPath in Siteconfig

/assign @imiller0 @serngawy @eggfoobar

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
@@ -32,6 +32,44 @@ type annotationValue struct {
 	fieldMessage string
 }
 
+type AnnotationMessage struct {
+	CRName string
+	annotationValue
+	ShouldBeApplied func(Clusters) bool
+}
+
+var annotationMessages = []AnnotationMessage{
+	{
+		CRName: "AgentClusterInstall",
+		annotationValue: annotationValue{
+			fieldName:    "cpuset",
+			fieldMessage: "cpuset will be deprecated after OCP 4.15, please use cpuPartitioningMode for OCP versions >= 4.13",
+		},
+		ShouldBeApplied: func(c Clusters) bool {
+			for _, node := range c.Nodes {
+				if node.Cpuset != "" {
+					return true
+				}
+			}
+			return false
+		},
+	},
+
+	{
+		CRName: "ConfigMap",
+		annotationValue: annotationValue{
+			fieldName:    "extraManifestPath",
+			fieldMessage: "extraManifestPath will be deprecated after OCP 4.15, please use ExtraManifests.SearchPaths for OCP versions >= 4.14",
+		},
+		ShouldBeApplied: func(c Clusters) bool {
+			if len(c.ExtraManifestPath) > 0 {
+				return true
+			}
+			return false
+		},
+	},
+}
+
 func NewAnnotationWarning(annoKey string) *annotationWarning {
 	return &annotationWarning{
 		annoKey: fmt.Sprintf("%s-%s", ZtpWarningAnnotation, annoKey),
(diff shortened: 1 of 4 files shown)


ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
```diff
@@ -366,25 +366,44 @@ func applyWorkloadPinningInstallConfigOverrides(clusterSpec *Clusters) (result s
...
-	)
 
-	for _, node := range clusterSpec.Nodes {
-		// If a CPUSet exists on any of the nodes, the whole cluster, regardless if SNO or not, must be setup
-		// for partitioning.
-		if node.Cpuset != "" {
-			// If Node has CPUSet add deprecation warning
-			deprecationWarnings.Add(deprecationCR, deprecationField, deprecationMessage)
-			break
+	for _, field := range deprecationField {
```
eggfoobar, line 374: I like this idea a lot, the only thing I'm worried about with this is scalability, if we need to add more. What do you think about changing this up a bit and moving the trigger logic away from here and let each deprecation annotation decide when it should be applied. Something like this?

```go
type AnnotationMessage struct {
	CRName          string
	Field           string
	Message         string
	ShouldBeApplied func(Clusters) bool
}
...
var (
	cpuSetAnnotationDeprecation = AnnotationMessage{
		CRName:  "AgentClusterInstall",
		Field:   "cpuset",
		Message: "cpuset will be deprecated after OCP 4.15, please use cpuPartitioningMode for OCP versions >= 4.13",
		ShouldBeApplied: func(c Clusters) bool {
			for _, node := range c.Nodes {
				if node.Cpuset != "" {
					return true
				}
			}
			return false
		},
	}

	extraManifestPathAnnotationDeprecation = AnnotationMessage{
		CRName:  "ConfigMap",
		Field:   "extraManifestPath",
		Message: "extraManifestPath will be deprecated after OCP 4.15, please use ExtraManifests.SearchPaths for OCP versions >= 4.14",
		ShouldBeApplied: func(c Clusters) bool {
			if len(c.ExtraManifestPath) > 0 {
				return true
			}
			return false
		},
	}
)
...
// getDeprecationWarnings is a helper function to add deprecation warnings.
func getDeprecationWarnings(clusterSpec Clusters, deprecationField ...AnnotationMessage) *annotationWarning {
	deprecationWarnings := NewAnnotationWarning(ZtpDeprecationWarningAnnotationPostfix)

	for _, field := range deprecationField {
		if field.ShouldBeApplied != nil && field.ShouldBeApplied(clusterSpec) {
			deprecationWarnings.Add(field.CRName, field.Field, field.Message)
		}
	}
	return deprecationWarnings
}
````

**Review by eggfoobar — COMMENTED** [review]
Just had a thought about scalability, if you don't want to change it up in this PR, we can do it down the road, but we should revisit this approach if we need to add another warning message.

File: ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go [hunk 2]
```diff
@@ -206,7 +206,7 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
 		return clusterCRs, err
 	}
 
-	deprecationWarnings := getDeprecationWarnings(cluster)
+	deprecationWarnings := getDeprecationWarnings(cluster, deprecateCPUset, deprecateExtraManifestPath)
```
serngawy, line 209: deprecateCPUset and deprecateExtraManifestPath are constants no need to pass them as variable, just assign them.

ztp/siteconfig-generator/siteConfig/testdata/filteredoutput/onlyUserCR.yaml
```diff
@@ -116,6 +116,7 @@ metadata:
     annotations:
         argocd.argoproj.io/sync-wave: "1"
         ran.openshift.io/ztp-gitops-generated: '{}'
+        ran.openshift.io/ztp-warning-field-deprecation-extraManifestPath: extraManifestPath will be deprecated after OCP 4.15, please use ExtraManifests.SearchPaths for OCP versions >= 4.14
```
serngawy, line 119: Shorter annotation key is better something like "ran.openshift.io/field-deprecation"

serngawy (changes requested):

ztp/siteconfig-generator/siteConfig/testdata/filteredoutput/onlyUserCR.yaml
sabbir-47: for the cpuset deprecation the annotation is formed like below and i followed to be consistent:

`ran.openshift.io/ztp-warning-field-deprecation-cpuset`

ztp/siteconfig-generator/siteConfig/siteConfigHelper.go
sabbir-47: @eggfoobar I really liked your suggestion, so i changed it accordingly with this commit df8d23ecfd531c39b05ef4c003088cf08c38a31e

serngawy:
Sounds good to me
/LGTM

serngawy (approved):

PR #1573, opened 2023-07-20
Numa test cases with multi interfaces excludeTopology true

gkopels:


cnf-tests/testsuites/pkg/performanceprofile/performanceprofile.go
```diff
@@ -78,6 +79,50 @@ func FindOrOverridePerformanceProfile(performanceProfileName, machineConfigPoolN
...
+	err = client.Client.Get(context.TODO(), goclient.ObjectKey{Name: machineConfigPoolName}, mcp)
+	if err != nil {
+		return err
+	}
+
+	if previousPerfProfile != nil {
+		OriginalPerformanceProfile = previousPerfProfile.DeepCopy()
+
+		// Clean and create a new performance profile for the dpdk application
+		err = CleanPerformanceProfiles()
```
SchSeba, line 102: I think we should not clean wait and then apply and wait again. this means for a multi node cluster that we will wait much more time then needed.

we can just remove and apply in the same time. it will reboot the first node to remove the profile but the other nodes will already apply the new policy on the first reboot.

also another option is to just apply the policy to a new machine config pool and move the nodes to that pull

cnf-tests/testsuites/pkg/performanceprofile/performanceprofile.go
```diff
@@ -154,7 +199,30 @@ func DiscoverPerformanceProfiles(enforcedPerformanceProfileName string) (bool, s
 	return false, fmt.Sprintf("Can not run tests in discovery mode. Failed to find a valid perfomance profile. %s", err), nil
 }
 
-func CreatePerformanceProfile(performanceProfileName, machineConfigPoolName string) error {
+func DefineSingleNUMANode(performanceProfileName, machineConfigPoolName string) *performancev2.PerformanceProfile {
+	isolatedCPUSet := performancev2.CPUSet("2-3")
```
SchSeba, line 203: how do you know these are the PCUs you have on the system?

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func withNetworkNameAndNamespace(namespace, name string) func(*sriovv1.SriovNetworkNodePolicy, *sriovv1.SriovNetwork) {
+	return func(p *sriovv1.SriovNetworkNodePolicy, n *sriovv1.SriovNetwork) {
+		n.ObjectMeta.Name = name
+		n.Spec.NetworkNamespace = namespace
+	}
+}
+
+func createSriovNetworkAndPolicy(opts ...func(*sriovv1.SriovNetworkNodePolicy, *sriovv1.SriovNetwork)) {
```
SchSeba, line 312: please move all this functions out of the DPDK package. and we already have some of them in the pkg folder please take a look

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,342 @@
...
+		perfProfile := performanceprofile.DefineSingleNUMANode("single-numa-node-pp", machineConfigPoolName)
+		err = performanceprofile.OverridePerformanceProfile("single-numa-node-pp", machineConfigPoolName, perfProfile)
+		Expect(err).ToNot(HaveOccurred())
+
+		sriovDevices, err := sriovCapableNodes.FindSriovDevices(testNode.Name)
+		Expect(err).ToNot(HaveOccurred())
+
+		numa0DeviceList, err = findDevicesOnNUMANode(testNode, sriovDevices, "0")
+		Expect(err).ToNot(HaveOccurred())
+		numa1DeviceList, err = findDevicesOnNUMANode(testNode, sriovDevices, "1")
```
SchSeba, line 67: we need to skip the tests if the node has only one numa not failed and that can be the case for RAN servers

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,342 @@
...
+	var numa1DeviceList []*sriovv1.InterfaceExt
+
+	BeforeAll(func() {
+		if discovery.Enabled() {
+			Skip("Discovery mode not supported")
+		}
+
+		isSNO, err := utilNodes.IsSingleNodeCluster()
+		Expect(err).ToNot(HaveOccurred())
+		if isSNO {
```
SchSeba, line 39: why we don't support SNO here?

cnf-tests/testsuites/e2esuite/dpdk/numa_node_sriov.go
```diff
@@ -0,0 +1,342 @@
+package dpdk
+
+import (
+	"context"
```
SchSeba, line 4: please re-arrange the imports as in all the other files

PR #1576, opened 2023-07-23
cnf-tests: log test pod events on failure

liornoy:
Add a print to the test pod's events on test failure, to help with troubleshooting.

Example for the output of GetStringEventsForPod:
`2023-07-23 14:14:54 +0300 IDT: Successfully assigned l2-6521/external-local-lb-lgwgg to kind-worker 2023-07-23 14:14:55 +0300 IDT: Container image "registry.k8s.io/e2e-test-images/agnhost:2.43" already present on machine 2023-07-23 14:14:55 +0300 IDT: Created container netexec 2023-07-23 14:14:55 +0300 IDT: Started container netexec`

### DIFF
--- a/cnf-tests/testsuites/e2esuite/sctp/sctp.go
+++ b/cnf-tests/testsuites/e2esuite/sctp/sctp.go
@@ -26,6 +26,7 @@ import (
 	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/images"
 	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/namespaces"
 	utilNodes "github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/nodes"
+	"github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/pods"
 
 	"k8s.io/utils/pointer"
 )
@@ -309,7 +310,7 @@ func startServerPod(node, namespace string, networks ...string) *k8sv1.Pod {
 		res, err = client.Client.Pods(namespace).Get(context.Background(), serverPod.Name, metav1.GetOptions{})
 		Expect(err).ToNot(HaveOccurred())
 		return res.Status.Phase
-	}, waitForPodRunningTimeout*time.Minute, 1*time.Second).Should(Equal(k8sv1.PodRunning))
+	}, waitForPodRunningTimeout*time.Minute, 1*time.Second).Should(Equal(k8sv1.PodRunning), pods.GetStringEventsForPodFn(client.Client, res))
 	return res
 }
 
@@ -343,6 +344,8 @@ func checkForSctpReady(cs *client.ClientSet) {
 }
 
 func testClientServerConnection(cs *client.ClientSet, namespace string, destIP string, port int32, clientNode string, serverPodName string, shouldSucceed bool, networks ...string) {
+	var pod *k8sv1.Pod
+
 	By("Connecting a client to the server")
 	clientArgs := []string{"-ip", destIP, "-port",
 		fmt.Sprint(port), "-lport", "30102"}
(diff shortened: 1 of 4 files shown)


cnf-tests/testsuites/e2esuite/sctp/sctp.go
```diff
@@ -356,10 +359,10 @@ func testClientServerConnection(cs *client.ClientSet, namespace string, destIP s
 
 	if !shouldSucceed {
 		Consistently(func() k8sv1.PodPhase {
-			pod, err := cs.Pods(namespace).Get(context.Background(), serverPodName, metav1.GetOptions{})
+			pod, err = cs.Pods(namespace).Get(context.Background(), serverPodName, metav1.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
 			return pod.Status.Phase
-		}, 30*time.Second, 1*time.Second).Should(Equal(k8sv1.PodRunning))
+		}, 30*time.Second, 1*time.Second).Should(Equal(k8sv1.PodRunning), fmt.Sprintf("pod %s failed to be in running state. Events:\n %s", pod.Name, pods.GetStringEventsForPod(cs, pod)))
```
zeeke, line 365: `pods.GetStringEventsForPod(cs, pod)` is invoked too early, before the first `Consistently` loop. You are probably looking for a lazy evaluation:

```
(...).Should(
    Equal(k8sv1.PodRunning), 
    func () string {
        return fmt.Sprintf("pod %s failed to be in running state. Events:\n %s", pod.Name, pods.GetStringEventsForPod(cs, pod)))
    })
```

And Gomega supports that:
https://onsi.github.io/gomega/#annotating-assertions

liornoy: Thanks for your comment. it really is a crucial point. fixed

liornoy: Thanks for your comment. it really is a crucial point. fixed

cnf-tests/testsuites/pkg/pods/pods.go
```diff
@@ -413,3 +413,17 @@ func DetectDefaultRouteInterface(cs *testclient.ClientSet, pod corev1.Pod) (stri
 	}
 	return "", fmt.Errorf("default route not present")
 }
+
+func GetStringEventsForPod(cs *testclient.ClientSet, pod *corev1.Pod) string {
```
zeeke, line 417: Please, use `corev1client.EventsGetter` as it's more portable.

(`correv1client "k8s.io/client-go/kubernetes/typed/core/v1"`)

cnf-tests/testsuites/e2esuite/bond/bond.go
```diff
@@ -20,15 +20,15 @@ import (
...
+var _ = Describe("[bondcni]", Ordered, func() {
 	apiclient := client.New("")
 
 	execute.BeforeAll(func() {
 		err := namespaces.Create(namespaces.BondTestNamespace, apiclient)
 		Expect(err).ToNot(HaveOccurred())
 	})
 
-	AfterEach(func() {
+	AfterAll(func() {
```
zeeke, line 31: Does it work? In this suite, we have only one test case, so AfterAll will be called at a very similar timing to AfterEach, doesn't it?

zeeke (commented):
Left some comments. My main concern is about replacing AfterEach with AfterAll, as:
a. I'm not sure the reporter is triggered after AfterEach and before AfterAll
b. When we add more tests to the suites, we'll have to care about the cleaning mechanism, as it's at the end of the suite.

+1 for the event reporter. I think printing much information to the console in case of assertion failing is better than relying on the reporter.

cnf-tests/testsuites/e2esuite/bond/bond.go
liornoy: you're right. I didn't test it and was sure the ReportAfterEach will come before the AfterAll, but in reality, it's not.
So I'll fix it by adding the last cleanup in the last test as we had before.

cnf-tests/testsuites/pkg/pods/pods.go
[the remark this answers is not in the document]
liornoy: done thanks

liornoy:
@zeeke I removed the commit with the change to the cleanup order because I want the other commit that adds the logs of the events to merge so we can faster catch the sriov-vrf flake.

zeeke:
Thanks,

/lgtm

zeeke:
`ci/prow/e2e-gcp-ovn` failing tests are not related to this PR

/override ci/prow/e2e-gcp-ovn

cnf-tests/testsuites/pkg/pods/pods.go
```diff
@@ -413,3 +414,23 @@ func DetectDefaultRouteInterface(cs *testclient.ClientSet, pod corev1.Pod) (stri
 	}
 	return "", fmt.Errorf("default route not present")
 }
+
+func GetStringEventsForPod(cs corev1client.EventsGetter, pod *corev1.Pod) string {
```
SchSeba, line 418: nit: this function can be private to the package

liornoy: thanks, fixed

PR #1577, opened 2023-07-25
Support node deletion using ZTP

Missxiaoguo:
As discussed, the SiteConfig API is updated with the following fields:
- crAnnotations: at site,cluster and node level to allow user add new annotations to a default install CR
- crSuppression: at cluster and node level to allow user to suppress the generation of a default install CR

The supported Gitops workflow for node deletion in ACM is first annotating the BMH CR on the hub cluster with the annotation "bmac.agent-install.openshift.io/remove-agent-and-node-on-delete" and then deleting the BMH CR.  A simple instruction is added for the workflow using ZTP. It can be updated with more details and a reference link to ACM doc when it's available in ACM.

/cc @lack @serngawy @imiller0


jgato:
sorry, do we know if this feature will reach ZTP4.14?

ztp/gitops-subscriptions/argocd/WorkerNodeDeletion.md
```diff
@@ -0,0 +1,52 @@
...
+1. ACM 2.8+ with MultiClusterHub created and configured, running on OCP 4.13+ bare metal cluster
+
+## Delete a worker node from spoke cluster
+1. Annotate the BMH CR of the worker node with the "bmac.agent-install.openshift.io/remove-agent-and-node-on-delete=true"    annotation. Add the annotation via SiteConfig as the following, then push the changes to git repo and wait for the BMH CR on the hub cluster has the annotation applied.
+    ```yaml
+    nodes:
+    - hostname: node6
+      role: "worker"
+      crAnnotations:
+        add:
```
serngawy, line 15: Why do we need to mentioned "add" ? can we remove "add" and become like 
` BareMetalHost:`
`    bmac.agent-install.openshift.io/remove-agent-and-node-on-delete: true`

ztp/ran-crd/site-config-crd.yaml
```diff
@@ -76,6 +76,22 @@ spec:
...
                     type: string
+                crAnnotations:
+                  description: |
+                    An optional CR annotations parameter enables users to patch
+                    the default generated CRs with defined annotations. The annotations
+                    defined in clusters or nodes take precedence over the annotations
+                    at this level.
+                  type: object
+                  properties:
+                    add:
```
serngawy, line 87: Same as above why do we need to mentioned Add ? Is it possible to have delete/remove as well ?

ztp/ran-crd/site-config-crd.yaml
```diff
@@ -288,6 +304,30 @@ spec:
...
+                        properties:
+                          add:
+                            description: |
+                              An map of CR type names to annotations which will be added
+                              to the default generated CR if they do not exist. The key is
+                              the Kind name of the CR to be patched (such as "KlusterletAddonConfig").
+                            type: object
+                            additionalProperties:
+                              type: string
+                      crSuppression:
```
serngawy, line 322: Can we simplify the crSuppression like in the node level when crSuppression is True SiteConfig does not generate BMH, NMState AND if crSuppression is True in cluster level SiteConfig does not generate any CR. 
What do you think ?

serngawy (changes requested):

ztp/ran-crd/site-config-crd.yaml
Missxiaoguo: In the paricular case of the node deletion, we only want to delete BMH but keep other CRs remain on the hub cluster. 
There may have other user cases for deleting other types of CRs (in the further), so it was decided to add a generic option. 
Another usercase is detaching a cluster by deleting only the ManagedCluster and retaching by removing the ManagedCluster CR from crSuppression at the cluster level.

Missxiaoguo:
> sorry, do we know if this feature will reach ZTP4.14?

QE/Doc didn't have capacity for this story so it was not added in the 4.14. The plan for this is 4.15 and we can back port this to 4.14 if needed.

ssugimot:
Hi,
Thank you for the PR. I am one of those who are willing to have this feature (ZTP support for scale-in node(s)). I went through the changes to the document and understood the expected operation to edit siteconfig. Let me address my view.

I understand that there is a need to indicate the necessity of removing a node by setting annotation to the corresponding BMH. IMHO, a better approach to handle this would be to let user remove the definition of worker node from the siteconfig. When a node gets removed from the siteconfig, it will be detected by the operator and necessary steps will be taken. Setting an annotation to the corresponding BMH can be one of them. In this way, I think siteconfig can remain declarative and user can focus on expressing desired state of the cluster in siteconfig. My two cents.

Missxiaoguo:
@ssugimot  Thanks for your comment!

Regarding the question about removing the definition of a worker node from siteconfig to trigger all actions to happen automatically in one shot, it will not work in the GitOps way. Argocd must first apply the changes for BMH annotation and have it synced on the live cluster before deleting or suppressing the generation of the annotated BMH. Therefore, two commits are required one by one. If there is only one action/commit to do that, it will end up deleting the BMH without the annotation, which is not the expected behavior supported in the assisted service.

Regarding the question about triggering node deletion by explicitly setting the BMH cr in a new field, there has been some discussion around this. As node deletion requires the NMstate CR to be preserved but only have this BMH CR deleted, it is not possible to wipe the whole worker section in siteconfig as it will clean up all CRs. Initially, it was considered to provide a "detach" option in the node section. When set to true, it triggers the deletion. However, we were thinking about other use cases that may require CR(s) deletion. One is detaching a cluster from a hub cluster by deleting only the ManagedCluster CR and re-importing it back after. There may be other use cases for deleting other types of CRs (in the future), so it was decided to add a generic option instead.

I hope this explanation is clear enough!

ssugimot:
@Missxiaoguo Thank you for your detailed explanation!

I see that things are not so simple.. Understood.

imiller0 (approved):
/lgtm

PR #1584, opened 2023-07-31
ztp: update performance profile to load vfio-pci

SchSeba:
load the vfio-pci driver on boot with the parameters needed by the sriov-fec


imiller0 (commented):
Do these parameters have any impact other than when the FEC operator is in use?

imiller0 (approved):
/lgtm

PR #1585, opened 2023-08-01
OCPBUGS-7092 SiteConfig crTemplates override does not work with a ConfigMap CR

serngawy:
SiteConfig crTemplates override does not work with a ConfigMap CR

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/testdata/ConfigMapOverride.yaml
+++ b/ztp/siteconfig-generator/siteConfig/testdata/ConfigMapOverride.yaml
@@ -1,0 +1,10 @@
+# Extra manifest will be added to the data section
+kind: ConfigMap
+apiVersion: v1
+metadata:
+    annotations:
+        argocd.argoproj.io/sync-wave: "1"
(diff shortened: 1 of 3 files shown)


ztp/siteconfig-generator/siteConfig/testdata/ConfigMapOverride.yaml
```diff
@@ -0,0 +1,10 @@
+# Extra manifest will be added to the data section
+kind: ConfigMap
+apiVersion: v1
+metadata:
+    annotations:
+        argocd.argoproj.io/sync-wave: "1"
+        overrideconfigmap-annotation/test: "new-annotation"
+    name: "{{ .Cluster.ClusterName }}"
+    namespace: "{{ .Cluster.ClusterName }}"
+data:
```
lack, line 10: To make this valid, should we maybe have this point at an empty map:
```suggestion
data: {}
```

serngawy: done

PR #1421, opened 2023-03-06
cnf-tests: Add rootless vhostnet test using tap and vlan plugins

mlguerrero12:

PR #1591, opened 2023-08-03
s2i: add supplementary group 801 to 1001 user

mlguerrero12:
In a non-priviledged pod, user 1001 needs to be part of group 801 to be able to access the vfio device.


zeeke:
/lgtm
/approve

To add context, 801 is the groupID used by:
- https://github.com/openshift-kni/cnf-features-deploy/pull/1421

SchSeba (approved):
/lgtm
/approve

PR #1594, opened 2023-08-08
ztp: [release 4.13] update fec config to use vfio-pci

SchSeba:
this commit also contains the needed parameters for the vfio driver


imiller0 (commented):
/lgtm

PR #1595, opened 2023-08-08
OCPBUGS-3513 SiteConfig uses deprecated manifestsConfigMapRef field

serngawy:


ztp/siteconfig-generator/siteConfig/clusterCRsV1.go
```diff
@@ -0,0 +1,185 @@
+package siteConfig
+
+// SiteConfigV1 and ClusterCRsV1 are deprecated use SiteConfigV2 and ClusterCRsV2 to add new field.
+// SiteConfigV1 uses clusterCRsV1 to generate the required CRs for cluster provisioning
+
+const siteConfigAPIV1 = "ran.openshift.io/v1"
+
+const clusterCRsV1 = `
```
lack, line 8: Is this v1/v2 split a good time to consider using [go:embed](https://pkg.go.dev/embed) instead of hard-coding strings in code?

lack:
We should probably update the docs explaining the difference and what the expected customer actions are when moving to v2

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -285,6 +303,16 @@ func (scbuilder *SiteConfigBuilder) getClusterCR(clusterId int, siteConfigTemp S
 				return mapIntf, err
 			}
 			mapIntf[k] = value
+		} else if reflect.ValueOf(v).Kind() == reflect.Slice {
```
sabbir-47, line 306: Can we get rid of else if with switch blocks for better readability?

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -25,29 +25,47 @@ type SiteConfigBuilder struct {
...
-	return &scBuilder, nil
+	return nil
 }
 
 func (scbuilder *SiteConfigBuilder) SetLocalExtraManifestPath(path string) {
 	scbuilder.scBuilderExtraManifestPath = path
 }
 
 func (scbuilder *SiteConfigBuilder) Build(siteConfigTemp SiteConfig) (map[string][]interface{}, error) {
+	if siteConfigTemp.ApiVersion == siteConfigAPIV1 {
```
sabbir-47, line 55: if we extend API version later to v3,v4,v5, probably better to use switch instead of "else if" for better readability

imiller0:
/hold
A change in api version should be put in place after we branch for the 4.14 release (we are now past code complete).

ztp/siteconfig-generator/siteConfig/clusterCRsV1.go
serngawy: not sure what id the benfites of using go:embed rather than defining the version as const.  will separate the version from the apiGroup.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: done

ztp/siteconfig-generator/siteConfig/clusterCRsV1.go
lack: The benefit is that the actual template CRs would be in separate files in the filesystem instead of just inline inside a go module.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
serngawy: okay, done

PR #1599, opened 2023-08-09
[release-4.12] Bump metallb vendor

liornoy:
Bump metallb vendor to latest commit on release-4.12
This is so we include a fix that extends a timeout for a flaky test case.


liornoy:
@sshnaidm this is suppose to fix the metallb flake in cnf-tests on branch 4.12, but due to severe conflicts issues
of the go versions I think we should wait for the cnf-tests split where it will overcome it by running from the local cloned metallb-operator project, imo trying to resolve this conflict will be difficult and time consuming.

PR #1605, opened 2023-08-11
ztp: Align reference PGT with vector

imiller0:
CNF-9132 Align the reference PolicyGenTemplates with use of vector for log collection. The PGT simply use the source CR to align with default collector. Update test code and README to align.


imiller0:
Note, this PR completes the update made under PR 1604

PR #1610, opened 2023-08-17
OCPBUGS-13313: fix: error in messaging for cpu partitioning

eggfoobar:
We incorrectly merged in messaging that stated this feature was available in 4.13, this is a 4.14 feature. Updated logic to allow users to use it as a TechPreview flag in 4.13 as originally intended but with correct wording

/assign @imiller0


imiller0 (approved):
This fix is unique/specific to 4.13 where cpuPartitioningMode is TechPreview. In 4.14 it is a GA feature and this warning will no longer apply. So a direct fix in 4.13 branch without corresponding change in 4.14+ is needed.
/lgtm

imiller0 (approved):
/lgtm

PR #1612, opened 2023-08-17
ztp: Remove redundant overlay content

imiller0:
The overlay content for the tuned patch in the group profile is identical to the content in the base source-cr. There is no need to maintain (or recommend to the user) the overlay content.


imiller0:
/hold
Pending verification of the acm policygen variants.

serngawy (approved):

PR #1602, opened 2023-08-10
[release-4.13] Log pod events on failure

liornoy:
Add a print to the test pod's events on test failure, to help with troubleshooting.

Example for the output of GetStringEventsForPod:
2023-07-23 14:14:54 +0300 IDT: Successfully assigned l2-6521/external-local-lb-lgwgg to kind-worker 2023-07-23 14:14:55 +0300 IDT: Container image "registry.k8s.io/e2e-test-images/agnhost:2.43" already present on machine 2023-07-23 14:14:55 +0300 IDT: Created container netexec 2023-07-23 14:14:55 +0300 IDT: Started container netexec

PR #1622, opened 2023-08-28
OCPBUGS-15790, OCPBUGS-15791: cnf-tests: Use `lspcu --parse` to get NUMA  node

zeeke:
If a node has many CPUs, `lscpu` command output may split over multiple lines, making it hard to get the list of CPUs per NUMA node.

Switching to `lscpu --parse=cpu,node` simplifies the parsing logic.

cc @SchSeba


SchSeba (approved):
this looks good nice work!

btw do we need to add a lane for unit test on the pkg folder?
we can do it in a separate PR for sure :)

/lgtm
/approve

liornoy:
This is being backported to 4.13 in: https://github.com/openshift-kni/cnf-features-deploy/pull/1602
We omit the `numa_node_sriov.go` file which is not a supported feature in 4.13.

PR #1627, opened 2023-08-29
OCPBUGS-17660: Disable rcu_expedited after node has finished booting

fontivan:
- This is accomplished by setting "rcu_normal" to 1


fontivan:
Although I have tested this manually by oc apply'ing the generated MC I want to verify it works straight from installation before allowing this to merge

imiller0:
Note that your commit message needs to be prefixed with "ztp: ...."

fontivan:
Finished retesting this morning with installation, all looking good from my perspective

imiller0 (approved):
/lgtm

PR #1633, opened 2023-09-06
cnf-tests: fix `ApplyKubeletConfigToNode` cleanup

zeeke:
When `ApplyKubeletConfigToNode` moves the node back to the previous role, we have to wait for the relative MachineConfigPool to update the node. Since the function has no clue about which MCP the node belonged to before applying the KubeletConfig, the best strategy is to wait for at least one MCP to start updating and then wait for the whole cluster (every MCP) to be stable.

Before these changes, the test case that runs after the cleanup function is likely to fail, as it found a cluster in a non-stable state.

Introduce utility functions `WaitForAllMCPStable()` and `WaitForAtLeastOneMCPUpdating()` to query the state of the cluster without specifying a single MCP.

Failing job example: [HTML Report](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.15-e2e-telco5g-cnftests/1699249554244767744/artifacts/e2e-telco5g-cnftests/telco5g-cnf-tests/artifacts/test_results.html), [Prow Job](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.15-e2e-telco5g-cnftests/1699249554244767744)

### DIFF
--- a/cnf-tests/testsuites/pkg/machineconfigpool/machineconfigpool.go
+++ b/cnf-tests/testsuites/pkg/machineconfigpool/machineconfigpool.go
@@ -25,17 +25,18 @@ func WaitForCondition(
 	conditionStatus corev1.ConditionStatus,
 	timeout time.Duration,
 ) error {
+	klog.Infof("Waiting for MCP %s: %s == %s", mcp.Name, conditionType, conditionStatus)
 	return wait.PollImmediate(3*time.Second, timeout, func() (bool, error) {
 		mcpUpdated, err := cs.MachineConfigPools().Get(context.Background(), mcp.Name, metav1.GetOptions{})
 		if err != nil {
 			return false, nil
 		}
 
-		for _, c := range mcpUpdated.Status.Conditions {
-			if c.Type == conditionType && c.Status == conditionStatus {
-				return true, nil
-			}
+		if mcpHasCondition(mcpUpdated, conditionType, conditionStatus) {
+			klog.Infof("Condition met for MCP %s: %s == %s", mcp.Name, conditionType, conditionStatus)
+			return true, nil
 		}
+
 		return false, nil
 	})
 }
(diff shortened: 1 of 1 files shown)


cnf-tests/testsuites/pkg/machineconfigpool/machineconfigpool.go
```diff
@@ -25,17 +25,18 @@ func WaitForCondition(
 	conditionStatus corev1.ConditionStatus,
 	timeout time.Duration,
 ) error {
+	klog.Infof("Waiting for MCP %s: %s == %s", mcp.Name, conditionType, conditionStatus)
 	return wait.PollImmediate(3*time.Second, timeout, func() (bool, error) {
 		mcpUpdated, err := cs.MachineConfigPools().Get(context.Background(), mcp.Name, metav1.GetOptions{})
```
liornoy, line 30: I was wondering why we call the mcp var here mcpUpdated,
because we might want to check other condition other than updated.
Also, why do we do "Get" if we already have the object as an input?

cnf-tests/testsuites/pkg/machineconfigpool/machineconfigpool.go
```diff
@@ -129,6 +130,50 @@ func WaitForMCPUpdated(mcp mcov1.MachineConfigPool) error {
...
+		if err != nil {
+			return false, nil
+		}
+
+		for _, mcp := range mcpList.Items {
+			if mcpHasCondition(&mcp, mcov1.MachineConfigPoolUpdating, corev1.ConditionTrue) {
+				klog.Infof("MCP %s is updating", mcp.Name)
+				return true, nil
+			}
+		}
```
liornoy, line 146: Do we not use the `WaitForCondition` helper function here on purpose?

cnf-tests/testsuites/pkg/machineconfigpool/machineconfigpool.go
zeeke: Both questions have the same answer: we need to fetch a "fresh" copy of the object every time the loop runs. Otherwise, the mcp.Status.Conditions field would never change. I guess `mcpUpdated` has been used to indicate it's an "updated version of the mcp". That variable name comes from a previous commit, not this change.

I can the name of that variable as part of this PR if you have proposals

liornoy: thanks for the clarification. the name change is no must.

cnf-tests/testsuites/pkg/machineconfigpool/machineconfigpool.go
zeeke: Unfortunately, we can't. `WaitForCondition()` waits on a single MCP to meet a condition. This function loops over all MCPs, jumping to the next one if the condition is not met.

If you have an idea on how to refactor this, please share it.

PR #1646, opened 2023-09-12
ztp: CNF-9171: DU reference config with hub side templating

irinamihai:
Description:
- add readme to propose options for grouping source CRs into PGTs for when using templating in the PGTs
- add ConfigMap example to be used with hub side templating
- add SNO site, SNO & 3NC group PGTs including usage of:
    * both .ManagedClusterName .ManagedClusterLabels templating vars
    * zoning and hardware-type labels


serngawy:
The ci job is failing because of the commit message, the commit message must start with **ztp** and would you add the example for the [acmPolicyGenerator](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example/acmpolicygenerator) as well.

irinamihai:
> The ci job is failing because of the commit message, the commit message must start with **ztp**. Would you add the example for the [acmPolicyGenerator](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example/acmpolicygenerator) as well.
@serngawy, this is a good point. @imiller0 , what do you think? I think it would make sense to keep the two reference config examples consistent.

ztp/gitops-subscriptions/argocd/example/policygentemplates/hub-side-templating/template-group-du-sno-ranGen.yaml
```diff
@@ -0,0 +1,254 @@
...
+      policyName: "sriov-policy"
+      metadata:
+        name: sriov-nw-du-mh # FIXED
+      spec:
+        resourceName: du_mh # FIXED
+        vlan: '{{hub fromConfigMap "" (printf "site-%s-configmap" .ManagedClusterName) (printf "%s-sriov-network-vlan-2" .ManagedClusterName) | toInt hub}}'
+        ipam: |
+          {
+            "type": "host-local",
+            "subnet": '{{hub fromConfigMap "" (printf "site-%s-configmap" .ManagedClusterName) (printf "%s-sriov-network-ipam-subnet-2" .ManagedClusterName) hub}}',
```
irinamihai, line 203: Maybe this would be more suitable for a zone label rather than .ManagedClusterName

ztp/gitops-subscriptions/argocd/example/policygentemplates/hub-side-templating/template-group-du-sno-ranGen.yaml
```diff
@@ -0,0 +1,254 @@
...
+            "rangeStart": "10.56.217.171",
```
irinamihai, line 204: Template the rangeStart, rangeEnd and gateway

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/hub-side-templating/acm-example-multinode-site.yaml
```diff
@@ -0,0 +1,47 @@
+# For this PolicyGenerator, we've considered a multinode managed cluster with the following label(s):
```
serngawy, line 1: Better to refer here to the ConfigMap that will to be used with this template. Same apply for all templates.

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/hub-side-templating/ns.yaml
```diff
@@ -0,0 +1,5 @@
+---
```
serngawy, line 1: No need for this , the NS already defined at https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/ns.yaml#L11

serngawy (commented):
Few comments, AND we need to add the template files & configMaps to the kustomization.yaml for both PGT and ACM PolicyGen

irinamihai:
> Few comments, AND we need to add the template files & configMaps to the kustomization.yaml for both PGT and ACM PolicyGen

@serngawy , I did think about that, but then again, we should only have one of the PG(T)s. I can add the templated PG(T)s in the kustomization.yaml and a comment explaining that either the one in the main directory or the one from the hub-side-templating directory should be used. Another option would have been to have a complete different kustomization.yaml and ns.yaml in the hub-side-templating directory, but that would mean to duplicate the common policies, which I didn't want.
I'll stick to the initial approach.

ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/kustomization.yaml
```diff
@@ -1,21 +1,31 @@
 generators:
 # This is common to all RAN deployments
 - acm-common-ranGen.yaml
-# This group policy is for all single-node deployments:
+# This group policy is for all single-node deployments. Use just one of the 2 below.
 - acm-group-du-sno-ranGen.yaml
+#- hub-side-templating/acm-group-du-sno-ranGen.yaml
```
serngawy, line 6: better to un-comment this line. We need the ci-job validate the change to template policies to avoid any changes that not recognized. Same for all template policies

ztp/gitops-subscriptions/argocd/example/policygentemplates/kustomization.yaml
```diff
@@ -1,24 +1,34 @@
 generators:
 # This is common to all RAN deployments
 - common-ranGen.yaml
-# This group policy is for all single-node deployments:
+# This group policy is for all single-node deployments. Use just one of the 2 below.
 - group-du-sno-ranGen.yaml
-# This group validator policy is for all single-node deployments:
+#- hub-side-templating/group-du-sno-ranGen.yaml
```
serngawy, line 6: same un-comment this and let the ci-job validate the generated policy

ztp/gitops-subscriptions/argocd/example/policygentemplates/hub-side-templating/site-du-standard-configmap.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
serngawy, line 1: Is it possible to create sys-link instead of creating exact same file twice ? Same for all repeated CM files

irinamihai: Unfortunately, this doesn't work with kustomize, as the files can't be in a directory higher up than the kustomization.yaml.

serngawy:
sounds good to me

PR #1660, opened 2023-09-19
Update feature-deploy.sh

lazharh:


yuvalk:
Hi @lazharh 
tnx for the contrib

but I dont think it's smart to remove that comment.

I think that kubectl PR is already merged and available in OC
so a smarter move would be to move back to the purer `oc apply -k`
can also stop downloading kubectl separately

best to do that be reverting the original commit.

lazharh:
closing

PR #1676, opened 2023-10-16
cnf-tests: add git submodule dependency

liornoy:
The Makefile target 'test-bin' needs to have the Git submodules initiated. Here, we add 'init-git-submodule' as a dependency for the 'test-bin' target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ custom-rpms:
 	@echo "Installing rpms"
 	RPMS_SRC="$(RPMS_SRC)" hack/custom_rpms.sh
 
-test-bin:
+test-bin: init-git-submodules
 	@echo "Making test binary"
 	cnf-tests/hack/build-test-bin.sh
 


Makefile
```diff
@@ -116,7 +116,7 @@ custom-rpms:
 	@echo "Installing rpms"
 	RPMS_SRC="$(RPMS_SRC)" hack/custom_rpms.sh
 
-test-bin:
+test-bin: init-git-submodules
```
ffromani, line 119: does this (actually) run correctly only once if submodules are not inited, right?

liornoy: This will not fail when they are already initiated.
Anyway this "test-bin" target will only be used by the cnf-tests image builder now.

liornoy: Are you ok with overriding gcp?

jlojosnegros: It really runs every time either the submodules are already initiated or not, but running it multiple times does not change the output, so it should be safe.

ffromani:
/override ci/prow/e2e-gcp-ovn

change needed to unblock, but the failure here needs to be investigated

PR #1677, opened 2023-10-25
ztp: CNF-9171: siteconfig-generator: generate new ConfigMap

irinamihai:
Description:
- add a new cluster field in the site-config-crd: siteConfigMap
- add new ConfigMap in the clusterCRsV2
- update the siteConfig-generator to create a new ConfigMap with the data from the siteConfigMap. The purpose for this data is to further be used by the ZTP PGT / ACM PG


ztp/ran-crd/site-config-crd.yaml
serngawy, line 1: Could you create siteConfig v2 CRD and add the siteConfigMap to it , lets make new changes to siteConfig v2 and keep v1 as it is. What do you think ?

serngawy (changes requested):

ztp/ran-crd/site-config-crd.yaml
sabbir-47: Yes, only add the new feature/work to v2 only. If any backport is required, only then add to v1 and backport to a targeted version.

ztp/siteconfig-generator-kustomize-plugin/testSiteConfig/site2-standard-du.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   sshPrivateKeySecretRef:
     name: "sshPrvKey"
   biosConfigRef:
-    filePath: "testSiteConfig/testHW.profile"
+    filePath: "../siteconfig-generator-kustomize-plugin/testSiteConfig/testHW.profile"
```
sabbir-47, line 16: No need to change paths. for kustomize, the visible path starts after siteconfig-generator-kustomize-plugin/

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -495,6 +496,34 @@ type BiosConfigRef struct {
...
+// Provide custom YAML unmarshal for SiteConfigMap which provides default values
+func (rv *SiteConfigMap) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type ValueDefaulted SiteConfigMap
+	var defaults = ValueDefaulted{
+		Name:      "ztp-site-configmap",
+		Namespace: "ztp-site",
+	}
+
+	out := defaults
+	err := unmarshal(&out)
```
sabbir-47, line 514: This is more readable and follows Go's error handling:

*rv = SiteConfigMap(out)
err := unmarshal(&out)
if err != nil {
		return err
	}
return nil

ztp/ran-crd/site-config-crd.yaml
irinamihai: @serngawy & @sabbir-47, definitely, I think that's a good approach. Thank you.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -212,14 +212,23 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
...
-				// For ConfigMap, add all the ExtraManifest files to it before doing further instantiation:
-				data := instantiatedCR["data"]
-				if data != nil {
-					for k, v := range data.(map[string]interface{}) {
-						extraManifestMap[k] = v
+				configMapMetadata := mapSourceCR["metadata"].(map[string]interface{})
+				configMapName := configMapMetadata["name"].(string)
+				if strings.Contains(configMapName, "SiteConfigMap") {
+					// If there is no siteConfigMap specified, don't create the corresponding CR.
+					if cluster.SiteConfigMapIsEmpty() {
```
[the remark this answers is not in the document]
irinamihai: If the new siteConfigMap is empty in the siteConfig CR, I don't want it to be added to the clusterCRs. If it's not empty, then it should directly be added to the clusterCRs without any other processing.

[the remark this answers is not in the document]
sabbir-47: SiteConfigMap could be a pointer to SiteConfigMap struct, in that case any empty SiteConfigMap would return nil value and probably it can make the logic easier. Just a thought.

ztp/siteconfig-generator/siteConfig/siteConfig.go
irinamihai: I agree, but I wanted to keep it consistent with the rest of the file.

ztp/ran-crd/site-config-crd.yaml
```diff
@@ -140,6 +140,23 @@ spec:
...
                         type: object
                         x-kubernetes-preserve-unknown-fields: true
+                      siteConfigMap:
+                        description: The content of a site specific ConfigMap with the configuration that can be further used by PGT
+                        type: object
+                        properties:
+                          name:
+                            description: The name of the site specific ConfigMap
+                            type: string
+                            default: ztp-site-configmap
```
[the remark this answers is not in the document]
irinamihai: @serngawy, what is wrong would that, given each configMap is going to be for a different cluster?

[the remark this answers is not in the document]
serngawy: If i didn't change the default values (name & namespace) of the configMap for every siteConfig a configMap with name: ztp-site-configmap and namesapce: ztp-site will be generated. AND kustomize will raise error as multiple  CRs with same metadata/type are generated.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -495,6 +496,34 @@ type BiosConfigRef struct {
...
+	Name      string            `yaml:"name"`
+	Namespace string            `yaml:"namespace"`
+	Data      map[string]string `yaml:"data"`
+}
+
+// Provide custom YAML unmarshal for SiteConfigMap which provides default values
+func (rv *SiteConfigMap) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type ValueDefaulted SiteConfigMap
+	var defaults = ValueDefaulted{
+		Name:      "ztp-site-configmap",
```
serngawy, line 509: Same, we have to force enter the CM name. It needs to be a mandatory field.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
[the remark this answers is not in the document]
serngawy: The CM logic is fine, I mean the if checks something like below more clean code
`
if !strings.Contains(configMapName, "SiteConfigMap") {

// For ConfigMap, add all the ExtraManifest files to it before doing further instantiation:
  data := instantiatedCR["data"]
  if data != nil {
     for k, v := range data.(map[string]interface{}) {
		extraManifestMap[k] = v
    }
  }
 instantiatedCR["data"] = extraManifestMap
}

`

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -582,6 +588,11 @@ func Test_ExtraManifestSearchPath(t *testing.T) {
 		mapSourceCR := cr.(map[string]interface{})
 
 		if mapSourceCR["kind"] == "ConfigMap" {
+			configMapMetadata := mapSourceCR["metadata"].(map[string]interface{})
```
serngawy, line 591: Not sure, why do we need to add this , this siteConfig is v1 so it suppose to not have the siteConfigMap. Do I'm missing something here ?

irinamihai: You are right, they are not needed. I think they were here since before I rebased with the SiteConfig V2.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -537,6 +538,43 @@ type BiosConfigRef struct {
 // IronicInspect
 type IronicInspect string
 
+type SiteConfigMap struct {
+	Name      string            `yaml:"name"`
+	Namespace string            `yaml:"namespace"`
+	Data      map[string]string `yaml:"data"`
```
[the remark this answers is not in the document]
irinamihai: @lack, that's actually a great point and something I struggled with quite a bit.
Because of the hub templating current implementation, we need to configure the values of the ConfigMap as strings, but we can still achieve multiline values in the ConfigMap as follows:

- The configmap:
> zone-1-cluster-log-fwd-outputs: "[{\"name\":\"kafka\", \"type\":\"kafka-open\", \"url\":\"tcp://10.46.55.190:9092/test\", \"secret\":{\"name\":\"kafka-open-secret\"}}]"

- The template:
> zone-1-cluster-log-fwd-outputs: "[{\"name\":\"kafka\", \"type\":\"kafka-open\", \"url\":\"tcp://10.46.55.190:9092/test\", \"secret\":{\"name\":\"kafka-open-secret\"}}]"

Links from the reference config:
- [configMap,](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/example/policygentemplates/hub-side-templating/group-du-sno-ranGen-templated.yaml#L25)
- [template](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/example/policygentemplates/hub-side-templating/site-du-sno-configmap.yaml#L27)

I hope this answers your concern.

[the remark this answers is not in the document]
serngawy: The ConfigMap API define [data](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L6869) as map[string]string . So it should be always map[string]string

serngawy (approved):

PR #1678, opened 2023-10-26
ztp: CNF-10088 container image updated to rhel9

sabbir-47:
- updated the build and base image to rhel9
- locally generated  CRS using siteconfig and policygenTemplate binary from the image


serngawy:
sounds good @sabbir-47, would you check why aws-ran job keep failing looks like some package are not available with image upgrade.

serngawy (approved):

PR #1690, opened 2023-11-05
Fix cnf-tests container build

liornoy:
Currently, the "test-bin" target invokes the "init-git-submodules".
This is, in his turn, invoke "fetch -all" which creates connection
to the public github repo. However, in Red Hat's mid-stream container
image builder, we are creating the image in a disconnected environment.

So here we remove the call to the "init-git-submodules" target,
and instead only invoke "git submodule update --init --force".
Also because we only care about the latest commit of the NTO.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,8 +116,9 @@ custom-rpms:
 	@echo "Installing rpms"
 	RPMS_SRC="$(RPMS_SRC)" hack/custom_rpms.sh
 
-test-bin: init-git-submodules
+test-bin:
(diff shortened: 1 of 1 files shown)


zeeke:
```
[It] [rfe_id:27363][performance] CPU Management when pod runs with the CPU load balancing runtime class [test_id:32646] should disable CPU load balancing for CPU's used by the pod
```
Same test is failing on every PR. Not related to this PR

/override ci/prow/e2e-gcp-ovn

PR #1694, opened 2023-11-09
ztp: Add V2 to the SiteConfig CRD

irinamihai:
Description:
- as this is not a CRD that lives on an actual cluster and it's not (yet) generated as customary operator CRDs, the CRD yaml schema was manually modified to include the v2 version.
- both versiones are served, but just v2 has 'storage: true' to show that v2 is the recommended version.


irinamihai:
/assign @serngawy @lack @sabbir-47
Thank you.

serngawy (approved):

lack (approved):

PR #1696, opened 2023-11-13
OCPBUGS-23425: Bump Go to v1.19 in release-4.12 

liornoy:
In order to bump the metallb-operator 4.12 vendor, we must Bump Go to v1.19.
This requires us to change the go-builder images in the dockerfiles of ZTP and cnf-tests.

After this PR will merge, the release repository will update it's own dockerfiles automatically.
A following PR will be filed with the metallb-operator bump.


liornoy:
Hi @lack, we change here ZTP's Containerfile and bump Go to v1.19, because we need this to bump metallb-operator.
is it ok with you?

liornoy:
/cc @lack @sabbir-47 @imiller0

Can I please get a review? as this also changes ZTP?

lack:
On its face, the change seems fine an probably low-risk; but there are a couple concerns:
- Is this attached to a defect or story?  We should make sure change to release branches are attached to valid defects.
- Before we merge this, we need to ensure that the change to the builder does not impact the ability of the ZTP container to function.  Part of this should be tested by unit tests that are passing as part of this PR, but a second part require manual verification - We should ensure that the mechanism we use to patch ArgoCD still functions properly.

lack:
I have checked and the binaries in the ZTP containier built here are compatible with ArgoCD just fine.

PR #1701, opened 2023-11-17
Revert "ztp:container image updated to rhel9"

sabbir-47:
- This reverts commit 35ce1ce7efcb2fb66bd81182836e1b2f17e4eefb.
- Reverts the images based on rhel9 to rhel8


lack:
We should add a comment in the Dockerfile saying why, and (not necessarily in this PR) a new CI check to enforce this.

PR #1706, opened 2023-11-21
cnf-tests: Add +x permissions to test-run.sh

liornoy:
Add the missing execution permissions to the test-run.sh file. Because without it we cannot run the latency tests.


ffromani:
/lgtm

but how did it work before? Since when the execution bit is missing?

liornoy:
> /lgtm
> 
> but how did it work before? Since when the execution bit is missing?

https://github.com/openshift-kni/cnf-features-deploy/commit/d80cd8a9014c9cf4a68be7003a2841b2cd2928c8#diff-5ba41b7fd3979591f0924c16b9c787a325a3502dbdaedd008d7dd446d97dcb15

Missing since here.
I don't know how we missed it until now. we had many unexpected mid-stream adventures in the way.
I wonder how locally we didn't caught it will testing the PR.

ffromani:
> > /lgtm
> > but how did it work before? Since when the execution bit is missing?
> 
> [d80cd8a#diff-5ba41b7fd3979591f0924c16b9c787a325a3502dbdaedd008d7dd446d97dcb15](https://github.com/openshift-kni/cnf-features-deploy/commit/d80cd8a9014c9cf4a68be7003a2841b2cd2928c8#diff-5ba41b7fd3979591f0924c16b9c787a325a3502dbdaedd008d7dd446d97dcb15)
> 
> Missing since here. I don't know how we missed it until now. we had many unexpected mid-stream adventures in the way. I wonder how locally we didn't caught it will testing the PR.

Thanks for the archeology. I agree we really should improve the testing in this area. We should have caught this earlier.
I actually prefer to restore the permissions to the file rather than fix them in the Dockerfile, but let's move on and improve later

liornoy:
> /override e2e-gcp-ovn

@ffromani can you please?

ffromani:
> > /override e2e-gcp-ovn
> 
> @ffromani can you please?

what's the reason we want to override?

liornoy:
> > > /override e2e-gcp-ovn
> > 
> > 
> > @ffromani can you please?
> 
> what's the reason we want to override?

because it fails with a known failure from  nto, not related to this PR

ffromani:
> > > > /override e2e-gcp-ovn
> > > 
> > > 
> > > @ffromani can you please?
> > 
> > 
> > what's the reason we want to override?
> 
> because it fails with a known failure from nto, not related to this PR

and this PR needs to go urgently in or can we fix or at least mitigate the ovn issue? because overriding is something I'd do VERY sparingly.

liornoy:
and afaik the nto team are aware of this bug and working on this according to their priorities.

ffromani:
why is the OVN lane running 
```
Performance Test: [It] [rfe_id:27363][performance] CPU Management when pod runs with the CPU load balancing runtime class [test_id:32646] should disable CPU load balancing for CPU's used by the pod
```
?

do we NEED to run this test in this lane (why?) or it's a side effect of incorrect/obsolete configuration?
If so, let's fix this instead of overriding.


(which, yes, is a known issue on compute side)

liornoy:
> U load balancing runtime class

it's running this tc because it's part of the "1_performance" suite. which we always ran in this suite.

liornoy:
do we NEED that? I'm not the one to judge tbh.

liornoy:
can you elaborate on why you think we shouldn't test it?

ffromani:
It depends on the purpose of the lane, which truth to be told I can't recall atm.
If the OVN lane is testing the OVN flows (as the name suggests?) then this tests hardly belong here.
If the OVN lane is testing all the features on a OVN setup, to avoid regressions for example, then the tests belongs here and overriding here and now makes much more sense indeed.

ffromani:
/override e2e-gcp-ovn

asking around offline, it's the latter (testing on OVN-enabled cluster)

PR #1707, opened 2023-11-22
Add .snyk file for static code analysis

liornoy:
Add a '.snyk' file to exclude the vendor directory for
the snyk code analysis.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,9 @@
+exclude:
+  global:
+    - vendor/github.com/onsi/ginkgo/v2/internal/suite.go:
+      reason: Temporarily ignoring until a fix is ready.
+      expires: 2024-03-01
+      created: 2024-01-01
+    - vendor/sigs.k8s.io/controller-runtime/pkg/log/log.go:
+      reason: Ignoring as the issue in the code is expected.
+      created: 2024-01-01


.snyk
```diff
@@ -0,0 +1,17 @@
+# References:
+# https://docs.snyk.io/scan-applications/snyk-code/using-snyk-code-from-the-cli/excluding-directories-and-files-from-the-snyk-code-cli-test
+# https://docs.snyk.io/snyk-cli/commands/ignore
+exclude:
+  global:
+    - "cnf-tests/submodules/metallb-operator/**"
+    - "cnf-tests/submodules/sriov-network-operator/**"
```
ffromani, line 7: if it works, I'd go as far as
```
- "cnf-tests/submodules/*/vendor/**"
```

ffromani: there are two things
1. we are ignoring _everything_ in the metallb-operator and sriov-network-operator submodels. Is that what we want to do?
2. do we want to ignore all submodules vendors?

liornoy: In my mind we only care about what's related to the cnf-tests image that we ship to customers.
and metallb+sriov are not part of it.

And in the NTO submodule vendor I don't want to ignore all because for example we compile and include ginkgo lib in the image.

liornoy: whereas on different repos they don't care about the ginkgo lib which is only used for testing and not shipped as part of the product binary. does that makes sense?

liornoy:
PR updated to exclude all vendor directory.
note that the vendor directories under the submodules are not tested,
as appears in this following rehearsal: 

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/45844/rehearse-45844-pull-ci-openshift-kni-cnf-features-deploy-master-security/1740331405809815552

SchSeba:
Hi @liornoy general question why do we want to skip the vendor folder??
that is loaded by our program meaning that if there is a CVE in one of our modules that we take as vendor we will be affected by it.

liornoy:
Thanks, @SchSeba . I initially considered focusing solely on scanning the code we wrote, a method I observed in other projects on Openshift. However, it does make sense to include the vendor directory in the scan as well. I'll revise the exclude list to specifically target problematic paths and update the commit message to provide a clear rationale.

.snyk
```diff
@@ -0,0 +1,7 @@
+# https://docs.snyk.io/scan-applications/snyk-code/using-snyk-code-from-the-cli/excluding-directories-and-files-from-the-snyk-code-cli-test
+exclude:
+  global:
```
SchSeba, line 3: I don't think we need this.

I am fine if the lane fails so we know there is a problem and we can contribute to this two projects to fix this issue :)

I can take the controller-runtime one no problem

SchSeba: I just check the code for controller runtime and there is not need to fix that one we just need to ignore it.

for the ginkgo one I think we can try to ask for a PR that should be easy I think

in general please use
https://docs.snyk.io/snyk-cli/commands/ignore#description
for the controller-runtime put non expiration as the code is expected
for ginkgo you can add 2 month or something like that and we can try to open a PR for them

liornoy: done

SchSeba (approved):
/lgtm
/approve

.snyk
ffromani: I agree with the general strategy, but I'm pretty convinced ginkgo _wants_ to emit a stacktrace in the logs in the reported case.

liornoy: @ffromani I can understand that. So do you suggest I'll reword the reason in ginkgo to "the code is expected." also?

ffromani: I don't think it's a blocking issue and we can revisit in 2 months

liornoy:
Failed to install the cluster. retesting.

SchSeba:

changes here are not related to the issues in the CI

PR #1712, opened 2023-11-27
cnf-tests: omit LatencyTestRun env variable

shajmakh:
Adapt https://github.com/openshift/cluster-node-tuning-operator/pull/857.

### DIFF
--- a/cnf-tests/entrypoint/test-run.sh
+++ b/cnf-tests/entrypoint/test-run.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-LATENCY_TEST_RUN="${LATENCY_TEST_RUN:-true}"
 DISCOVERY_MODE="${DISCOVERY_MODE:-true}"
 
 if [ "$IMAGE_REGISTRY" != "" ] && [[ "$IMAGE_REGISTRY" != */ ]]; then
@@ -8,4 +7,4 @@ if [ "$IMAGE_REGISTRY" != "" ] && [[ "$IMAGE_REGISTRY" != */ ]]; then
 fi
 
 echo running "/usr/bin/latency-e2e.test"
-DISCOVERY_MODE="$DISCOVERY_MODE" LATENCY_TEST_RUN="$LATENCY_TEST_RUN" "/usr/bin/latency-e2e.test" "$@"
+DISCOVERY_MODE="$DISCOVERY_MODE" "/usr/bin/latency-e2e.test" "$@"
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -19,9 +19,6 @@ export SCTPTEST_HAS_NON_CNF_WORKERS="${SCTPTEST_HAS_NON_CNF_WORKERS:-true}"
 # or need to run the tests again. In both cases the execution will be faster without deleting the profile.
 export CLEAN_PERFORMANCE_PROFILE="false"
 
-# Latency tests env variables
-export LATENCY_TEST_RUN=${LATENCY_TEST_RUN:-false}
-
 export IS_OPENSHIFT="${IS_OPENSHIFT:-true}"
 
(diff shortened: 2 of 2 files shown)


shajmakh:
/hold
for testing

liornoy:
thanks

shajmakh:
local testing passed successfully
/hold cancel
the OVN failure is known to be a test code bug and will be tackled in a later PR under the cluster-node-tuning-operator repo.

ffromani:
> local testing passed successfully /hold cancel the OVN failure is known to be a test code bug and will be tackled in a later PR under the cluster-node-tuning-operator repo.

please share the issue about this.
I do want to limit the overrides to be exceptional, not a (very bad) habit we pick up.

shajmakh:
> > local testing passed successfully /hold cancel the OVN failure is known to be a test code bug and will be tackled in a later PR under the cluster-node-tuning-operator repo.
> 
> please share the issue about this. I do want to limit the overrides to be exceptional, not a (very bad) habit we pick up.

sure, after investigation and reproduction attempts, it turned out that PAO functional tests configure fixed cpu values under the PP (according to 0_config). This is considered misconfiguration when the system cpus doesn't match the configured spec.cpu in PP, and there is no guarantee that the functionality will work properly with not all cpus reflected in the CPU section in the PP, and if the test passed with this configuration till now it can be considered a coincidence.

 It was reproduced when only part of the cpus are set to PP's spec.cpus, while the test passed successfully when the spec.cpu was fixed to reflect all IDs of the system (node) cpus. 
The fix should be done in the tests code, which is to modify the PP's spec.cpus to contain all system cpus. The card to track that is https://issues.redhat.com/browse/CNF-10472.

ffromani:
> > > local testing passed successfully /hold cancel the OVN failure is known to be a test code bug and will be tackled in a later PR under the cluster-node-tuning-operator repo.
> > 
> > 
> > please share the issue about this. I do want to limit the overrides to be exceptional, not a (very bad) habit we pick up.
> 
> sure, after investigation and reproduction attempts, it turned out that PAO functional tests configure fixed cpu values under the PP (according to 0_config). This is considered misconfiguration when the system cpus doesn't match the configured spec.cpu in PP, and there is no guarantee that the functionality will work properly with not all cpus reflected in the CPU section in the PP, and if the test passed with this configuration till now it can be considered a coincidence.
> 
> It was reproduced when only part of the cpus are set to PP's spec.cpus, while the test passed successfully when the spec.cpu was fixed to reflect all IDs of the system (node) cpus. The fix should be done in the tests code, which is to modify the PP's spec.cpus to contain all system cpus. The card to track that is https://issues.redhat.com/browse/CNF-10472.

I'm pleasently surprised to learn that "just" settings the perfprof correctly fixes this, whose connection wasn't obvious, and that it seems there are not obvious other issues past that.
considering we have the tracking issue, I'm ok overriding this

nevertheless, overriding must never become a habit so we will _still_ evaluate each and every override on a case-by-case basis. The correct way is to (us all) fix the flake, not keep overriding.

shajmakh:
> /hold
> 
> I'm about to approve (and override), but don't we need to pull in the new vendored dep before this PR or together with this PR?

good point! I missed that. glad that you raised this. Let me start another PR so we can keep this one will all the labels provided.

shajmakh:
depends on https://github.com/openshift-kni/cnf-features-deploy/pull/1717

shajmakh:
updating the dependencies (submodules) was done here: https://github.com/openshift-kni/cnf-features-deploy/pull/1719
will rebase and repush.

shajmakh:
Hi @zeeke, considering conversation on other PRs about that `ci/prow/e2e-aws-ci-tests` failures are known issues and the team is looking into that, is there any chance you can override the job?

PR #1721, opened 2023-12-05
cnf-tests: Remove NTO's 4_latency test suite

liornoy:
Remove the NTO's 4_latency test suite from the `TESTS_PATHS` variable, to prevent the functests from running this test suite.

We need that because we want to remove the "LATENCY_TEST_RUN" variable, which always skipped this suite during the conformance tests.

For clarity - the "LATENCY_TEST_RUN" was necessary when we compiled all of the suites together in the cnf-tests container image, but now it's redundant.


MarSik:
/approve

(I never remember which tense to use :)

PR #1722, opened 2023-12-05
OCPBUGS-24435: Fix for `VRF integration Integration: SRIOV, IPAM: static, Interfaces: 1, Scheme: 2 Pods 2 VRFs OCP Primary network overlap`

zeeke:
VRF negative tests may fail after deleting a pod, as happened in
the following case:

```
  STEP: Run client/server ICMP negative connectivity tests @ 12/05/23 16:49:55.92
  [FAILED] in [It] - /home/apanatto/dev/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/vrf/vrf.go:248 @ 12/05/23 16:50:02.622
  << Timeline

  [FAILED] Expected ping error, but output was: {ping: Warning: source address might be selected on device other than blue.
  PING 10.255.255.2 (10.255.255.2) from 10.255.255.1 blue: 56(84) bytes of data.
  64 bytes from 10.255.255.2: icmp_seq=1 ttl=64 time=0.058 ms

  --- 10.255.255.2 ping statistics ---
  5 packets transmitted, 1 received, 80% packet loss, time 4098ms
  rtt min/avg/max/mdev = 0.058/0.058/0.058/0.000 ms
   %!s(int=0) %!s(bytes.readOp=0)}
  Expected an error to have occurred.  Got:
      <nil>: nil
```

This PR adds an eventually block to the assertions around the `ping` command

### DIFF
--- a/cnf-tests/testsuites/e2esuite/vrf/vrf.go
+++ b/cnf-tests/testsuites/e2esuite/vrf/vrf.go
@@ -91,17 +91,26 @@ func pingIPViaVRF(cs *client.ClientSet, client *k8sv1.Pod, vrfName string, DestI
 }
 
 func assertIPIsReachableViaVRF(cs *client.ClientSet, client *k8sv1.Pod, vrfName string, DestIPAddr string) {
-	output, err := pingIPViaVRF(cs, client, vrfName, DestIPAddr)
+	EventuallyWithOffset(1, func(g Gomega) {
+		output, err := pingIPViaVRF(cs, client, vrfName, DestIPAddr)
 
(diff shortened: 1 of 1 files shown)


SchSeba:
nice work!

/lgtm
/approve
/override ci/prow/e2e-gcp-ovn

zeeke:
Failing tests:
```
[It] [sctp] Test Connectivity Connectivity between client ...
[It] [rfe_id:27363][performance] CPU Management when pod runs with the CPU load balancing runtime class [test_id:32646] should disable CPU load balancing for CPU's used by the pod
```
are not related to this PR

/override ci/prow/e2e-gcp-ovn

PR #1724, opened 2023-12-07
cnf-tests: Set DISCOVERY_MODE to true by default

liornoy:
Modify the 'test-run.sh' script, which serves as the entry point for the latency tests, and set the DISCOVERY_MODE variable to true. This adjustment eliminates the need for users to set it manually. However, users still have the option to override the variable and set it to 'false' if desired.

### DIFF
--- a/cnf-tests/entrypoint/test-run.sh
+++ b/cnf-tests/entrypoint/test-run.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
 LATENCY_TEST_RUN="${LATENCY_TEST_RUN:-true}"
+DISCOVERY_MODE="${DISCOVERY_MODE:-true}"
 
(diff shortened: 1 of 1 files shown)


liornoy:
> I believe now that cnf-tests image carry only latency tests, we can omit that variable from the root in NTO in the test binary, then bump NTO submodules only: https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/4_latency/latency.go#L76

The reason I opened the fix for it here, is to provide the short and easy fix, knowing that tomorrow is FF.
The removal from the root in NTO is less trivial and might need more time to think through, and requires more changes to the code nonetheless.

shajmakh:
thanks for this.  
/lgtm

PR #1725, opened 2023-12-07
cnf-tests: bump everything from 4.14 to 4.15

SchSeba:


ffromani:
/override ci/prow/e2e-gcp-ovn

we need to stop overriding and fix the test ASAP. This is likely the last time I override.
/cc @yanirq

PR #1726, opened 2023-12-11
ztp: Add cpuPartitioningMode to site-config-crd

leo8a:
Adds `cpuPartitioningMode` description to the siteConfig CRD, allowing users looking to write their siteconfigs to know the supported field name and the possible values that can be specified.

### DIFF
--- a/ztp/ran-crd/site-config-crd.yaml
+++ b/ztp/ran-crd/site-config-crd.yaml
@@ -787,6 +787,19 @@ spec:
                             items:
                               type: string
                             type: array   
+                      cpuPartitioningMode:
+                        type: string
+                        default: None
+                        description: CPUPartitioning determines if a cluster should be setup for
+                          CPU workload partitioning at install time. When this field is set the
+                          cluster will be flagged for CPU Partitioning allowing users to segregate
+                          workloads to specific CPU Sets. This does not make any decisions on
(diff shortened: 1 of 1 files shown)


serngawy:
Thank for adding this, You need to add the field into the node struct [here](https://github.com/serngawy/cnf-features-deploy/blob/largeSC/ztp/siteconfig-generator/siteConfig/siteConfig.go#L428) as well as the related CR corresponding field [here](https://github.com/serngawy/cnf-features-deploy/blob/largeSC/ztp/siteconfig-generator/siteConfig/clusterCRsV2.go)

leo8a:
I noticed that it was previously added under the Cluster struct, seems that it was only missed from the site-config-crd description

leo8a:
wondering if the site-config-crd could be populated automatically using `kubebuilder` or a similar tool. 

In this way every time a new struct / field is added to [siteConfig.go](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/siteconfig-generator/siteConfig/siteConfig.go), it gets automatically added to the CRD spec as well.

serngawy:
oh I see. Yes, I agree the CRD should be generated using kbuilder or codegenerator.

serngawy (approved):

PR #1734, opened 2023-12-22
cnf-tests: sriov-network-operator bump 4.16

zeeke:
SR-IOV Network Operator dependency bumped with commands:

```
go install -mod=mod github.com/openshift/sriov-network-operator@release-4.16
go mod edit -replace \
github.com/k8snetworkplumbingwg/sriov-network-operator=\
github.com/openshift/sriov-network-operator@release-4.16
go mod tidy
go mod vendor
```

cc  @gregkopels , @evgenLevin this bump includes the possibility to select the SR-IOV NIC to use during tests, via env var `SRIOV_NODE_AND_DEVICE_NAME_FILTER`

https://github.com/openshift-kni/cnf-features-deploy/pull/1734/files#diff-7ec8d476a373441296a790f93f698d05f15f2435454b44dbb1345249ab367bb7R113

Also bumped `k8s.io/*` deps to v0.28.3 to solve compiling error:
```
go vet -mod=vendor ./cnf-tests/testsuites/...
# k8s.io/client-go/applyconfigurations/meta/v1
vendor/k8s.io/client-go/applyconfigurations/meta/v1/unstructured.go:64:38: cannot use doc (variable of type *"github.com/google/gnostic/openapiv2".Document) as *"github.com/google/gnostic-models/openapiv2".Document value in argument to proto.NewOpenAPIData
```


zeeke:
update: waiting for
- https://github.com/openshift/sriov-network-operator/pull/880

before rebasing this

zeeke:
In order to bump sriov-network-operator depenendency, I had to:
- bump k8s dependencies to v0.28.3, as they are
- remove direct pin to `github.com/openshift/api` and `github.com/openshift/library-go`
- bump `cluster-nfd-operator` to release-4.16

ffromani:
>     * remove direct pin to `github.com/openshift/api` and `github.com/openshift/library-go`

not sure about these. And I see you seem to remove deps to machine-config-operator?
Are any of these obsolete (meaning no longer needed)?

fedepaol:
All in all, I'd say that if the deps do not require to be pinned, it's safe not to (and running go mod tidy should ensure all the deps are satisfied).

zeeke:
> > ```
> > * remove direct pin to `github.com/openshift/api` and `github.com/openshift/library-go`
> > ```
> 
> not sure about these. And I see you seem to remove deps to machine-config-operator? Are any of these obsolete (meaning no longer needed)?

You're right, I forgot to mention the machine-config-operator. The gist here is that it's difficult to pin all of those dependencies to a specific commit. 

`openshift/api`, `openshift/library-go` and `machine-config-operator` are transitive dependencies in cnf-tests and bumping all of them to the latest commit of release-4.16 may break the build (e.g. some of them can consume a new k8s incompatible dependency). Finding a good version of each transitive dependency is a Golang ecosystem duty, so I tend not to specify dependencies we do not directly need.

May it sound reasonable?

PR #1769, opened 2024-01-12
ocpbugs-26931: Fix wrong cast

daosman:


ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -362,11 +362,15 @@ func (scbuilder *SiteConfigBuilder) getClusterCR(clusterId int, siteConfigTemp S
 		case reflect.Slice:
 			sliceValues := make([]interface{}, 0)
 			for _, item := range v.([]interface{}) {
-				val, err := scbuilder.getClusterCR(clusterId, siteConfigTemp, item.(map[string]interface{}), nodeId)
-				if err != nil {
-					return mapIntf, err
+				if reflect.ValueOf(item).Kind() == reflect.String {
+					sliceValues = append(sliceValues, item)
+				} else {
```
imiller0, line 367: If the user has, in their template, a list which includes templated values, will this fail to replace the value, ie:
```
somelist:
- {{ .Node.somefield }}
```

daosman: Templates does not seem to work, if you include it in the over-rinding CR

serngawy: For list item template it needs to go through same check as [here](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go#L373). Not sure if it is a valid case usually template used to set specific field not an item in a list .

imiller0 (approved):

PR #1773, opened 2024-01-15
OCPBUGS-27031: Don't skip a NMStateConfig over-ride

daosman:
If the user ommits the NodeNetwork details from the siteConfig but includes it in an crTemplate over-ride of NMStateConfig, we end up not generating the NMStateConfig because we think the network is empty. Check if the user did not include an over-ride before skiping the NMStateConfig generation.

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/siteConfig.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfig.go
@@ -486,8 +486,11 @@ func (node *Nodes) CrAnnotationSearch(kind string, action string, cluster *Clust
 }
 
 // Return true if the NodeNetwork content is empty or not defined
-func (node *Nodes) nodeNetworkIsEmpty() bool {
-	if len(node.NodeNetwork.Config) == 0 && len(node.NodeNetwork.Interfaces) == 0 {
+func (node *Nodes) nodeNetworkIsEmpty(cluster *Clusters, site *Spec) bool {
+	// Check if we have an NMStateConfig as a crTemplate over-ride. If we do,
+	// we want to use the NodeNetwork details from there.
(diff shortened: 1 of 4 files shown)


sabbir-47:
@daosman Can we add a unit test here to verify that if Node.Network and interface is empty but NMStateConfig is provided via crTemplates, the resulting CRs list, NMStateConfig CR exists. Otherwise, the changes look good to me.

ztp/siteconfig-generator/siteConfig/siteConfig.go
```diff
@@ -486,8 +486,11 @@ func (node *Nodes) CrAnnotationSearch(kind string, action string, cluster *Clust
 }
 
 // Return true if the NodeNetwork content is empty or not defined
-func (node *Nodes) nodeNetworkIsEmpty() bool {
-	if len(node.NodeNetwork.Config) == 0 && len(node.NodeNetwork.Interfaces) == 0 {
+func (node *Nodes) nodeNetworkIsEmpty(cluster *Clusters, site *Spec) bool {
+	// Check if we have an NMStateConfig as a cfTemplate over-ride. If we do,
```
imiller0, line 490:
```suggestion
	// Check if we have an NMStateConfig as a crTemplate over-ride. If we do,
```

imiller0 (commented):
Can you also add unit tests that cover this issue. Thanks!

ztp/siteconfig-generator/siteConfig/siteConfig.go
daosman: Uploaded a uni test

imiller0 (approved):
/lgtm

PR #1776, opened 2024-01-16
cnf-tests: `sriov/numa` clean configuration

zeeke:
Add a step to clean the configuration after the `sriov/NUMA` test suite has been finished.
This is important because other test suites could expect to find a clean and stable cluster state.


SchSeba (approved):
/lgtm
/approve

PR #1785, opened 2024-01-18
ztp: pin policygenerator version to v1.12.4

liornoy:
Latest version of policygenerator breaks CI with the error message:

go: open-cluster-management.io/policy-generator-plugin/cmd/PolicyGenerator@latest (in open-cluster-management.io/policy-generator-plugin@v1.13.0):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
make: *** [Makefile:27: build] Error 1


sabbir-47:
moving from latest to pinned version, normally we shouldn't move away from latest. The further dependencies should be added to keep working with the latest version.

/lgtm

PR #1829, opened 2024-02-20
cnf-tests: Print podman logs before the check

liornoy:
This commit modifies the setup-build-index-image.sh script to print the logs of the Podman pod before checking if the index build succeeded. This improvement enhances debuggability because, currently, if the build fails, it does not display the logs, forcing us to consult the must-gather report.


liornoy:
@SchSeba Thanks for the suggestion!

PR #1836, opened 2024-02-26
ztp: OCPBUGS-25259: remove workaround for ice module in kdump

irinamihai:
Description:
- this removes the changes from PR #1381 that was a needed workaround at that time
- [OCPBUGS-25259](https://issues.redhat.com/browse/OCPBUGS-25259)


sabbir-47:
/uncc [simon3z](https://github.com/simon3z) [vladikr](https://github.com/vladikr)

bartwensley (commented):
Since this change is in the ZTP image running on the ACM, I have a couple questions:
1. Is it possible that the updated 4.15 ZTP image would be used to install an SNO that still needs this fix? I.e. an SNO running an earlier release than OCP 4.14.6 or 4.12.45? That would be an issue as the fix would be missing.
2. What about the case where the updated 4.15 ZTP image is applied to an ACM that is still managing nodes that were installed with the fix (e.g. OCP 4.14 or OCP 4.12)? Will the next policy sync on these nodes seamlessly remove the MC? Has that been tested?

sabbir-47:
> Since this change is in the ZTP image running on the ACM, I have a couple questions:
> 
> 1. Is it possible that the updated 4.15 ZTP image would be used to install an SNO that still needs this fix? I.e. an SNO running an earlier release than OCP 4.14.6 or 4.12.45? That would be an issue as the fix would be missing.

when a 4.15 ZTP image is used to install a SNO, it will be a fresh installation based on OCP 4.15 image which i think already have the updated kernel version, unless otherwise user mentions [clusterImageSetNameRef]([http://registry.kni-qe-0.lab.eng.rdu2.redhat.com:3000/kni-qe/ztp-site-configs/src/cnfde23-4.15/siteconfig/cnfde23.yaml#L10) to a older version.

I think we need to verify whether we can upgrade from OCP 4.14.6 or 4.12.45 to 4.15 without having this workaround and having no issues.

> 2. What about the case where the updated 4.15 ZTP image is applied to an ACM that is still managing nodes that were installed with the fix (e.g. OCP 4.14 or OCP 4.12)? Will the next policy sync on these nodes seamlessly remove the MC? Has that been tested?

The extra-manifests are directly applied to the cluster at installation time and not monitored via acm policies.

donpenney:
> > Since this change is in the ZTP image running on the ACM, I have a couple questions:
> > 
> > 1. Is it possible that the updated 4.15 ZTP image would be used to install an SNO that still needs this fix? I.e. an SNO running an earlier release than OCP 4.14.6 or 4.12.45? That would be an issue as the fix would be missing.
> 
> when a 4.15 ZTP image is used to install a SNO, it will be a fresh installation based on OCP 4.15 image which i think already have the updated kernel version, unless otherwise user mentions clusterImageSetNameRef to a older version.
> 

But this is the single ZTP gitops image running on the hub that is used to install cluster, regardless of version being installed on that cluster, right? ie. this 4.15 ZTP gitops image could be used to install a cluster with 4.14.4?

sabbir-47:
> > > Since this change is in the ZTP image running on the ACM, I have a couple questions:
> > > 
> > > 1. Is it possible that the updated 4.15 ZTP image would be used to install an SNO that still needs this fix? I.e. an SNO running an earlier release than OCP 4.14.6 or 4.12.45? That would be an issue as the fix would be missing.
> > 
> > 
> > when a 4.15 ZTP image is used to install a SNO, it will be a fresh installation based on OCP 4.15 image which i think already have the updated kernel version, unless otherwise user mentions clusterImageSetNameRef to a older version.
> 
> But this is the single ZTP gitops image running on the hub that is used to install cluster, regardless of version being installed on that cluster, right? ie. this 4.15 ZTP gitops image could be used to install a cluster with 4.14.4?

A customer must not use 4.15 ztp image to install 4.14.4, they should remain in the same z-stream.

irinamihai:
> > > > Since this change is in the ZTP image running on the ACM, I have a couple questions:
> > > > 
> > > > 1. Is it possible that the updated 4.15 ZTP image would be used to install an SNO that still needs this fix? I.e. an SNO running an earlier release than OCP 4.14.6 or 4.12.45? That would be an issue as the fix would be missing.
> > > 
> > > 
> > > when a 4.15 ZTP image is used to install a SNO, it will be a fresh installation based on OCP 4.15 image which i think already have the updated kernel version, unless otherwise user mentions clusterImageSetNameRef to a older version.
> > 
> > 
> > But this is the single ZTP gitops image running on the hub that is used to install cluster, regardless of version being installed on that cluster, right? ie. this 4.15 ZTP gitops image could be used to install a cluster with 4.14.4?
> 
> A customer must not use 4.15 ztp image to install 4.14.4, they should remain in the same z-stream.

I though the ZTP plugin of release N can be used to deploy and manage clusters with N, N-1 and N-2. Does this only cover Y releases and it does not account for z-streams? @sabbir-47, kindly explain.

sabbir-47:
> > > > > Since this change is in the ZTP image running on the ACM, I have a couple questions:
> > > > > 
> > > > > 1. Is it possible that the updated 4.15 ZTP image would be used to install an SNO that still needs this fix? I.e. an SNO running an earlier release than OCP 4.14.6 or 4.12.45? That would be an issue as the fix would be missing.
> > > > 
> > > > 
> > > > when a 4.15 ZTP image is used to install a SNO, it will be a fresh installation based on OCP 4.15 image which i think already have the updated kernel version, unless otherwise user mentions clusterImageSetNameRef to a older version.
> > > 
> > > 
> > > But this is the single ZTP gitops image running on the hub that is used to install cluster, regardless of version being installed on that cluster, right? ie. this 4.15 ZTP gitops image could be used to install a cluster with 4.14.4?
> > 
> > 
> > A customer must not use 4.15 ztp image to install 4.14.4, they should remain in the same z-stream.
> 
> I though the ZTP plugin of release N can be used to deploy and manage clusters with N, N-1 and N-2. Does this only cover Y releases and it does not account for z-streams? @sabbir-47, kindly explain.


ZTP plugin can deploy and manage N...N-2 release with their respective reference CRs(source-crs and extra manifests)

irinamihai:
Just wanted to put together the concerns raised above.

> Is it possible that the updated 4.15 ZTP image would be used to install an SNO that still needs this fix? I.e. an SNO running an earlier release than OCP 4.14.6 or 4.12.45? That would be an issue as the fix would be missing.

To add to Sabbir's point, if we use the 4.15 ZTP image to install let's say 4.13, then the 2 ways to do that are:

Option 1 - using the version independence model
 - create a version_4.13 directory under both siteconfig and policygentemplate
 - Use the 4.13 ZTP image to extract the 4.13 extra-manifests and source-crs and put them under their respective locations in Git - the 4.13 manifests would have the workaround, as needed.
 - set clusterImageSet to the desired 4.13 image 
 - set the correct binding rules for PGT
 
Options 2 - using the 4.13 ZTP image
  - update the initContainer of the openshift-gitops ArgoCD to ztp-site-generator:4.13
  - this time, the extra-manifests will be those from the initContainer which again, contain the needed workaround

> What about the case where the updated 4.15 ZTP image is applied to an ACM that is still managing nodes that were installed with the fix (e.g. OCP 4.14 or OCP 4.12)? Will the next policy sync on these nodes seamlessly remove the MC? Has that been tested?

As Sabbir said, the MCs are created and applied as part of the initial installation and are not monitored/updated through ACM policies.

> But this is the single ZTP gitops image running on the hub that is used to install cluster, regardless of version being installed on that cluster, right? ie. this 4.15 ZTP gitops image could be used to install a cluster with 4.14.4?

The answer from question 1 also applies here.
 
> What happens when we upgrade from an OCP version with the workaround to an OCP version without the workaround.

From everything that I could gather, the MCPs resulted from the ZTP extra-manifests will stay in place. From Bart's initial comment in his PR, this should be fine: "When this MachineConfig is removed, the changes to the kdump config file will not be undone. This is OK because the ice module is not used when doing a kdump."

@bartwensley @sabbir-47 @donpenney kindly let me know what your opinion on this is and if you consider anything further is needed to have this revert merged. Thank you.

bartwensley:
Thanks @irinamihai for addressing all the concerns that were raised - I think we should be OK to merge this.
/lgtm

PR #1845, opened 2024-03-07
ztp: Add IPSec N-S configuration

sabinaaledort:


yliu127:
Can we pls add usage examples/recommended configs in https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example ?

sabinaaledort:
> Can we pls add usage examples/recommended configs in https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd/example ?

Since this is an optional feature the files are stored in the optional directory. There is an example in the README so I'm not sure what config example you are missing, can point to one?

yliu127:
@sabinaaledort this is a small example of PGT for optional feature: https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml#L49-L54 
It is usually commented due to optional.

The example for PGT is needed to show user how to use it under the context of ZTP.

ztp/source-crs/optional-extra-manifest/ipsec/build.sh
```diff
@@ -0,0 +1,75 @@
...
+      mode: 0400
+      overwrite: true
+      contents:
+        local: ca.pem
+    - path: /etc/pki/certs/left_server.p12
+      mode: 0400
+      overwrite: true
+      contents:
+        local: left_server.p12
+    - path: /usr/local/bin/ipsec-endpoint-config.yml
```
yuvalk, line 55: config file in /usr/local/bin ?
let's move it somewhere else..

ztp/source-crs/optional-extra-manifest/ipsec/ipsec-endpoint-config.yml
lack, line 1: Is there a reason this is .yml and not .yaml?

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -184,3 +184,19 @@ spec:
 #                user:
 #                  name: root
 #     ---- sources needed for updating CRI-O workload-partitioning ends here ----
+
+    # ---- Sources needed to configure IPsec encryption for external traffic ----
+    # 1. Set the `$mode` in the `enable-ipsec.yaml`
```
yuvalk, line 189: if we want to be strict, can hard code this to `External`
and then add `Full` option once we discover there an actual telco use case for it

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -184,3 +184,19 @@ spec:
...
+    # 2. Provide the following certificate files:
+    #    * left_server.p12
+    #    * ca.pem
+    # 3. Set the following values in the `ipsec-endpoint-config.yml`:
+    #    * `$clusterNode` - the host name or IP of the cluster node for the IPsec tunnel on the cluster side
```
yuvalk, line 194: since this is for SNO
can be just `$node` 
note that `left` should be reachable/resolvable
so it's the IP address or DNS hostname (not host name)

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -184,3 +184,19 @@ spec:
...
+    #    * `$externalHost` - the external host name or IP
+    #    * `$externalAddress` - the IP address or subnet of the external host
```
yuvalk, line 196: I think we should support multiple subnets

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -184,3 +184,19 @@ spec:
...
 #     ---- sources needed for updating CRI-O workload-partitioning ends here ----
+
+    # ---- Sources needed to configure IPsec encryption for external traffic ----
+    # 1. Set the `$mode` in the `enable-ipsec.yaml`
+    # 2. Provide the following certificate files:
+    #    * left_server.p12
+    #    * ca.pem
+    # 3. Set the following values in the `ipsec-endpoint-config.yml`:
+    #    * `$clusterNode` - the host name or IP of the cluster node for the IPsec tunnel on the cluster side
+    #    * `$externalHost` - the external host name or IP
```
yuvalk, line 195: same here, name need to be resolvable (ie dns)

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
sabinaaledort: I was thinking the source-crs can be generic for both MNO and SNO, so maybe keep it `$clusterNode`

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
sabinaaledort: For that you need `rightsubnets` field which is currently not supported in nmstate

yuvalk:
/lgtm
@lack can you approve?

lack (approved):

ztp/source-crs/optional-extra-manifest/ipsec/build.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
yliu127, line 1: ztp won't execute this sh asaik. @lack @sabbir-47 could you confirm? 

So the MC yaml should be added to source-cr directly instead.

ztp/source-crs/optional-extra-manifest/ipsec/enable-ipsec.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: operator.openshift.io/v1
```
yliu127, line 1: should these configs be applied at install time or config time? 

extra-manifest indicates install time - at least this is the convention for other files under extra-manifests

ztp/source-crs/optional-extra-manifest/ipsec/ipsec-endpoint-config.yml
```diff
@@ -0,0 +1,15 @@
+interfaces:
```
yliu127, line 1: is this for day2 config? If so, it should stay in a different folder than day0 config so that for day0, user can include a folder instead of individual files.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -184,3 +184,18 @@ spec:
...
+    # ---- Sources needed to configure IPsec encryption for external traffic ----
+    # 1. Provide the following certificate files:
+    #    * left_server.p12
+    #    * ca.pem
+    # 2. Set the following values in the `ipsec-endpoint-config.yml`:
+    #    * `$clusterNode` - the node IP address or DNS hostname
+    #    * `$externalHost` - the external host IP address or DNS hostname
+    #    * `$externalAddress` - the IP address or subnet of the external host
+    # 3. Run the `build.sh` script to create the machine config
+    # - fileName: enable-ipsec.yaml
```
yliu127, line 197: This file path does not align with what we added to the source-crs, so the deployment will fail due to file does not exist. 

I think this the file path ztp can recognize, however, usually "extra-manfest" gets applied at install time, so there are some inconsistencies here as we are trying too apply these at day2 config time.

`optional-extra-manifest/ipsec/ipsec-endpoint-config.yml`

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
```diff
@@ -184,3 +184,18 @@ spec:
...
+    #   policyName: "config-policy"
+    # - fileName: 99-ipsec-master-endpoint-config.yaml
```
yliu127, line 199: This file does not exist unless MC yaml is added directly to ztp/source-cr dir.

ztp/source-crs/optional-extra-manifest/ipsec/ipsec-endpoint-config.yml
yliu127: nit: should be .yaml to be consistent in the project.

ztp/source-crs/optional-extra-manifest/ipsec/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+* Enable IPsec encryption yaml
+* IPsec endpoint configuration with NMState yml
+* A MachineConfig to import external certs into IPsec NSS and apply the IPsec configuration
+
+Also available in OpenShift docs: [IPsec encryption for external traffic](https://docs.openshift.com/container-platform/4.15/networking/ovn_kubernetes_network_provider/configuring-ipsec-ovn.html#nw-ovn-ipsec-external_configuring-ipsec-ovn)
+
+### Prerequisites
+
+* `butane` utility installed
```
yliu127, line 13: There should be no pre-req for ZTP. 

ZTP gets applied to new/uninstall server, and there should be no user interruption in the middle of the ZTP until day2 configs are completed.

ztp/source-crs/optional-extra-manifest/ipsec/build.sh
yliu127: Maybe the build.sh can go here if we really want to create MC files using script. https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/extra-manifests-builder

sabinaaledort: The script generates a MC with the certificates to import in base64, since the certificates are changing between our partners and our setups it can't be a static MC.
Please follow the README and the official documentation https://docs.openshift.com/container-platform/4.15/networking/ovn_kubernetes_network_provider/configuring-ipsec-ovn.html#nw-ovn-ipsec-external_configuring-ipsec-ovn 
You should run it as a pre step to generate the MC that will be applied with ZTP. You can run that in your automation as a pre step to ZTP.

ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml
sabinaaledort: Since the script generates a MC you can choose where to store it, please note this feature requires a pre step and also an optional feature that's why it is stored in optional dir.
Please follow the README and the official documentation https://docs.openshift.com/container-platform/4.15/networking/ovn_kubernetes_network_provider/configuring-ipsec-ovn.html#nw-ovn-ipsec-external_configuring-ipsec-ovn

sabbir-47:
@sabinaaledort Was this tested before creating a PR in a ZTP setting? I would expect the PR description contains the changes/addition and the impact of the content.

ztp/source-crs/optional-extra-manifest/ipsec/README.md
sabinaaledort: Please note this is an optional feature that requires a pre step, you can automate that as pre step to ZTP or generate the MC once with your certificates and store it for ZTP. 
Please see the official documentation https://docs.openshift.com/container-platform/4.15/networking/ovn_kubernetes_network_provider/configuring-ipsec-ovn.html#nw-ovn-ipsec-external_configuring-ipsec-ovn

ztp/source-crs/optional-extra-manifest/ipsec/build.sh
sabbir-47: @sabinaaledort The ipsec folder should move to extra-manifests-builder: https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/extra-manifests-builder. The scripts should be as much automatic as possible and the static part should be highlighted.

ztp/source-crs/optional-extra-manifest/ipsec/ipsec-endpoint-config.yml
sabbir-47: @sabinaaledort a day-2 config can't stay in day-0 manifests folder as Yang mentioned.

sabbir-47:
@sabinaaledort @yuvalk This PR needs to be reverted. It doesn't follow any convention on how day-0 and day-2 configurations are organized in ztp container. Please consider the below suggestions:

1. The ipsec folder must move under [extra-manifests-builder](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/extra-manifests-builder) folder following the examples provided there.
2. A proper PR description of changes since it is a new feature and not some minor changes in a specific CR. Please include:  what this PR is about, when customer needs to use it and the impact, should it be a day-0 or day-2 change., etc.
  -  Day-0: changes that goes in the ignition during installation. 
  -  Day-2: any configuration changes made via acm policies after a successful installation.
3. A test plan on how the QE is expected to test it.

/cc @lack @imiller0 @yliu127

ztp/source-crs/optional-extra-manifest/ipsec/build.sh
sabinaaledort: The script generates a MC with the certificates to import in base64, since the certificates are changing between our partners and our setups it can't be a static MC.
The parts to change are described in the README and the official documentation https://docs.openshift.com/container-platform/4.15/networking/ovn_kubernetes_network_provider/configuring-ipsec-ovn.html#nw-ovn-ipsec-external_configuring-ipsec-ovn
This is a pre step to do once or as an automation in the pipeline before ZTP.
This is the path we agreed to store the files since this is an optional feature. If that's the day 0 path we can move it to the path you suggested.

sabinaaledort:
1. We can change the path to the day2 optional features, that shouldn't be a problem
2. I updated the description, please also see the Jira https://issues.redhat.com/browse/CNF-9053
3. The QE test plan is in [https://polarion.engineering.redhat.com/polarion/#/project/OSE/wiki/CNF/IP-SEC%20N-S%20Test%20Plan](https://polarion.engineering.redhat.com/polarion/#/project/OSE/wiki/CNF/IP-SEC%20N-S%20Test%20Plan%20_4_)

ztp/source-crs/optional-extra-manifest/ipsec/build.sh
bradyallenjohnson: Something to keep in mind here:
In customer/partner deployments, there will be a unique MC generated per target far-edge SNO deployment. This is because a unique IPSec certificate is needed per SNO cluster and the ipsec-endpoint-config.yml file will have a unique SNO IP per SNO cluster.
Also, the IPSec config and cert should be applied as a Day2 operation. This is because usually these SNO clusters will first be deployed with OCP, etc (many times in a factory environment) without IPSec, then post-deployment the IPSec tunnel will be configured and started.

yliu127: @sabinaaledort everything in source-crs folder is meant to be consumable as part of the ZTP process directly regardless of core features or optional features.

The "extra-manifests" subfolders are for day0 config, and the rest are for day2 configs. User should be able to use/reference them from user git repo in siteconfig or in PGT.
 
For example BMER is an optional feature, but all the BMER related source-crs stay in this folder because its user's choice on whether to use/reference them in day2 PGT at the end.

If the solution requires prereq steps prior to ZTP, it needs to be documented separately. So far, I've seen most of the readmes go into this folder: https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd. Maybe @sabbir-47 can suggest something better if any. 

@sabbir-47 , it seems the sh requires 2 external cert files for each site as @bradyallenjohnson mentioned, I guess in this case, there is no easy way to integrate this script to extra-manifests-builder in this case?

ztp/source-crs/optional-extra-manifest/ipsec/ipsec-endpoint-config.yml
lack: As a counterpoint, this is not enabled by default, so in my opinion the question of *where* it lives inside the ZTP container is less important compared to the facts that:
- It is not enabled by default
- We provide good instructions to customers on how to enable it

There are other places it could go; one would be under `ztp/gitops-subscriptions/argocd/examples/`.  But keep in mind that the `source-crs/` hierarchy already has a combination of day-0, day-2, and optional content that can be enabled at either phase by a customer opt-in adding the yamls to the right extra-manifests or PGT as needed.

lack:
As a counterpoint, this is not enabled by default, so in my opinion the question of *where* it lives inside the ZTP container is less important compared to the facts that:
- It is not enabled by default
- We provide good instructions to customers on how to enable it

There are other places it could go; one would be under `ztp/gitops-subscriptions/argocd/examples/`.  But keep in mind that the `source-crs/` hierarchy already has a combination of day-0, day-2, and optional content that can be enabled at either phase by a customer opt-in adding the yamls to the right extra-manifests or PGT as needed.

ztp/source-crs/optional-extra-manifest/ipsec/build.sh
sabbir-47: I think, the CRs which are directly consumable should go under source-crs directory, in this case it is not.

> If the solution requires prereq steps prior to ZTP, it needs to be documented separately. So far, I've seen most of the readmes go into this folder: https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd. Maybe @sabbir-47 can suggest something better if any.

Readme must go there.

> @sabbir-47 , it seems the sh requires 2 external cert files for each site as @bradyallenjohnson mentioned, I guess there is no easy way to integrate this script to extra-manifests-builder in this case?

We don't provide a way to integrate scripts directly into ztp, apart from the ones that can be build using exta-manifests-builder which doesn't depend on the user input.

lack: > I think, the CRs which are directly consumable should go under source-crs directory, in this case it is not.

Keep in mind also that many of the source-crs already present are NOT directly applicable to user systems; many of them must have at least some parts overridden in PGTs.

> > If the solution requires prereq steps prior to ZTP, it needs to be documented separately. So far, I've seen most of the readmes go into this folder: https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd. Maybe @sabbir-47 can suggest something better if any.
> 
> Readme must go there.

I disagree about the readme.  This readme is specific to this set of yamls and the related build.sh - I think it makes more sense for them all to be together in their own subdirectory.

> > @sabbir-47 , it seems the sh requires 2 external cert files for each site as @bradyallenjohnson mentioned, I guess there is no easy way to integrate this script to extra-manifests-builder in this case?
> 
> We don't provide a way to integrate scripts directly into ztp, apart from the ones that can be build using exta-manifests-builder which doesn't depend on the user input.

I agree that this cannot and should not go in extra-manafifests-builder.  That is a source-control of mechanism to build static MachineConfigs that end up in source-crs/extra-manifests.  That doesn't fit this use-case, as these are not static yaml files.

lack: > The build.sh script should not be included in ztp. It should only be used as a pre-step by the partner/customer to generate the MC.

The build.sh script is a critical part of the customer-facing solution.  It simplifies creation of yaml files that are essential to customers implementing this feature.  I think that is exactly why it belongs in the ZTP container.

bradyallenjohnson: Ok, perhaps this is a question of nomenclature. The customers will not be executing this script on the SNO cluster, but from a staging env where they launch the SNO deployments. When I say it should not be in ZTP, I mean ZTP should not install the script on the SNO cluster.

lack: > ZTP should not install the script on the SNO cluster.

Agreed!

lack: So it seems to me like we have 2 viable options here, armed with the knowledge that these files are not going to be installed directly on an SNO by default:
- Keep these files where they are under source-crs/optional-extra-manifests/ipsec
- Move the files into the "examples" subdirectory alongside the example siteconfig and example PGTs we already ship in the ZTP container

sabbir-47: @lack 
> Keep in mind also that many of the source-crs already present are NOT directly applicable to user systems; many of them must have at least some parts overridden in PGTs.

Yes, we can override in some of the provided source-crs, e.g. performanceProfile, tuned, sriov etc. But in this case, can a user override some parameter and consume the yaml as it is? I don't think so.

I would be up for moving the ipsec under the examples directory documenting how a user is supposed to use it and the steps.

sabbir-47:
The idea is to make everything as coherent as possible. If we look at contents under source-crs, some of the CRs needs to be overriden by customer based on their needs , e.g. performanceProfile, tuned, sriov etc. But in this case, can a user override some parameter and consume the yaml as it is? I don't think so.

I would be up for moving the ipsec under the examples directory documenting how a user is supposed to use it and the steps.

>But keep in mind that the source-crs/ hierarchy already has a combination of day-0, day-2, and optional content that can be enabled at either phase by a customer opt-in adding the yamls to the right extra-manifests or PGT as needed.

Although i agree on some part, customer depends on our recommendation on any new change or a new feature that is introduced. The reference configs that we provide acts like a blueprint.

yliu127:
@sabbir-47 QE was not able to follow the steps in the PR to enable this feature via ZTP. The whole point for integrating this to ZTP is to provide an exact workflow that customer can follow. The ZTP container is not only a readme, but more importantly an auto yaml/policy generator. The files in different folders are used for different purposes.

- files directly under source-crs dir are for day2 configs, and consumer can decide which yamls they want to reference
- examples dir is a place to guide customer how they can reference these yamls with or without custom configs. 
- extra-manifests under source-crs are for day0 configs during ocp install time, all of those are meant to be automatically applied by default.

In generally, things in source-crs are  meant to  be consumable directly in the ZTP workflow. 

So this definitely needs  to be reorganized with clear steps on the pre-ZTP setups, and how can user consume the MC and the ipsec enable yaml **during ZTP**.

yuvalk:
@yliu127 If you cant manage to set it up, please ask for help
there are multiple ppl within ENG who succeeded ..

last I saw is a miss configuration of
https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/source-crs/optional-extra-manifest/ipsec/enable-ipsec.yaml#L9
on your cluster (set to `Full` for some unknown reason)

but really I think the PR is not the right place for this discussion

sabbir-47:
@yuvalk So if QE is having issues testing this feature following the instructions and need ENG help, will ENG also help the customers? We are all talking about the process where things must land properly.

Has someone from your team tested this feature with a full ZTP flow?

yuvalk:
I expect CX to be able to help with issues such as the above.
(someone setting the wrong ipsec mode)

sabbir-47:
For example this yaml file https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/source-crs/optional-extra-manifest/ipsec/enable-ipsec.yaml shouldn't exist in the first place. Because when we are targeting a CR with same api, kind, and name with multiple object defintions, only one of the object definition is applied and gets compliant via ACM and not the both. The spec section should've gone under https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/source-crs/DisableSnoNetworkDiag.yaml#L7 as a commented line specifying, if ipsec feature needs to be enabled, these lines must be uncommented.

yuvalk:
that's a valid comment, but maybe better to open as an issue rather than on a merged PR.

IDK what the team thinks about it
@lack  ?

BTW: also the whole discussion about how optional/extra manifests should work is a valid discussion, just not on a PR. it's wider issue than ipsec..

yliu127:
@yuvalk ZTP is not about adding a readme with what yamls customer needs to apply (otherwise the existing OCP doc would have been sufficient), but rather "how" they can be applied/integrated via ZTP as an end to end solution. 

If QE is not able to do the config by following the example in this repo, which is part of the product shipped to customer, then customer would not be able to do that either.  

I will open bug.

lack:
I have a lot of thoughts about this, but I agree with @yuvalk that this merged PR is not the right place for the discussion!

@yliu127 Please CC me on the bug when you file it, thanks!

PR #1852, opened 2024-03-25
ztp: OCPBUGS-31325: Backport fix to disable rcu_expedited after node recovers from reboot

fontivan:
ztp: OCPBUGS-31325: Backport fix to disable rcu_expedited after node recovers from reboot

- Update accelerated-container-startup.sh to also disable rcu_expedited after reboot recovery
- Update README with details on that change
- Update generated manifests for those changes


ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
```diff
@@ -205,6 +210,7 @@ main() {
   logger "Recovery: Recovery Mode Starting"
   setUnrestricted
   waitForReady
+  setRcuNormal
```
bartwensley, line 213: I think we want to do this in all cases - so probably need some restructuring to allow this to run even if there is no unrestricted/restricted cpuset.

fontivan:
cc. @daosman @donpenney @browsell

ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
fontivan: I have a proposed fix to conditionally perform the set unrestricted but to delay the potential error return until the node has recovered and rcu can be set to normal, but I need to investigate the implications of delaying that error return
```
diff --git a/ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh b/ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
index 54f76f36..976efd36 100644
--- a/ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
+++ b/ztp/extra-manifests-builder/04-accelerated-container-startup/accelerated-container-startup.sh
@@ -192,25 +192,46 @@ setRcuNormal() {
 }
 
 main() {
+  # We will use a local variable to store the state of whether we need to perform this action
+  local do_set_unrestricted
+  do_set_unrestricted=0
+
+  # If unrestricted cpuset was not detected then we cannot set unrestricted
   if ! unrestrictedCpuset >&/dev/null; then
     logger "Recovery: No unrestricted Cpuset could be detected"
-    return 1
+    do_set_unrestricted=1
   fi
 
+  # If restricted cpuset is not configured then we do not need to set unrestricted
   if ! restrictedCpuset >&/dev/null; then
     logger "Recovery: No restricted Cpuset has been configured.  We are already running unrestricted."
-    return 0
+    do_set_unrestricted=2
   fi
 
-  # Ensure we reset the CPU affinity when we exit this script for any reason
-  # This way either after the timer expires or after the process is interrupted
-  # via ^C or SIGTERM, we return things back to the way they should be.
-  trap setRestricted EXIT
+  # If we need to change unrestricted then we will do that now
+  if [[ ${do_set_unrestricted} -eq 0 ]]; then
+    # Ensure we reset the CPU affinity when we exit this script for any reason
+    # This way either after the timer expires or after the process is interrupted
+    # via ^C or SIGTERM, we return things back to the way they should be.
+    trap setRestricted EXIT
+
+    logger "Recovery: Recovery Mode Starting"
+    setUnrestricted
+  fi
 
-  logger "Recovery: Recovery Mode Starting"
-  setUnrestricted
+  # We will always wait for ready here
   waitForReady
+
+  # We will always set rcu normal here
   setRcuNormal
+
+  # Return an error code if unstricted was not detected
+  if [[ ${do_set_unrestricted} -eq 1 ]]; then
+    return 1
+  fi
+
+  # Return 0 by default
+  return 0
 }
```

fontivan: I discussed with @daosman and it seems like nobody is waiting for this script to exist, so it should be safe to delay the error return

bartwensley (approved):
Looks good

fontivan:
@daosman  @bartwensley  @donpenney @browsell testing has completed and I am very happy with the results, please review at your leisure

fontivan:
Still need a /approve from a ZTP-Approver
cc. @ijolliffe, @serngawy, @imiller0, @vitus133, @nishant-parekh, @lack, @Missxiaoguo

PR #1853, opened 2024-03-26
ztp: ptp add ptpconfig for High Availability

aneeshkp:
4.16 : ptp operator enables HA via ptpconfig.
This PR adds config required to enable HA

### DIFF
--- a/ztp/source-crs/PtpConfigForHA.yaml
+++ b/ztp/source-crs/PtpConfigForHA.yaml
@@ -1,0 +1,23 @@
+apiVersion: ptp.openshift.io/v1
+kind: PtpConfig
+metadata:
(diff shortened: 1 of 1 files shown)


hhassid:
lgtm

ztp/source-crs/PtpConfigForHA.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: ptp.openshift.io/v1
```
sabbir-47, line 1: @aneeshkp There are other configs available in the source-crs for HA. Would you mind adding a comment at the beginning on when and why user needs to use it?

nishant-parekh (approved):
/lgtm

PR #1862, opened 2024-04-05
Retry BeforeAll when initial setup fails

mlguerrero12:
When the BeforeEach inside BeforeAll fails with a Ginkgo exception, subsequent tests (or retries of the first test) fail because the initial setup (code inside this BeforeEach) has not being executed. This change forces the BeforeEach to run again in case of failure.

### DIFF
--- a/cnf-tests/testsuites/pkg/execute/ginkgo.go
+++ b/cnf-tests/testsuites/pkg/execute/ginkgo.go
@@ -8,8 +8,8 @@ func BeforeAll(fn func()) {
 	first := true
 	ginkgo.BeforeEach(func() {
 		if first {
-			first = false
 			fn()
+			first = false
 		}
 	})
 }


zeeke:
I'm not 100% sure every BeforeAll is designed to be run multiple times. (e.g. https://github.com/openshift-kni/cnf-features-deploy/blob/a4cc8a810e13bdf99280644c96abb5c1887472c6/cnf-tests/testsuites/e2esuite/vrf/vrf.go#L71 )

But I think it would improve the overall stability of the test suite. LGMT

Waiting others (@SchSeba, @ffromani) opinion

/test e2e-telco5g-cnftests

mlguerrero12:
@zeeke, @cgoncalves, could you please review this? Thank!

SchSeba:
Yep I agree with you on that

maybe if the beforeAll failed we just failed all the tests that called this BeforeAll?

another point I think after the bump to ginkgo v2 there is a BeforeAll in the ginkgo package no?

zeeke:
> another point I think after the bump to ginkgo v2 there is a BeforeAll in the ginkgo package no?

Yes, there is. But it can be used in Ordered containers only, and when a BeforeAll section fails, it marks one test as failed and all the others are skipped.

As I don't see any 100% clean fix, I would go with the changes Marcelo is proposing

mlguerrero12:
These sriov failures could have been avoided with this change

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.16-e2e-telco5g-cnftests/1776067470998638592/artifacts/e2e-telco5g-cnftests/telco5g-cnf-tests/artifacts/test_results.html

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.14-e2e-telco5g-cnftests/1777275442659594240/artifacts/e2e-telco5g-cnftests/telco5g-cnf-tests/artifacts/test_results.html

Nothing wrong to retry the BeforeAll if the initial setup fails.

SchSeba:
sounds good

/lgtm
/approve

PR #1872, opened 2024-04-16
cnf-tests: Delete macvlan nad to retry allowlist test

mlguerrero12:
In case of failure, subsequent retries of the allowlist test fail because the macvlan is already present. This changes ensure that the macvlan nad is deleted.

### DIFF
--- a/cnf-tests/testsuites/e2esuite/security/tuning.go
+++ b/cnf-tests/testsuites/e2esuite/security/tuning.go
@@ -151,10 +151,6 @@ var _ = Describe("[tuningcni]", func() {
 			Expect(ok).To(BeTrue())
 		})
 
-		AfterEach(func() {
-			updateAllowlistConfig(originalSysctls)
-		})
-
 		It("should start a pod with custom sysctl only after adding sysctl to allowlist", func() {
 			macvlanNadName := "macvlan-nad1"
 			sysctl := "net.ipv4.conf.IFNAME.accept_local"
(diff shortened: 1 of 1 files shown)


cnf-tests/testsuites/e2esuite/security/tuning.go
```diff
@@ -152,16 +160,19 @@ var _ = Describe("[tuningcni]", func() {
 		})
 
 		AfterEach(func() {
-			updateAllowlistConfig(originalSysctls)
+			err := client.Client.Delete(context.Background(), macVlandNad)
```
zeeke, line 163: Does it panic if `macVlandNad` is nil? What about using `DeferCleanup(...)` after line 176?

mlguerrero12: I'm not sure if it panics with nil. Nevertheless, thanks for the suggestion. I'll use `DeferCleanup`.

cnf-tests/testsuites/e2esuite/security/tuning.go
```diff
@@ -166,6 +162,8 @@ var _ = Describe("[tuningcni]", func() {
 			err = client.Client.Create(context.Background(), macVlandNad)
 			Expect(err).ToNot(HaveOccurred())
 
+			DeferCleanup(client.Client.Delete, context.Background(), macVlandNad)
```
zeeke, line 165: nice!!

SchSeba:
failed test is not related
/lgtm
/approve

PR #1873, opened 2024-04-17
ztp: add disableDrain true in ztp

SchSeba:


vitus133:
Hi @SchSeba ! Could you please also update this example with the new file name:
https://github.com/openshift-kni/cnf-features-deploy/blob/4e6537dfeeea721a9d865e3945c5a651e04d2ef2/ztp/gitops-subscriptions/argocd/example/policygentemplates/group-du-sno-ranGen.yaml#L67
?

SchSeba:
sure thanks!

PR #1877, opened 2024-04-18
ztp: CNF-12174 Rename proxy to proxySettings in SiteConfig YAML

irinamihai:
CNF-12174
Rename YAML `proxy` to `proxySettings`. This avoids possible confusions and keeps the consistency between the YAML and the golang SiteConfig definitions.

/cc @sabbir-47


ztp/ran-crd/site-config-crd.yaml
```diff
@@ -244,8 +244,8 @@ spec:
                       ignitionConfigOverride:
                         description: Json formatted string containing the user overrides for the initial ignition config.
                         type: string
-                      proxy:
-                        description: Proxy defines the proxy settings for the cluster. If unset, the cluster will not be configured to use a proxy.
+                      proxySettings:
```
sabbir-47, line 247: I wonder if it is a good idea to break the API since some customer may use the `proxy` field and now with the change to `proxySettings` it will be a breaking change for them.

I think it is a good idea to change our cluster template and the variable from `proxySettings` to `proxy`. For the [unified siteconfig in the MCE side](https://github.com/sakhoury/siteconfig/blob/main/config/crd/bases/metaclusterinstall.openshift.io_siteconfigs.yaml#L406), the field is also set to `proxy`.

PR #1891, opened 2024-04-30
CNF-12277: chore: move LVMS to mgmt cores

jakobmoellerdev:
This adds a namespace annotation so that the LVMS workloads will be scheduled into the management core set instead of the isolated cores.


jakobmoellerdev:
@imiller0 do you mind checking this for correctness and adding ok-to-test?

lack:
/override ci/prow/e2e-aws-ran-profile

We really need to fix this job :(

jakobmoellerdev:
@imiller0 @lack would need an approved here, could one of you PTAL?

PR #1893, opened 2024-05-06
OCPBUGS-33299:ztp: fix GM clock type

vitus133:
This commit sets the clock type in the GM reference configuration to OC as per IEEE1588


vitus133:
Closing - not a bug

PR #1897, opened 2024-05-07
ztp: Fix fips=true tag in siteconfig-generator test code

donpenney:
Enabling fips in site-config requires setting fips=true as a boolean in the installConfigOverrides, which means no quotes around the `true`. Having it quoted means it is read as a string and fails to install.

This update has no functional change, it just fixes the test case that includes fips in the installConfigOverrides string, removing the quotes around `true`.


Missxiaoguo (commented):
/lgtm

PR #1901, opened 2024-05-13
OCPBUGS-33573: ztp: IPsec: configure local gateway mode

sabinaaledort:
For IPsec support for encrypting traffic to external hosts OVN-Kubernetes network plugin must be configured in local gateway mode

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/optional-extra-manifest/ipsec/enable-ipsec.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/optional-extra-manifest/ipsec/enable-ipsec.yaml
@@ -7,3 +7,5 @@ spec:
     ovnKubernetesConfig:
       ipsecConfig:
         mode: External
+      gatewayConfig:
+        routingViaHost: true


yuvalk:
> For IPsec support for encrypting traffic to external hosts OVN-Kubernetes network plugin must be configured in local gateway mode

are you sure about that?
from our tests with OCP QE, it works very well in both modes.

lack:
Are there any other consequences to this change that we should document for end users (or us engineers that don't know a lot about the different routing modes)?

bradyallenjohnson:
> Are there any other consequences to this change that we should document for end users (or us engineers that don't know a lot about the different routing modes)?

I thought there were performance impacts. This should be very easy to confirm. @hnhan can you confirm this?

yuvalk:
> Are there any other consequences to this change that we should document for end users (or us engineers that don't know a lot about the different routing modes)?

in SGW network from OVS does not go through the kernel. (what is considered "offloading"):
pod---->br-int--->br-ex---->physical nic
in LGW all traffic goes through the kernel tcpip stack and is subject to it's routing:
pod -->br-int-->ovn-k8s-mp0---kernel route to ipsec tunnel--->physical nic

so LGW is subject to kernel tcpip stack in general and kernel ip routing in particular
allow much much more flexibility and utilizing full networkin featureset of the kernel (bonding, etc). one of the major benefits is probably routing between primary nic and secondary nics which is not available in SGW.
having said all that, which is nice to know, I dont think we need to document that. and LGW/SGW are well documented in the official OCP networking documents (pretty well?)

and though it might come with a cpu overhead price, I think that for RAN, for primary NIC, that is negligible (the primary nic speed is low anyhow.
which is why I believe we should separately discuss moving to LGW across the board for all deployments. it'll be much easier for Telco 5G in general (one less mode to worry about)

in any case, this specific PR is targeting IPsec envs only, that already agreed to have additional CPU, and that is how they are tested (in LGW) so I dont see any risk here.

PR #1903, opened 2024-05-14
ztp: CNF-12580: Remove LVMO operator

sabbir-47:
- LVMO subscription, namespace, operatorGroup has been removed since only LVMS and local storage operator must remain in the source-crs.

/assign @imiller0 @lack @irinamihai


sabbir-47:
> [This](https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/tools/pgt2acmpg/README.md) readme contains references to the removed files and should also be updated. Otherwise LGTM 👍

@irinamihai  Good catch! Updated.

PR #1905, opened 2024-05-15
ztp: CNF-12718: improve json data readability in Siteconfig

sabbir-47:
- Updated json data readability in the SiteConfig examples
- Updated test siteconfig files to validate with the Siteconfig plugin

/assign @lack @irinamihai


lack (approved):
So much cleaner!

PR #1914, opened 2024-05-27
OCPBUGS-33597: Add TIMELS indication on T-GMs

vitus133:
This commit enables NAV-TIMELS indications on WPC T-GM
configurations. This is required to enable automatic
leap seconds file update.
The phc2sys options are also changed to obtain the leap
seconds from phc instead of passing  them as the command
line argument
/cc @josephdrichard @jzding @aneeshkp @nishant-parekh


jzding (approved):
/lgtm

PR #1920, opened 2024-06-05
OCPBUGS-34914: remove leapfile from reference config

vitus133:
Leapfile is managed by linuxptp-daemon, but still appears in the reference config. This config entry will also be logged, which is misleading. This commit removes the entry from the reference config ts2phcConf section.

/cc @aneeshkp @josephdrichard @jzding  @nishant-parekh


vitus133:
closing - will be done later as a tech debt

PR #1925, opened 2024-06-06
CNF-12933: Adding OpenAPI support for Custom CRs when using ACM Policy Generator

edcdavid:
This PR is to integrate support for the openAPI support for custom CR added by policy-generator-plugin: https://github.com/open-cluster-management-io/policy-generator-plugin/pull/159

Missing:
- ~Documentation~
- ~New image containing latest version of policy-generator-plugin -> waiting for http://registry.redhat.io/rhacm2/multicluster-operators-subscription-rhel9:v2.11 that will be available July 18th~
- ~End to end testing~


ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-group-du-sno-ranGen.yaml
```diff
@@ -69,126 +69,16 @@ policies:
...
+        - metadata:
+            name: du-ptp-slave
+          spec:
+            profile:
+                - interface: ens5f0
+                  name: slave
+                  phc2sysOpts: -a -r -n 24
+                  ptp4lOpts: -2 -s --summary_interval -4
+      openapi:
+        path: schema.openapi
```
sabbir-47, line 81: @edcdavid   Shouldn't we have a ptp specific file name? schema.openapi sounds a like a generic name, wheras the schema is based on ptp operator if i'm not mistaken.

edcdavid: Actually, I named the schema.openapi file on purpose to be generic since the schema does not necessarily addresses a single area/resource. It could mix schemas for multiple resources (this is actually how it is retrieved from a running cluster for instance) and a single file with all schemas needed could be easier to manage. If you feel strongly about it I can rename it.

sabbir-47: What is the recommendation when there are multiple different schema? Should we keep in a single file, just trying to understand the best practice.

edcdavid: yes, that would be my recommendation. Keep a single file for all required schemas for a given repo and pass the same file to all patches that require it.

PR #1939, opened 2024-06-17
infra: fixes branch to build the list of commits

rauhersu:
These commits were merged in 4.16 and skipped the syntax verification:

1ccd7824 Revert "Add ClusterImageSet CR in the whitelist of policies-app-project"
c8187edc Add ClusterImageSet CR in the whitelist of policies-app-project

They don't respect the component syntax defined on .githooks/components.sh so any posterior syntax verification makes commits fail Prow tests.

We need this hack to skip the syntax checking.

### DIFF
--- a/hack/verify-commits.sh
+++ b/hack/verify-commits.sh
@@ -23,7 +23,9 @@ if [[ -z "$UPSTREAM_COMMIT" ]]; then
         cp -a /go/src/github.com/openshift-kni/cnf-features-deploy /tmp/
         cd /tmp/cnf-features-deploy
 
-		latest_upstream_commit=$(curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/openshift-kni/cnf-features-deploy/commits?per_page=1 | jq -r '.[0].sha')
+                # Use Prow PULL_BASE_REF to determine the branch.See: https://docs.prow.k8s.io/docs/jobs/#job-environment-variables
+                echo "name of the base branch: $PULL_BASE_REF"
+		latest_upstream_commit=$(curl -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/openshift-kni/cnf-features-deploy/commits?per_page=1&sha=$PULL_BASE_REF" | jq -r '.[0].sha')
 	else
 		if [[ -z "$UPSTREAM_BRANCH" ]]; then
 			latest_upstream_commit="origin/master"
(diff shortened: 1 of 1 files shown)


lack:
Why are we checking the entire commit history? Shouldn't we just check the new pending commits?

lack (approved):

hack/verify-commits.sh
```diff
@@ -23,7 +23,9 @@ if [[ -z "$UPSTREAM_COMMIT" ]]; then
         cp -a /go/src/github.com/openshift-kni/cnf-features-deploy /tmp/
         cd /tmp/cnf-features-deploy
 
-		latest_upstream_commit=$(curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/openshift-kni/cnf-features-deploy/commits?per_page=1 | jq -r '.[0].sha')
+                # Use Prow PULL_BASE_REF to determine the branch.See: https://docs.prow.k8s.io/docs/jobs/#job-environment-variables
+                echo "name of the base branch: $PULL_BASE_REF"
```
sshnaidm, line 27: `PULL_BASE_REF` might be not defined in job, like nighties or others, it should be secured with `${PULL_BASE_REF-}`

rauhersu:
/assign @SchSeba

Apologies for bothering you once again, Sebastian.

SchSeba:
can we please have the 4.16 branch config in the openshift/release repo so we can have the CI lanes running here
I just want to get the confidence that the lanes are green before we merge this PR

rauhersu:
Thanks for commenting @SchSeba . Actually, the commit you  are referring to was the reason to introduce this one, kind of a catch22: the release preparation for 4.16 fails to merge because this cnf-features-deploy test is failing. And, at the same time, we cannot run any test per MR here because we don't have that release preparation merged.

The reason to reach this status was that the 4.16 branch was available firstly and thus, commits were merged but illegitimately : as no presubmit job is currently run for 4.16 commits, they are being merged while tests shouldn't pass.

The 4.16 branch config we need to merge when this test is fixed was prepared here: https://github.com/openshift/release/pull/52175 and its current Prow report prevents it from being merged because of the test we are fixing here.

SchSeba:
so let me first merge the release one and we can see the lanes running here

SchSeba:
ok so all good here I will merge this one
/lgtm
/approve

PR #1959, opened 2024-07-04
infra: Create a special hypershift-ci feature config

sshnaidm:
Create a feature-config for Hypershift which doesn't include
anything not supported like MachineConfigs for example.
Add a special variable for setting hypershift environment to avoid
setting MachineConfigs in the script.


SchSeba (approved):
/lgtm
/approve

PR #1961, opened 2024-07-05
OCPBUGS-36695: remove space from ptp4lopts for PTP HA

aneeshkp:
The ptp4lOpts should be set to an empty string without any spaces for setting up the HA profile in PTP. This PR addresses the issue by removing the unintended empty space.


nishant-parekh (approved):
/lgtm

PR #1967, opened 2024-07-11
cnf-tests: use sriov config daemon for exec instead of mco

SchSeba:
for env like hypershift we don't have MCO so we need to use the sriov-config-daemon that is running on the cluster


SchSeba:
Hi @zeeke when you have time please take a look :)

PR #1980, opened 2024-07-30
[release-4.16] infra: Updated submodules refs

zeeke:
Makefile rule `init-git-submodules` reads environment variables `METALLB_OPERATOR_TARGET_COMMIT`, `SRIOV_NETWORK_OPERATOR_TARGET_COMMIT`, `CLUSTER_NODE_TUNING_OPERATOR_TARGET_COMMIT` to
checkout the right branch for each submodules.
Though we need to find a better way to manage these dependency, this commit prevents Prow jobs to checkout master branch in 4.16 jobs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,9 @@
 export FEATURES?=sctp performance vrf container-mount-namespace metallb tuningcni bondcni
 export SKIP_TESTS?=
 export FOCUS_TESTS?=
-export METALLB_OPERATOR_TARGET_COMMIT?=main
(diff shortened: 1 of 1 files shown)


SchSeba (approved):
I think this is a good solution here so on every but off we make a PR to change the versions no?

zeeke:
> I think this is a good solution here so on every but off we make a PR to change the versions no?

Yeah, but I was wondering if there is a way to get rid of doing this job at every branch cut off. At least an automated PR, I don't know...

PR #1989, opened 2024-08-07
ztp: Update example-sno.yaml

bzvonar:
In the 4.17 RAN RDS, we want to have secure boot enabled as the default boot mode, which is the mode that our System Test enables in their testing of the RAN RDS (this can optionally be disabled by changing bootMode back to just "UEFI").


lack:
@bzvonar If you edit your git commit message and prefix it with `ztp:` the CI should become happy again.

lack (approved):

bzvonar:
Abandoning this since CI is still choking on the commit name, will just start a new one w/ the properly formatted name.

PR #1990, opened 2024-08-08
CNF-13652: Test integration of SRIOV and NMState operators

zeeke:
This changes implements a simple test to verify the NMState operator does not interfere with the SR-IOV network operator.


cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,200 @@
...
+		return "", err
+	}
+
+	return exp.ReplaceAllString(string(buf), ""), nil
+}
+
+func waitForNMStatePolicyToBeStable(nmstatePolicy *nmstatev1.NodeNetworkConfigurationPolicy) {
+	key := runtimeclient.ObjectKey{Name: nmstatePolicy.Name}
+	Eventually(func(g Gomega) {
+		err := client.Client.Get(context.Background(), key, nmstatePolicy)
```
karampok, line 188: the nmstatePolicy is using the arg var, which  creates confusion about the scope of the variables. Maybe have a new one.

Help functions imo are easier to read/be refactored if they are not having many assertions. I always suggest the assertion to be in the main block and have function that return error/true.

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,200 @@
...
+			node, devices, err := findNonPrimarySriovDevicesAndNode(sriovInfos)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(devices)).ToNot(BeZero())
+
+			By("Using node " + node)
+
+			testDevice := devices[0]
+			By("Using device " + testDevice.Name)
+
+			By("Configuring IPAM on Physical Function via NMState")
```
karampok, line 91: Is that different than  "Configuring and IPv4 address"? just IPAM (Address Management) makes me think whereabouts/or other IPAM is being deployed

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,200 @@
...
+    address:
+    - ip: 192.0.2.2
+      prefix-length: 24
+    enabled: true`
+
+func init() {
+	sriovclient = sriovtestclient.New("")
+}
+
+var _ = Describe("[nmstate] SR-IOV Network Operator Integration", Ordered, func() {
```
karampok, line 49: Why the `ordered` is needed?

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,200 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			DeferCleanup(sriovclient.Delete, context.Background(), sriovNetworkNodePolicy)
+
+			Eventually(func() int64 {
+				testedNode, err := sriovclient.Nodes().Get(context.Background(), node, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				resNum := testedNode.Status.Allocatable[corev1.ResourceName("openshift.io/testnmstateresource")]
+				capacity, _ := resNum.AsInt64()
+				return capacity
+			}, 10*time.Minute, time.Second).Should(Equal(int64(10)))
```
karampok, line 120: So the end spec is " there should be 10 VRFs"? 

I suggest to  extend the description that  appears if the spec fails, e.g.
```
Should(Equal(int64(10))," the number of VFs observes is not that same as the one sriov created")

``

File: cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go [hunk 5]
```diff
@@ -0,0 +1,200 @@
...
+	Eventually(func(g Gomega) {
+		err := client.Client.Get(context.Background(), key, nmstatePolicy)
+		g.Expect(err).ToNot(HaveOccurred())
+		availableCondition := nmstatePolicy.Status.Conditions.
+			Find(nmstateshared.NodeNetworkConfigurationPolicyConditionAvailable)
+		g.Expect(availableCondition).ToNot(BeNil())
+		g.Expect(availableCondition.Status).
+			To(Equal(corev1.ConditionTrue))
+	}).
+		WithOffset(1).
```
karampok, line 196: before you did the other version
```
}, 10*time.Minute, time.Second).Should(Equal(int64(10)
```

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,200 @@
...
+				resNum := testedNode.Status.Allocatable[corev1.ResourceName("openshift.io/testnmstateresource")]
+				capacity, _ := resNum.AsInt64()
+				return capacity
+			}, 10*time.Minute, time.Second).Should(Equal(int64(10)))
+		})
+	})
+})
+
+func findNonPrimarySriovDevicesAndNode(n *cluster.EnabledNodes) (string, []*sriovv1.InterfaceExt, error) {
+	errs := []error{}
```
karampok, line 126: You could use this lib  https://pkg.go.dev/errors  to wrap errors

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,200 @@
...
+		for _, device := range devices {
+			if device.Name == mainDeviceName {
+				continue
+			}
+			nonPrimaryDevices = append(nonPrimaryDevices, device)
+		}
+
+		if len(nonPrimaryDevices) > len(retDevices) {
+			retNode = node
+			retDevices = nonPrimaryDevices
```
karampok, line 154: Can you just return here in the happy path, and return the wrap errors.  I might be missing if there is selection between two nodes that have sriov or not

feature-configs/deploy/nmstate/operator_namespace.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    kubernetes.io/metadata.name: openshift-nmstate
+    name: openshift-nmstate
+  name: openshift-nmstate
+spec:
+  finalizers:
+  - kubernetes
```
karampok, line 10: [nit] my yamllint is complaining that it need `---`  and two spaces here (same for all yamls)

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: >the nmstatePolicy is using the arg var, which creates confusion about the scope of the variables. Maybe have a new one.

good point. And since the function only use the name of the policy, I refactor it to pass a string only.

> Help functions imo are easier to read/be refactored if they are not having many assertions. I always suggest the assertion to be in the main block and have function that return error/true.

I'm not 100% with it: helper functions must be short (<20 lines, as a rule of thumb) and that's where they get their readability. On the other side, the call site needs to be way more readable, as it contains the complicated test logic. With the current implementation, the call site reads:

```
waitForNMStatePolicyToBeStable(nodeNetworkConfigPolicy.Name)
```

which reads straightforward to me.
With functions like these, we might lose reusability; a function that waits for different conditions is almost copy/paste, but that's another story.

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: IPv4 is definitely more clear! changing

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: changing

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: I'm using that package at the end of the function:
```
if len(retDevices) == 0 {
	return "", nil, fmt.Errorf("can't find any SR-IOV devices in cluster's nodes: %w", errors.Join(errs...))
}
```

How would you use it with `Wrap(...)`?

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: good point: this function is not very clear. The point is to find the node with the most number of SR-IOV devices. Going to reword the variables and comment to make it clearer.

feature-configs/deploy/nmstate/operator_namespace.yaml
zeeke: Let's make the linter happy :wink:

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
karampok: Pretty sure you can use BeforeEach,  AfterEach stacked on top of Describe/Context to achieve what you need in the future.

According to the doc

> Ginkgo did not include support for Ordered containers for quite some time. As you can see Ordered containers make it possible to circumvent the "Declare in container nodes, initialize in setup nodes" principle; and they make it possible to write dependent specs This comes at a cost, of course - specs in Ordered containers cannot be fully parallelized which can result in slower suite runtimes. Despite these cons, pragmatism prevailed and Ordered containers were introduced in response to real-world needs in the community. Nonetheless, we recommend using Ordered containers only when needed.

So I would say specs(tests) should be not ordered unless they really really need to be.

zeeke: okay, let's go for the BeforeEach way

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
karampok: I think you can avoid the []array
```
func main() {
	var err error
    if false{
	err1 := errors.New("err1")
	err = errors.Join(err, err1)
	err2 := errors.New("err2")
	err = errors.Join(err, err2)
}
	fmt.Println(err)

}
```
but to be honest, both works

feature-configs/ci/nmstate/operator_subscription.yaml
```diff
@@ -0,0 +1,14 @@
...
+    operators.coreos.com/kubernetes-nmstate-operator.openshift-nmstate: ""
+  name: kubernetes-nmstate-operator
+  namespace: openshift-nmstate
+spec:
+  channel: stable
+  installPlanApproval: Automatic
+  name: kubernetes-nmstate-operator
+  source: art-nightly-operator-catalog
+  sourceNamespace: openshift-marketplace
+---
```
karampok, line 14: the `---`  goes to the top
```
yamllint  feature-configs/ci/nmstate/kustomization.yaml
feature-configs/ci/nmstate/kustomization.yaml
  1:1       warning  missing document start "---"  (document-start)
```

zeeke: just noticed all the yaml under `feature-configs` does not have `---` . Lets make it as all the other and live with that until we decide to put yamllint in the CI. Does it sound good?

zeeke:
Testcase `[It] [tuningcni] tuningcni over macvlan should be able to create pod with sysctls over macvlan` is flaking because the image pull might take more than 1 minute. 
[job log](https://storage.googleapis.com/test-platform-results/pr-logs/pull/openshift-kni_cnf-features-deploy/1990/pull-ci-openshift-kni-cnf-features-deploy-master-e2e-aws-ci-tests/1831581177186619392/build-log.txt).

```
[38;5;243m/go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security/tuning.go:25[0m
  tuningcni over macvlan
  [38;5;243m/go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security/tuning.go:38[0m
    [0m[1mshould be able to create pod with sysctls over macvlan[0m
    [38;5;243m/go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security/tuning.go:39[0m
  > Enter [1m[BeforeEach][0m [tuningcni] [38;5;243m- /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/execute/ginkgo.go:9 @ 09/05/24 08:23:29.364[0m
  < Exit [1m[BeforeEach][0m [tuningcni] [38;5;243m- /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/pkg/execute/ginkgo.go:9 @ 09/05/24 08:23:29.378 (14ms)[0m
  > Enter [1m[BeforeEach][0m [tuningcni] [38;5;243m- /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security/tuning.go:34 @ 09/05/24 08:23:29.378[0m
  < Exit [1m[BeforeEach][0m [tuningcni] [38;5;243m- /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security/tuning.go:34 @ 09/05/24 08:23:29.435 (57ms)[0m
  > Enter [1m[It][0m should be able to create pod with sysctls over macvlan [38;5;243m- /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/security/tuning.go:39 @ 09/05/24 08:23:29.435[0m
  [38;5;9m[FAILED] 0001-01-01 00:00:00 +0000 UTC: Successfully assigned tuning-testing/testpod-trz96 to ip-10-0-11-164.ec2.internal
  2024-09-05 08:23:29 +0000 UTC: Add eth0 [10.129.2.15/23] from ovn-kubernetes
  2024-09-05 08:23:30 +0000 UTC: Add net1 [10.10.0.1/24] from tuning-testing/tuning-nad
  2024-09-05 08:23:30 +0000 UTC: Pulling image "quay.io/openshift-kni/cnf-tests:4.16"

  Unexpected error:
      <wait.errInterrupted>: 
      timed out waiting for the condition
      {
          cause: <*errors.errorString | 0xc000481600>{
              s: "timed out waiting for the condition",
          },
      }
  occurred[0m
```

though we can increase the timeout, the problem is that the image is quite big:
```
podman images quay.io/openshift-kni/cnf-tests:4.16
REPOSITORY                       TAG         IMAGE ID      CREATED     SIZE
quay.io/openshift-kni/cnf-tests  4.16        ae9bc250c0cd  6 days ago  1.69 GB
```

Makefile
```diff
@@ -1,5 +1,5 @@
 #TODO add default features here
-export FEATURES?=sctp performance vrf container-mount-namespace metallb tuningcni bondcni
+export FEATURES?=sctp performance vrf container-mount-namespace metallb tuningcni bondcni nmstate
```
fedepaol, line 2: nit: knmstate is more accurate, here and in the following commits

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,207 @@
...
+func init() {
+	sriovclient = sriovtestclient.New("")
+}
+
+var _ = Describe("[nmstate] SR-IOV Network Operator Integration", func() {
+
+	var sriovInfos *cluster.EnabledNodes
+
+	BeforeEach(func() {
+		if !networks.IsSriovOperatorInstalled() {
```
fedepaol, line 54: does it make sense to check for knmstate too?

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,207 @@
+package nmstate
```
fedepaol, line 1: typo on the commit message: opertor

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,207 @@
...
+				WithTimeout(10*time.Minute).
+				Should(Equal(int64(10)), "the number of the observed VFs is not that same as the one sriov created")
+		})
+	})
+})
+
+// findNonPrimarySriovDevicesAndNode returns a node name with its SR-IOV devices. The device list does not include
+// the one used for primary networking, if it is an SR-IOV device.
+// The node with the highest number of SR-IOV devices is returned.
+func findNonPrimarySriovDevicesAndNode(n *cluster.EnabledNodes) (string, []*sriovv1.InterfaceExt, error) {
```
fedepaol, line 131: The name of this is not super clear, and I feel this is doing too many things all together. It finds the node with the highest number of interfaces, and it finds the interfaces for that (excluding the primary).

In general, I'd split this in two steps:
- find the first node with at least two sriov interfaces
- fetch all the interfaces for that node, filtering out the default interface

Also, is there a need to find the one with the highest number of devices or one with one plus the default interface would be enough?

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
fedepaol: also, sohuld

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,207 @@
...
+			Eventually(func() int64 {
+				testedNode, err := sriovclient.Nodes().Get(context.Background(), node, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				resNum := testedNode.Status.Allocatable[corev1.ResourceName("openshift.io/testnmstateresource")]
+				capacity, _ := resNum.AsInt64()
+				return capacity
+			}).
+				WithPolling(5*time.Second).
+				WithTimeout(10*time.Minute).
+				Should(Equal(int64(10)), "the number of the observed VFs is not that same as the one sriov created")
```
fedepaol, line 123: Wdyt about going deeper in checking that both operators did their part? I would check the ip on the interface on the host, to make sure (k)nmstate did its part and probably make some tests on the vf.

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func getPrimaryNICForNode(node string) (string, error) {
+	nodeObj, err := client.Client.Nodes().Get(context.Background(), node, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	buf, err := nodes.ExecCommandOnMachineConfigDaemon(client.Client, nodeObj,
+		[]string{"/bin/bash", "-c", "chroot /rootfs /usr/bin/ovs-vsctl list-ports br-ex | /usr/bin/grep -v patch"})
```
fedepaol, line 178: probably you can get away with checking interfaces which do not have br-ex as master, not sure it would be less fragile

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
```diff
@@ -0,0 +1,207 @@
...
+
+		By("Discover SRIOV devices")
+		sriovInfos, err = cluster.DiscoverSriov(sriovclient, namespaces.SRIOVOperator)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(sriovInfos.Nodes)).ToNot(BeZero(), "no node with SR-IOV NICs found")
+
+		nmstate := nmstatev1.NMState{ObjectMeta: metav1.ObjectMeta{Name: "nmstate"}}
+		err = client.Client.Create(context.Background(), &nmstate)
+		Expect(err).ToNot(HaveOccurred())
+		DeferCleanup(client.Client.Delete, context.Background(), &nmstate)
```
fedepaol, line 72: here we should wait for the resource to be deleted, to avoid state leaking across the tests. Also, please be aware that with knmstate, deleting doesn't always mean "restoring the previous state", ie if the pf was down, it will probably stay up and keep the assigned ip.

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: sounds good

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: > here we should wait for the resource to be deleted, to avoid state leaking across the tests

Do you mean waiting for all the related resources (pods, webhooks) to be cleared?

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: > - find the first node with at least two sriov interfaces
There is no need to have two interfaces. If we have one node with one interface that is not used for primary networking, that node/device pair is good.

>Also, is there a need to find the one with the highest number of devices or one with one plus the default interface would be enough?

No special need. The function can exit early as soon as it finds a good node-device pair.
changing

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: sounds good! changing

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: `ip link` on my dev worker node says `master ovs-system`, but it's not unique:

```
$ oc debug node/cnfdr22.telco5g.eng.rdu2.redhat.com -- chroot /host ip link | grep master
4: eno12399: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq master ovs-system state UP mode DEFAULT group default qlen 1000
10: genev_sys_6081: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 65000 qdisc noqueue master ovs-system state UNKNOWN mode DEFAULT group default qlen 1000
14: 39d9e5dfde97b52@if2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1400 qdisc noqueue master ovs-system state UP mode DEFAULT group default 
16: f50c6bcd4922aa0@if2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1400 qdisc noqueue master ovs-system state UP mode DEFAULT group default 
17: a9c644ea1db880d@if2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1400 qdisc noqueue master ovs-system state UP mode DEFAULT group default 
...
```

while `ovs-vsctl` appears to be a little bit more accurate:
```
$ oc debug node/cnfdr22.telco5g.eng.rdu2.redhat.com -- chroot /host /usr/bin/ovs-vsctl list-ports br-ex | /usr/bin/grep -v patch
eno12399
```

I'd go with the ovs-vsctl way, and I'll consider that if I see flakes, I have plan B. WDYT?

cnf-tests/testsuites/e2esuite/knmstate/knmstate_sriov.go
```diff
@@ -0,0 +1,275 @@
...
+	return true
+}
+
+func ipAddrShow(node, intf string) (string, error) {
+	nodeObj, err := client.Client.Nodes().Get(context.Background(), node, metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("can't get node[%s]: %w", node, err)
+	}
+
+	cmd := []string{"chroot", "/rootfs", "ip", "-brief", "address", "show", "dev", intf}
```
karampok, line 161: The chroot is not needed I think, the container lands in the host networking so ip command should work.
Also just fyi in case you did no know, the ip command has `json` param `ip --json ` in case helps (but here brief and ContainerString is just fine)

cnf-tests/testsuites/e2esuite/knmstate/knmstate_sriov.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func getPrimaryNICForNode(node string) (string, error) {
+	nodeObj, err := client.Client.Nodes().Get(context.Background(), node, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	buf, err := nodes.ExecCommandOnMachineConfigDaemon(client.Client, nodeObj,
+		[]string{"/bin/bash", "-c", "chroot /rootfs /usr/bin/ovs-vsctl list-ports br-ex | /usr/bin/grep -v patch"})
```
karampok, line 210: here chroot is needed because ovs-vsctl command does not exist outside the rootfs

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
fedepaol: yep

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
fedepaol: sorry I wasn't clear. I meant having two functions, one  that finds the node and the other one that gives you the interface for that node. I think it's clearer and each function would have a clearer responsibility

cnf-tests/testsuites/e2esuite/knmstate/knmstate_sriov.go
```diff
@@ -0,0 +1,275 @@
...
+			DeferCleanup(client.Client.Delete, context.Background(), &nodeNetworkConfigPolicy)
+
+			waitForNMStatePolicyToBeStable(nodeNetworkConfigPolicy.Name)
+
+			By("Configuring VF on device via SriovNetworkNodePolicy")
+			sriovNetworkNodePolicy, err := sriovnetwork.CreateSriovPolicy(sriovclient, "test-nmstate-", namespaces.SRIOVOperator, testDevice.Name, node, 10, "testnmstateresource", "netdevice")
+			Expect(err).ToNot(HaveOccurred())
+			DeferCleanup(sriovclient.Delete, context.Background(), sriovNetworkNodePolicy)
+
+			By("Verify Virtual Functions have been correctly configured")
```
fedepaol, line 118: nit: verifying that (same below)

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
fedepaol: and not only that, as I wrote above anything that you applied via the knmstate operator won't be unconfigured when you remove it

zeeke: I added the code to correctly clean the API resources (NMState and NodeNetworkConfigurationPolicy). Other resources created by the operator (handler pod, operator pod) might change over time and it would be out of this test scope.

Regarding the node's state, the best thing we can do is avoid applying changes that might damage other tests. The proposed test case adds a test-only IP address to an unused NIC. 
I believe implementing a revert logic could generate more flakes than not managing it.

WDYT? Is it a fair tradeoff?

cnf-tests/testsuites/e2esuite/nmstate/nmstate_sriov.go
zeeke: Sorry, I'm not figuring out how this function division should be:
the first function should find a node, but I need a node with at least an SRIOV device that is not used as a primary NIC. Then, the function should return the node name, and another function takes the node name and returns all the non-Primary NIC. 

If this is what you meant, then the first function would be 99% equal to what we have now. 

If not, please provide some code skeleton

zeeke: Changed the name of the function to `findNodeWithNonPrimarySriovNIC`

zeeke:
Failing tests
```
CNF Features e2e integration tests: [It] [bondcni] bond over macvlan should be able to create pod with bond interface over macvlan interfaces expand_more	1m4s
CNF Features e2e integration tests: [It] [tuningcni] tuningcni over macvlan should be able to create pod with sysctls over macvlan expand_more
```
are known bugs. See https://issues.redhat.com/browse/CNF-14296

PR #1999, opened 2024-08-13
ignore more annotations

borball:
ignore more annotations for sriovnetwork CR and DisableOLMPprof CM

### DIFF
--- a/ztp/kube-compare-reference/metadata.yaml
+++ b/ztp/kube-compare-reference/metadata.yaml
@@ -204,5 +204,7 @@ fieldsToOmit:
       - pathToKey: metadata.labels."pod-security.kubernetes.io/warn-version"
       - pathToKey: metadata.annotations."capability.openshift.io/name"
       - pathToKey: metadata.annotations."olm.providedAPIs"
+      - pathToKey: metadata.annotations."operator.sriovnetwork.openshift.io/last-network-namespace"
+      - pathToKey: metadata.annotations."include.release.openshift.io/hypershift"
       - pathToKey: metadata.labels."olm.operatorgroup.uid/"
         isPrefix: true


ztp/kube-compare-reference/metadata.yaml
```diff
@@ -204,5 +204,7 @@ fieldsToOmit:
       - pathToKey: metadata.labels."pod-security.kubernetes.io/warn-version"
       - pathToKey: metadata.annotations."capability.openshift.io/name"
       - pathToKey: metadata.annotations."olm.providedAPIs"
+      - pathToKey: metadata.annotations."operator.sriovnetwork.openshift.io/last-network-namespace"
+      - pathToKey: metadata.annotations."include.release.openshift.io/hypershift"
```
imiller0, line 208: Where does this annotation come in? Our current RDS configuration is not hypershift based so this should not be included here/ignored if found in the comparison.

borball:
The "include.release.openshift.io/hypershift" is introduced on all 4.16 clusters seems like. 
I checked two of my clusters(one installed with ABI, the other with ZTP), both of them have annotation "include.release.openshift.io/hypershift" in the cm collect-profiles-config. But on 4.14 clusters I don't see it.

ztp/kube-compare-reference/metadata.yaml
borball: The "include.release.openshift.io/hypershift" is introduced on all 4.16 clusters seems like.
I checked two of my clusters(one installed with ABI, the other with ZTP), both of them have annotation "include.release.openshift.io/hypershift" in the cm collect-profiles-config. But on 4.14 clusters I don't see it.

Seems like added in this PR: https://github.com/openshift/operator-framework-olm/pull/697

ztp/kube-compare-reference/metadata.yaml
```diff
@@ -204,5 +204,7 @@ fieldsToOmit:
       - pathToKey: metadata.labels."pod-security.kubernetes.io/warn-version"
       - pathToKey: metadata.annotations."capability.openshift.io/name"
       - pathToKey: metadata.annotations."olm.providedAPIs"
+      - pathToKey: metadata.annotations."operator.sriovnetwork.openshift.io/last-network-namespace"
```
imiller0, line 207: Verified this is an operator managed field used in tracking where NAD are created for SriovNetworks. 
https://github.com/openshift/sriov-network-operator/blob/master/controllers/generic_network_controller.go#L141-L152

imiller0 (approved):
/lgtm

PR #2000, opened 2024-08-14
ztp: Add ClusterInstance to allowable CR types

imiller0:
CNF-12633 Add ClusterInstance to the set of allowable CR types to be synchronized via gitops. This CR is part of the SiteConfig operator which will allow cluster deployment from a ClusterInstance CR.


sakhoury (approved):
/lgtm

PR #2001, opened 2024-08-15
ztp: make UEFISecureBoot default in 4.17 example yamls

bzvonar:
In the 4.17 RAN RDS, we want to have secure boot enabled as the recommended default boot mode, which is the mode that our System Test enables in their testing of the RAN RDS (this can optionally be disabled by changing bootMode back to just "UEFI").


sakhoury (approved):
/lgtm

PR #2002, opened 2024-08-16
OCPBUGS-38670: ztp: enable automatic leap second handlling

vitus133:
This commit enables NAV-TIMELS indications on WPC T-GM configurations. This is required to enable automatic leap seconds file update.
The phc2sys options are also changed to obtain the leap seconds from phc instead of passing  them as the command line argument.
In addition, it fixes a typo in Tuned patch allowing correct ice dpll scheduling

This is is a manual cherry-pick of #1931, #1943, #1922 and #1975


nishant-parekh (approved):
/lgtm
/approve

PR #2005, opened 2024-08-19
OCPBUGS-38489: ztp: remove unsupported KlusterletAddonConfig clusterLabels and update managecluster CR to include cloud and vendor label

pixelsoccupied:
- Starting acm 2.4 `KlusterletAddonConfig` no longer supports `clusterLabels` in spec. 
   - Removed all the instance of it from code.
- Starting acm 2.10 `Managecluster` CR requires `vendor` label for observability stack to work correctly.
   - if user provides `vendor` and/or `cloud` label through Siteconfig we will respect it
   - otherwise we will add `vendor` and `cloud` with the value `auto-detect`
   - Because `auto-detect` will get replace by ACM (e.g with `OpenShift` for `vendor` key and `Other` for the key `cloud`)...we have to configure argoCD app to allow this change using `RespectIgnoreDifferences`

The change needed for argoCD app may have impact on performance.

An equivalent change will be in Siteconfig operator 

/cc @imiller0 
/cc @irinamihai 
/cc @sabbir-47

### DIFF
--- a/ztp/gitops-subscriptions/argocd/README.md
+++ b/ztp/gitops-subscriptions/argocd/README.md
@@ -108,7 +108,6 @@ In order to deploy the OpenShift GitOps operator v1.12 you may apply the provide
 ```
     $ oc patch multiclusterengines.multicluster.openshift.io multiclusterengine --type=merge --patch-file out/argocd/deployment/disable-cluster-proxy-addon.json
 ```
-
 5. Prepare the ArgoCD pipeline configuration
 - Create a git repository with a directory structure similar to the example directory.
 - Configure access to the repository using the ArgoCD UI. Under *Settings* configure:
@@ -121,6 +120,11 @@ In order to deploy the OpenShift GitOps operator v1.12 you may apply the provide
 6. Apply pipeline configuration to your hub cluster using the following command.
 ```
     oc apply -k out/argocd/deployment
+```
+
+7. Optionally: If configuring an existing hub cluster (i.e skipped step 6, pipeline configuration), starting with ACM 2.10, the following patch is needed to allow ACM to update `vendor` and `cloud` labels in ManagedCluster CRs. This enables observability without user intervention.
+```
+    $ oc patch applications.argoproj.io clusters -n openshift-gitops --type='json' --patch-file out/argocd/deployment/allow-acm-managedcluster-control.json
 ```
 
 ### Deploying a site
--- a/ztp/gitops-subscriptions/argocd/deployment/clusters-app.yaml
+++ b/ztp/gitops-subscriptions/argocd/deployment/clusters-app.yaml
@@ -21,10 +21,20 @@ spec:
     # the sitconfig.yaml exist AND use the ../../hack/patch-argocd-dev.sh script to re-patch the deployment-repo-server
 #    plugin:
 #      name: kustomize-with-local-plugins
+  ignoreDifferences: # recommended way to allow ACM controller to manage its fields. alternative approach documented below (1)
+    - group: cluster.open-cluster-management.io
+      kind: ManagedCluster
+      managedFieldsManagers:
+        - controller
+# (1) alternatively you can choose to ignore a specific path like so (replace managedFieldsManagers with jsonPointers)
+#      jsonPointers:
+#        - /metadata/labels/cloud
+#        - /metadata/labels/vendor
   syncPolicy:
     automated:
       prune: true
       selfHeal: true
     syncOptions:
-    - CreateNamespace=true
-    - PrunePropagationPolicy=background
+      - CreateNamespace=true
+      - PrunePropagationPolicy=background
+      - RespectIgnoreDifferences=true
--- a/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
+++ b/ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
@@ -268,6 +268,24 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp 
 					instantiatedCR = updateAgentClusterInstall(cluster, instantiatedCR)
 				}
 			}
+
+			if kind == "ManagedCluster" {
+				if metadata, ok := instantiatedCR["metadata"].(map[string]interface{}); ok {
+					// ManagedCluster CR needs to have vendor and cloud label for observability stack to work without additional setup
+					labels, ok := metadata["labels"].(map[string]string)
+					if !ok {
+						labels = make(map[string]string)
+						metadata["labels"] = labels
+					}
+					if _, ok := labels["vendor"]; !ok {
+						labels["vendor"] = acmAutoDetect
+					}
+					if _, ok := labels["cloud"]; !ok {
+						labels["cloud"] = acmAutoDetect
+					}
+				}
+			}
+
 			clusterCRs = append(clusterCRs, instantiatedCR)
 		}
 	}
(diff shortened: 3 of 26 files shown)


ztp/gitops-subscriptions/argocd/README.md
```diff
@@ -109,7 +109,12 @@ In order to deploy the OpenShift GitOps operator v1.12 you may apply the provide
     $ oc patch multiclusterengines.multicluster.openshift.io multiclusterengine --type=merge --patch-file out/argocd/deployment/disable-cluster-proxy-addon.json
 ```
 
-5. Prepare the ArgoCD pipeline configuration
+5. Optionally: If configuring an existing hub cluster, starting ACM 2.10 apply the following patch to allow ACM to update `vendor` and `cloud` labels in ManagedCluster CR if needed.
```
irinamihai, line 112:
```suggestion
5. Optionally: If configuring an existing hub cluster, starting with ACM 2.10, the following patch is needed to allow ACM to update `vendor` and `cloud` labels in ManagedCluster CRs. This enables observability.
```
or something along those lines.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
```diff
@@ -1862,6 +1862,74 @@ spec:
 	assert.NotEqual(t, nmState, nil)
 }
 
+func Test_managedclusterLabel(t *testing.T) {
```
irinamihai, line 1865: Might also be worth having a test where no labels are included in the SiteConfig.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
jgato, line 1: wondering if it makes sense to directly set  `vendor=Openshift` and `cloud=baremetal`. Due to the fact that we are creating the cluster and we know it will be Openshift/Baremetal. I guess auto-detect is oka when you are importing a cluster on ACM, and you dont have these values so clear.

sabbir-47: Nope, i think auto-detect should be used, the values must not be hardcoded.

ztp/siteconfig-generator/siteConfig/siteConfigBuilder_test.go
pixelsoccupied: Yep! Added a new test for that.

ztp/gitops-subscriptions/argocd/README.md
pixelsoccupied: Updated

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
pixelsoccupied: @jgato if you need anything else you can always put that label in SiteConfig and ZTP will respect it. But like Sabbir mentioned, you should just let ACM auto figure it out that.

ztp/gitops-subscriptions/argocd/README.md
```diff
@@ -123,6 +122,11 @@ In order to deploy the OpenShift GitOps operator v1.12 you may apply the provide
     oc apply -k out/argocd/deployment
 ```
 
+7. Optionally: If configuring an existing hub cluster (i.e skipped step 6, pipeline configuration), starting with ACM 2.10, the following patch is needed to allow ACM to update `vendor` and `cloud` labels in ManagedCluster CRs. This enables observability without user intervention.
+```
+    $ oc patch applications.argoproj.io clusters -n openshift-gitops --type='json' --patch-file out/argocd/deployment/allow-acm-to-override-vendor-and-cloud-value.json
```
sabbir-47, line 127: Does it mean that patching this json file will add those labels under the ManagedCLusterCR for the existing deployment?

pixelsoccupied: Yeah essentially! So the labels will be there in the ManagedCluster CR right away. But with this patch...argocd will allow "vendor: auto-detect" to update to appropriate value (e.g "vendor: OpenShift"). 

Without this patch argocd will not sync since by default it always tries to match all the content that's in git with the live cluster. So ACM tries to update and Argocd reverts

ztp/siteconfig-generator/siteConfig/siteConfigBuilder.go
```diff
@@ -268,6 +268,24 @@ func (scbuilder *SiteConfigBuilder) getClusterCRs(clusterId int, siteConfigTemp
...
+			if kind == "ManagedCluster" {
+				if metadata, ok := instantiatedCR["metadata"].(map[string]interface{}); ok {
+					// ManagedCluster CR needs to have vendor and cloud label for observability stack to work without additional setup
+					labels, ok := metadata["labels"].(map[string]string)
+					if !ok {
+						labels = make(map[string]string)
+						metadata["labels"] = labels
+					}
+					if _, ok := labels["vendor"]; !ok {
+						labels["vendor"] = "auto-detect"
```
imiller0, line 281: minor nit. "auto-detect" should probably be a defined constant, here and below.

pixelsoccupied:
/hold

There's a feature which allows the ACM controller to take priority over Argocd controller. See "managedFieldsManagers" https://blog.argoproj.io/new-sync-and-diff-strategies-in-argocd-44195d3f8b8c  

Maybe it's better to use that instead of specifying path /metadata/label/bla? Will try it

pixelsoccupied:
with the following we can see the fields that are controlled by acm. 
```
❯ oc get managedclusters mycluster -oyaml --show-managed-fields=true                                                                                        
apiVersion: cluster.open-cluster-management.io/v1
kind: ManagedCluster
...
  managedFields:
...
  - apiVersion: cluster.open-cluster-management.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          v:"managedclusterinfo.finalizers.open-cluster-management.io": {}
          v:"open-cluster-management.io/managedclusterrole": {}
        f:labels:
          f:cloud: {}
          f:clusterID: {}
          f:openshiftVersion: {}
          f:openshiftVersion-major: {}
          f:openshiftVersion-major-minor: {}
          f:vendor: {}
      f:spec:
        f:managedClusterClientConfigs: {}
    manager: controller
```

If we update the argocd app to this...
```
ignoreDifferences:
  - group: cluster.open-cluster-management.io
    kind: 'ManagedCluster'
    managedFieldsManagers:
    - controller
 ```

We dont have to specify the path to label and basically let acm contoller win as long as it's a path that's being managed by it

**Comment by imiller0** [comment]
The managed fields approach sounds good as the default behavior. If a user wants to explicitly managed one of those fields (eg vendor) they could change to using the jsonpath approach. 
The other consideration would be whether there is a notable difference in CPU use associated with either approach.

**Comment by pixelsoccupied** [comment]
/unhold 

Sounds good @imiller0 - updated the CR and left `jsonPointers` as comment to note the alternate approach

**Comment by sabbir-47** [comment]
@imiller0 @irinamihai  Today is the 4.17 branch cut date. Can we get another round of review or approval?

**Comment by imiller0** [comment]
Scale validation will need to be done to ensure the operator runs in approximately the same memory/cpu/time footprint as before.

File: ztp/gitops-subscriptions/argocd/deployment/clusters-app.yaml [hunk 5]
```diff
@@ -21,10 +21,20 @@ spec:
     # the sitconfig.yaml exist AND use the ../../hack/patch-argocd-dev.sh script to re-patch the deployment-repo-server
 #    plugin:
 #      name: kustomize-with-local-plugins
+  ignoreDifferences: # recommended way to allow ACM controller to manage its fields. alternative approach documented below (1)
+    - group: cluster.open-cluster-management.io
+      kind: ManagedCluster
+      managedFieldsManagers:
+        - controller
+# (1) alternatively you can choose to ignore a specific path like so (replace managedFieldsManagers with jsonPointers)
```
irinamihai, line 29: Extra words? ("like so")

PR #2010, opened 2024-08-27
ztp: Remove OperatorHub from metadata.yaml of compare tool

borball:
Since 4.16 disabled the marketplace operator via cluster capabilities in the siteConfig, there is no need to disable default catalogsources with OperatorHub, so we can remove this check from the metadata.yaml


imiller0:
The OperatorHub remains in the reference config for use when the customer deploys a multi-node cluster. In those cases the capabilities are not reduced as they are for SNO so the OperatorHub CR is still needed.

borball:
Close the PR since OperatorHub may be required on MNO.

PR #2021, opened 2024-09-04
OCPBUGS-39149: ztp: Fix for siteconfig-generator panic for empty file

abraham2512:
This includes a fix for the ZTP siteconfig-generator panic if an empty file exists in the extra-manifest directory.
The condition checks and ignores empty files silently.
Also includes a unit test for testing the scenario with empty file.

/cc @sabbir-47


abraham2512:
Closed this PR because of improper commits

PR #2032, opened 2024-09-12
OCPBUGS-39093: reference: Enhance namespace templates to allow optional pod-security.kubernetes.io labels

lack:

### DIFF
--- a/ztp/kube-compare-reference/optional_labels.tmpl
+++ b/ztp/kube-compare-reference/optional_labels.tmpl
@@ -1,0 +1,21 @@
+{{- define "allowLabels" }}
+{{- $prefixes := slice . 1 }}
+{{- $result := dict }}
+{{- range $key, $value := (index . 0) }}
+    {{- range $prefix := $prefixes }}
+        {{- if (hasPrefix $prefix $key) }}
+            {{- $_ := set $result $key $value }}
+        {{- end }}
+    {{- end }}
+{{- end }}
+{{- $result | toYaml | nindent 4 }}
+{{- end }}
+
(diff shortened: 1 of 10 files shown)


ztp/kube-compare-reference/optional_labels.tmpl
```diff
@@ -0,0 +1,21 @@
...
+    {{- range $prefix := $prefixes }}
+        {{- if (hasPrefix $prefix $key) }}
+            {{- $_ := set $result $key $value }}
+        {{- end }}
+    {{- end }}
+{{- end }}
+{{- $result | toYaml | nindent 4 }}
+{{- end }}
+
+{{- define "requiredLabels" }}
```
nocturnalastro, line 14: I don't see where you're using the first index of the list should it be the starting value of the result?

ztp/kube-compare-reference/optional_labels.tmpl
```diff
@@ -0,0 +1,21 @@
...
+{{- $labels := slice . 1 }}
+{{- $result := dict }}
```
nocturnalastro, line 16: I don't see where you're using the first index of the list should it be the starting value of the result?

nocturnalastro: Thinking about this again I think setting the index 0 as the starting value of result would break using `allowLabels` and `requiredLabels` together.

lack: I'm not using the first argument, but figured I'd keep a similar-looking calling convention

natifridman (approved):
/lgtm

PR #2042, opened 2024-09-20
ztp: CNF-14458: Follow up fixes for channel name & kubecompare-references

fontivan:
- Fix channel name to match what should be the final version in the catalog
- Update kube-compare references based on test result feedback

### DIFF
--- a/ztp/kube-compare-reference/required/cluster-logging/ClusterLogForwarder.yaml
+++ b/ztp/kube-compare-reference/required/cluster-logging/ClusterLogForwarder.yaml
@@ -1,4 +1,4 @@
-apiVersion: "logging.openshift.io/v1"
+apiVersion: observability.openshift.io/v1
 kind: ClusterLogForwarder
 metadata:
   name: instance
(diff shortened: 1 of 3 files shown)


ztp/kube-compare-reference/required/cluster-logging/ClusterLogForwarder.yaml
```diff
@@ -1,37 +1,26 @@
...
   namespace: openshift-logging
 spec:
   {{- if .spec.outputs }}
   outputs:
-    - type: {{ (index .spec.outputs 0).type }}
-      name: {{ (index .spec.outputs 0).name }}
-      kafka:
-        url: {{ (index .spec.outputs 0).kafka.url }}
+  - name: {{ (index .spec.outputs 0).name }}
+    type: kafka
```
fontivan, line 10: @imiller0 as far as the kube-compare reference goes should we only allow kafka forwarder or any output?

imiller0: For now kafka as that is what we have evaluated the resource utilzation with. Anything outside of that would need additional verification in the RAN DU use case.

imiller0 (approved):
/lgtm

PR #2046, opened 2024-09-23
ztp: CNF-14458: Add cluster log service account CRs to argocd example…

fontivan:
… and policygen test templates


imiller0 (approved):
/lgtm

PR #2050, opened 2024-09-24
OCPBUGS-39093: ztp: reference: Fix ImageRegistryConfig.yaml to not show differences for null values

lack:

### DIFF
--- a/ztp/kube-compare-reference/optional/image-registry/ImageRegistryConfig.yaml
+++ b/ztp/kube-compare-reference/optional/image-registry/ImageRegistryConfig.yaml
@@ -5,12 +5,16 @@ metadata:
 spec:
   logLevel: Normal
   managementState: Managed
-  observedConfig: {{ .spec.observedConfig | toYaml | indent 2 }}
-  operatorLogLevel: {{ .spec.operatorLogLevel }}
-  {{- if .spec.proxy }}
-  proxy:
-{{ .spec.proxy | toYaml | indent 4 }}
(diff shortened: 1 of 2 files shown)


ztp/kube-compare-reference/optional/image-registry/ImageRegistryConfig.yaml
```diff
@@ -5,11 +5,15 @@ metadata:
 spec:
   logLevel: Normal
   managementState: Managed
+  {{- if .spec.observedConfig }}
   observedConfig: {{ .spec.observedConfig | toYaml | indent 2 }}
```
natifridman, line 9: Not sure how this CR looks like, just making sure it should be `indent` instead of newline with `nindent`

lack: That's a good question.  In the one example I have it's just `null`, so shouldn't need `indent` at all.

lack: Reading more about it [in the api docs](https://docs.openshift.com/container-platform/4.16/rest_api/operator_apis/config-imageregistry-operator-openshift-io-v1.html#spec) didn't help much...

I think it looked like `observedConfig` should just be passed-through directly, whereas `unsupportedConfigOverrides` should be actively prohibited.

natifridman (approved):
/lgtm

PR #2051, opened 2024-09-25
CNF-14285: ztp: reference: Change "pod-security.kubernetes.io/" label to be prefix based

natifridman:


lack:
Given that this is a better way, perhaps we should also remove my `allowLabels` and `requiredLabels` hacks too?

lack:
FYI: The commit log message must begin with `ztp:` and not `CNF-14285: ztp:` - It's fine for the PR to have the CNF-XXXX prefix, but not the git commit.

natifridman:
> Given that this is a better way, perhaps we should also remove my `allowLabels` and `requiredLabels` hacks too?

Removed `allowLabels`

ztp/kube-compare-reference/optional/storage/StorageLVMSubscriptionNS.yaml
```diff
@@ -9,7 +9,4 @@ metadata:
         "workload.openshift.io/allowed: \"management\""
         "openshift.io/cluster-monitoring: \"true\""
```
lack, line 10: We can also get rid of this `requiredLabels` template helper...  In fact, it may be easier to just revert 8ec2e880 entirely.

lack (changes requested):

ztp/kube-compare-reference/metadata.yaml
```diff
@@ -202,10 +202,8 @@ fieldsToOmit:
       - pathToKey: metadata.annotations."include.release.openshift.io/single-node-developer"
       - pathToKey: metadata.annotations."release.openshift.io/create-only"
       - pathToKey: metadata.labels."lca.openshift.io/target-ocp-version"
-      - pathToKey: metadata.labels."pod-security.kubernetes.io/audit"
-      - pathToKey: metadata.labels."pod-security.kubernetes.io/audit-version"
-      - pathToKey: metadata.labels."pod-security.kubernetes.io/warn"
-      - pathToKey: metadata.labels."pod-security.kubernetes.io/warn-version"
+      - pathToKey: metadata.labels."pod-security.kubernetes.io/"
+        isPrefix: true
```
lack, line 206: There's a little more we can remove: Look for every other `pod-security.kubernetes.io` in this section:
(line 187 and line 193 can be delete too)

lack (approved):

PR #2053, opened 2024-09-26
OCPBUGS-42750: Support upgrade from Cluster Logging 5.x to 6.0

imiller0:
The service account "logcollector" is owned by the Cluster Logging Operator version 5.x. If we use the same name for our user-created SA the upgrade from 5.x to 6.0 will delete the SA resulting in a loss of logging. Use log6collector to make it unique.

### DIFF
--- a/ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-group-du-clo5-cleanup.yaml
+++ b/ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-group-du-clo5-cleanup.yaml
@@ -1,0 +1,37 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: PolicyGenerator
+metadata:
+  name: group-du-sno-clo5-cleanup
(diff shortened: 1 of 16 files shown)


ztp/gitops-subscriptions/argocd/example/acmpolicygenerator/acm-group-du-clo5-cleanup.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: policy.open-cluster-management.io/v1
```
sabbir-47, line 1: @imiller0 Not related to this CR, but was wondering if we should update any readme under acmpolicygenerator or in the [argod folders .md file](https://github.com/openshift-kni/cnf-features-deploy/tree/master/ztp/gitops-subscriptions/argocd) have a small section explaining how this policy and the file under source-cr named ClusterLogging5Cleanup.yaml should be used, although the comments on the CR explains that

PR #2055, opened 2024-10-01
ztp: Added compare script to compare rendered CRs with source-crs

zzztx:
This PR includes the following changes:
1. The helm-convert tool is used to generate helm chart from reference files and default_value.yaml;
2. The helm template cmd is used to render chart to CR files;
3. Using script compare.sh to compare the rendered CR files with source-crs.

### DIFF
--- a/ztp/kube-compare-reference/Makefile
+++ b/ztp/kube-compare-reference/Makefile
@@ -1,5 +1,10 @@
+GOPATH  ?= $(HOME)/go
+GOBIN ?= $(GOPATH)/bin
+export HELMCONVERTOR := $(GOBIN)/helm-convert
 DOWNLOAD_URL := $(shell curl -s https://api.github.com/repos/openshift/kube-compare/releases/latest | jq -r '.assets[] | select(.name? | match("linux_amd64")) | .browser_download_url')
 ARCHIVE_NAME := kubectl-cluster_compare-linux_amd64.tar.gz
+HELM_URL := https://get.helm.sh/helm-v3.16.1-linux-amd64.tar.gz
+HELM_PKG := helm-linux-amd64.tar.gz
 
 .PHONY: check
 check: metadata_lint
@@ -24,4 +29,28 @@ metadata_lint: kubectl-cluster_compare
 
 .PHONY: clean
 clean:
-	rm -rf kubectl-cluster_compare
+	rm -rf kubectl-cluster_compare Chartv1 renderedv1 helm
+
+
+.PHONY: convert
+convert: $(HELMCONVERTOR) helm
+	@echo "Converting reference files to Helm Charts."
+	@$(HELMCONVERTOR) -r ./metadata.yaml -n Chartv1 -v default_value.yaml
+	@echo "Rendering Helm Charts to CR files."
+	@./helm template renderedv1 ./Chartv1 --output-dir renderedv1
+
+.PHONY: $(HELMCONVERTOR)
+$(HELMCONVERTOR):
+	@echo "Installing helm-convert tool..."
+	go install github.com/openshift/kube-compare/addon-tools/helm-convert@latest
+
+helm:
+	@echo "Installing helm..."
+	curl -fsSL $(HELM_URL) -o $(HELM_PKG)
+	tar -zxvf $(HELM_PKG) linux-amd64/helm --strip-components 1
+	rm -f $(HELM_PKG)
+
+
+.PHONY: compare
+compare: convert
+	./compare.sh
--- a/ztp/kube-compare-reference/compare.sh
+++ b/ztp/kube-compare-reference/compare.sh
@@ -1,0 +1,60 @@
+#! /bin/bash
+trap cleanup EXIT
+
+function cleanup() {
+  rm -rf source_file rendered_file same_file
+}
+
+function read_dir() {
+  local dir=$1
+  local file
+
+  for file in $(ls "$dir"); do
+    if [ -d "$dir""/""$file" ]; then
+      read_dir "$dir""/""$file"
+    else
+      echo "$dir""/""$file"
+    fi
+  done
+}
+
+
+function compare_cr {
+  local rendered_dir=$1
+  local source_dir=$2
+  status=0
+
+  read_dir "$rendered_dir" |grep yaml  > rendered_file
+  read_dir "$source_dir" |grep yaml  > source_file
+
+
+  while IFS= read -r file1; do
+    while IFS= read -r file2; do
+      if [ "${file1##*/}" = "${file2##*/}" ]; then
+        diff -u "$file1" "$file2"
+        status=$(( "$status" || $? ))
+        printf "\n\n" 
+        echo "$file1" >> same_file
+      fi
+    done < rendered_file
+  done < source_file
+
+
+  while IFS= read -r file; do
+    sed -i "/${file##*/}/d" source_file
+    sed -i "/${file##*/}/d" rendered_file
+  done < same_file
+
+}
+
+
+
+compare_cr renderedv1 ../source-crs
+
+if [[ -s source_file || -s rendered_file ]]; then
+  [ -s source_file ] && printf "\n\nThe following files exist in source-crs only, but not found in reference:\n" && cat source_file
+  [ -s rendered_file ] && printf "\nThe following files exist in reference only, but not found in source-crs:\n" && cat rendered_file
+  exit 1
+else
+  exit $status
+fi
--- a/ztp/kube-compare-reference/default_value.yaml
+++ b/ztp/kube-compare-reference/default_value.yaml
@@ -1,0 +1,177 @@
+optional_image_registry_ImageRegistryConfig:
+- spec:
+    httpSecret: {}
+    observedConfig: {}
+    operatorLogLevel: {}
+    proxy: {}
+    storage: {}
+optional_local_storage_operator_StorageClass:
+- metadata:
+    name: example-storage-class
+optional_local_storage_operator_StorageLV:
+- metadata:
+    name: local-disks
+optional_local_storage_operator_StorageSubscription:
+- spec:
+    source: redhat-operators-disconnected
+optional_ptp_config_PtpConfigDualCardGmWpc:
+- spec:
+    profile:
+    - plugins:
+        e810:
+          pins: $e810_pins
+    recommend:
+    - match:
+      - nodeLabel: "node-role.kubernetes.io/$mcp"
+optional_ptp_config_PtpConfigForHA:
+- spec:
+    recommend:
+    - match:
+      - nodeLabel: "node-role.kubernetes.io/$mcp"
+optional_ptp_config_PtpConfigForHAForEvent:
+- spec:
+    recommend:
+    - match:
+      - nodeLabel: "node-role.kubernetes.io/$mcp"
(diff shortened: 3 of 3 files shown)


ztp/kube-compare-reference/Makefile
```diff
@@ -24,4 +27,32 @@ metadata_lint: kubectl-cluster_compare
...
+	curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+
+
+.PHONY: compare
+compare: convert
+	@if [ -e compare.sh ]; then \
+		chmod 700 ./compare.sh; \
+		./compare.sh; \
+	else \
+		echo "Error: compare.sh no found!"; \
```
lack, line 56: I would not have the if and chmod; just check-in compare.sh with +x permissions, and just run ./compare.sh

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,41 @@
...
+
+for file in `cat same_file`
+do
+    sed -i "/${file##*/}/d" source_file
+done
+
+printf "\n\nThe following files exist in source-crs only, but not found in reference:\n"
+cat source_file
+
+rm -rf same_file rendered_file source_file
```
lack, line 41: In order to serve as a pass/fail for the CI in the future, we should make sure this script exits `0` only if there were no differences found and no extra files detected, and `1` if we saw any differences or missing files.

ztp/kube-compare-reference/Makefile
```diff
@@ -24,4 +27,32 @@ metadata_lint: kubectl-cluster_compare
...
+	rm -rf kubectl-cluster_compare Chartv1 renderedv1
+
+
+.PHONY: convert
+convert: $(HELMCONVERTOR) helm
+	@if [ -e default_value.yaml ]; then \
+		$(HELMCONVERTOR) -r ./metadata.yaml -n Chartv1 -v default_value.yaml; \
+	else \
+		echo "Error: default_value.yaml doesn't exist!"; \
+		exit 0; \
```
lack, line 39: If it's an error, we should not exit 0, we should exit 1.  It may be possible to just call:

```
    $(HELMCONVERTOR) -r ./metadata.yaml -n Chartv1 -v default_value.yaml
```

And if it fails because there's no default_value.yaml that's okay.

ztp/kube-compare-reference/Makefile
```diff
@@ -24,4 +27,32 @@ metadata_lint: kubectl-cluster_compare
...
+	fi
+	@helm template renderedv1 ./Chartv1 --output-dir renderedv1
+
+$(HELMCONVERTOR):
+	go install github.com/openshift/kube-compare/addon-tools/helm-convert@latest
+
+helm:
+	curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
```
lack, line 47: Maybe preface this with something like:
```
    @echo "Installing helm..."
```

ztp/kube-compare-reference/Makefile
```diff
@@ -24,4 +27,32 @@ metadata_lint: kubectl-cluster_compare
 
 .PHONY: clean
 clean:
-	rm -rf kubectl-cluster_compare
+	rm -rf kubectl-cluster_compare Chartv1 renderedv1
```
lack, line 30: If the `helm` step below installs the helm executable to the local directory, perhaps we should also remove it as part of the `clean` step?

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,41 @@
+#! /bin/bash
+
+function read_dir() {
+for file in `ls $1`
```
lack, line 4: It's good practice to declare incoming variables with something like
```
  local dir=$1
```
instead of using $1 everywhere.

Likewise we should declare local variables like `file` as such as well:
```
  local file
```

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,41 @@
...
+    fi
+done
+}
+
+
+read_dir renderedv1 |grep yaml  > rendered_file
+read_dir ../source-crs |grep yaml  > source_file
+
+for file1 in `cat source_file`
+do
```
lack, line 20: I prefer `for $something; do` instead of `do` on its own line

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,41 @@
...
+    for file2 in `cat rendered_file`
+    do
+      if [ "${file1##*/}" = "${file2##*/}" ]
+      then
+        diff -u $file1 $file2
+        printf "\n"
+        echo $file1 >> same_file
```
lack, line 27: Instead of a temporary file, you could use a bash array for this if you like

ztp/kube-compare-reference/compare.sh
lack, line 1: If you haven't done so yet, I recommend running `shellcheck` and `bashate` on bash files.  They're great static analysis tools that can find potential corner cases you may not normally consider.

ztp/kube-compare-reference/compare.sh
yangzhengtx: It's difficult to filter out the same file name from source_file by using array because of different prefix.
I can try that approach in later release.

ztp/kube-compare-reference/Makefile
```diff
@@ -24,4 +27,25 @@ metadata_lint: kubectl-cluster_compare
 
 .PHONY: clean
 clean:
-	rm -rf kubectl-cluster_compare
+	rm -rf kubectl-cluster_compare Chartv1 renderedv1 /usr/local/bin/helm
```
lack, line 30: This seems like it would only work if run as root?

We shouldn't be adding or removing tools in /usr/local/bin as part of our build process.

ztp/kube-compare-reference/Makefile
```diff
@@ -24,4 +27,25 @@ metadata_lint: kubectl-cluster_compare
...
+	@echo "Rendering Helm Charts to CR files."
+	@helm template renderedv1 ./Chartv1 --output-dir renderedv1
+
+$(HELMCONVERTOR):
+	@echo "Installing helm-convert tool..."
+	go install github.com/openshift/kube-compare/addon-tools/helm-convert@latest
+
+helm:
+	@echo "Installing helm..."
+	curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
```
lack, line 46: Can we ask this to install to an alternate location like `$pwd`?

ztp/kube-compare-reference/compare.sh
lack: That's fine!  If a temporary file is easier, stick with it!

ztp/kube-compare-reference/Makefile
yangzhengtx: Yes, I changed the method to curl the binary directly instead of running installation script.

ztp/kube-compare-reference/Makefile
```diff
@@ -24,4 +29,27 @@ metadata_lint: kubectl-cluster_compare
...
+
+
+.PHONY: convert
+convert: $(HELMCONVERTOR) helm
+	@echo "Converting reference files to Helm Charts."
+	@$(HELMCONVERTOR) -r ./metadata.yaml -n Chartv1 -v default_value.yaml
+	@echo "Rendering Helm Charts to CR files."
+	@./helm template renderedv1 ./Chartv1 --output-dir renderedv1
+
+$(HELMCONVERTOR):
```
lack, line 42: By having the HELMCONVERTER executable as the target, we may end up running against older versions of the tool by accident.  This will ensure we run `go install...` every time and this will forcibly keep the executable up to date:
```suggestion
.PHONY: $(HELMCONVERTOR)
$(HELMCONVERTOR):
```

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,46 @@
+#! /bin/bash
+
+function read_dir() {
+local dir=$1
```
lack, line 4: Usually it's good to indent the contents of function blocks by one level

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,46 @@
...
+            printf "\n"
+            echo "$file1" >> same_file
+        fi
+     done < rendered_file
+done < source_file
+
+
+while IFS= read -r file; do
+    sed -i "/${file##*/}/d" source_file
+done < same_file
```
lack, line 36: This will catch whether there are source-crs that have no equivalency in the rendered output.  Do we also need to check for the opposite condition?  Rendered CRs that do not have any matching entry in source-crs?

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,46 @@
...
+
+read_dir renderedv1 |grep yaml  > rendered_file
+read_dir ../source-crs |grep yaml  > source_file
+
+
+while IFS= read -r file1; do
+    while IFS= read -r file2; do
+        if [ "${file1##*/}" = "${file2##*/}" ]
+        then
+            diff -u "$file1" "$file2"
```
lack, line 26: We need to capture whether this command returned non-zero so that if we detected any differences we can exit non-zero at the end of this script.

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,46 @@
...
+done < source_file
+
+
+while IFS= read -r file; do
+    sed -i "/${file##*/}/d" source_file
+done < same_file
+
+
+if [ -s source_file ]
+then
```
lack, line 40: Just like `for ...; do`  I prefer having `if ...; then` all one one line

lack (changes requested):
A couple more issues in comments, but the other thing you'll need to do is fixup your commit history and edit the message(s).  This repository has a requirement that any changes inside the `ztp/**` source tree must all have the prefix `ztp:` at the beginning of the git commit messages.

ztp/kube-compare-reference/compare.sh
yangzhengtx: Yes, I will add more steps here to check that, and output the result.

ztp/kube-compare-reference/compare.sh
```diff
@@ -0,0 +1,60 @@
...
+
+  read_dir "$rendered_dir" |grep yaml  > rendered_file
+  read_dir "$source_dir" |grep yaml  > source_file
+
+
+  while IFS= read -r file1; do
+    while IFS= read -r file2; do
+      if [ "${file1##*/}" = "${file2##*/}" ]; then
+        diff -u "$file1" "$file2"
+        status=$(( "$status" || $? ))
```
lack, line 35: Inside a an arithmetic evaluation like this, you don't need '$' on your variables (But you still do on `$?` for a reason I don't fully understand).

This is equivalent:
```suggestion
        status=$(( status || $? ))
```

lack: Not really important, as I think this works fine as-is.

lack (approved):

lack:
@yangzhengtx I think this is ready to go!  We just need to clean up your git commit history so it passes our git commit message checker.

Rebase everything (squash it into a single commit, probably), and make sure the commit log message starts with `ztp: `.

PR #2056, opened 2024-10-02
Update fieldsToOmit default to list

nocturnalastro:
Depends on: https://github.com/openshift/kube-compare/pull/115


lack:
@nocturnalastro You'll have to edit your commit log message so it starts with `ztp:` instead of the CNF ticket.

nocturnalastro:
Having talked with @lack we decided to go a different way and this is no longer needed.

PR #2059, opened 2024-10-03
Build-in the RAN reference to the ZTP container

lack:
We should include the reference in the container so customers don't have to go upstream for the reference.


lack:
Oops, forget the `ztp:` prefix rule myself :facepalm:

PR #2062, opened 2024-10-04
CNF-14780: fix ptp templates

edcdavid:
This is to fix a typo in ptp templates using interface ens6f0 instead of interface enf5f0 and phc2sysOpts: "-a -r -n 25" instead of phc2sysOpts: "-a -r -n 24"


imiller0 (approved):
/lgtm

PR #2065, opened 2024-10-07
CNF-14818: reference sync

lack:
- **ztp: reference: Ignore helm-specific differences in the helm-convert tool**
- **ztp: reference: Add missing ztp-deploy-wave annotations to the reference**


nocturnalastro:
@lack I think you need to add the wave to the fields to omit

lack:
Yes, good point!

nocturnalastro:
great :) 
/lgtm

PR #2066, opened 2024-10-08
CNF-14818: sync quotes, comments, and whitespace between references and source-crs

lack:
The source-crs changes should not show any functional difference, as
they are largely whitespace or reordering changes.

The reference changes mostly add comments to mirror those in source-crs.


ztp/source-crs/ClusterLogForwarder.yaml
```diff
@@ -1,13 +1,18 @@
...
 kind: ClusterLogForwarder
 metadata:
   name: instance
   namespace: openshift-logging
   annotations:
     ran.openshift.io/ztp-deploy-wave: "10"
 spec:
-# outputs: $outputs
-# pipelines: $pipelines
+  outputs:
```
nocturnalastro, line 9: Looks like this is adding something to the source CRs should this be?

lack: According to the CRD, these 2 fields are required anyway, so having a placeholder instead of a commented-out section is identical... It must be overridden by the end-user either way.

Plus it makes it easier to generate/validate the CRD from the reference.

lack: - https://github.com/openshift/cluster-logging-operator/blob/release-6.0/api/observability/v1/clusterlogforwarder_types.go#L49-L55
- https://github.com/openshift/cluster-logging-operator/blob/release-6.0/api/observability/v1/clusterlogforwarder_types.go#L67-L73

PR #2070, opened 2024-10-15
CNF-14823: Enforce reference and source-crs synchronization via CI

lack:
- **ztp: reference: enforce source-cr and reference alignment in CI**
- **ztp: Introduce 'make sync' to synchronize reference -> source-crs**

### DIFF
--- a/ztp/kube-compare-reference/compare.sh
+++ b/ztp/kube-compare-reference/compare.sh
@@ -1,11 +1,5 @@
 #! /bin/bash
 
-DIFF=${DIFF:-colordiff}
-if ! command -v "$DIFF" >/dev/null; then
-    echo "Warning: Requested diff tool '$DIFF' is not found; falling back to plain old 'diff'"
-    DIFF="diff"
-fi
-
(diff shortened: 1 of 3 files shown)


ztp/kube-compare-reference/compare.sh
```diff
@@ -70,4 +70,98 @@ function compare_cr {
...
+    local status=0
+
+    local -a renderedFiles
+    readarray -t renderedFiles < <(read_dir "$rendered_dir" | grep yaml)
+
+    local -a sourceFiles
+    readarray -t sourceFiles < <(read_dir "$source_dir" | grep yaml)
+
+    local -a excludedFiles
+    readarray -t excludedFiles < <(grep -v '^#' "$exclusionfile" | grep -v '^$')
```
imiller0, line 86: Maybe, for resilience:
```suggestion
    readarray -t excludedFiles < <(grep -v '^ *#' "$exclusionfile" | grep -v '^$')
```

imiller0 (commented):
/lgtm

PR #2075, opened 2024-10-18
infra: Fix gimme version (again...)

fontivan:
- By using .x previously, it would cause the $PATH to not be set correctly and thus the go binaries were not being found
- A more permanent fix it being worked (CNF-14990) but this will fix 4.16 for now


imiller0 (approved):

imiller0:
/override ci/prow/e2e-aws-ran-profile ci/prow/e2e-aws-ci-tests

PR #2077, opened 2024-10-18
ztp: CNF-14991: Fix instability in policy-generator-plugin usage

fontivan:
- Specifically check for and use the exact policy-generator-plugin version used by ACM
- When used in older branches the ACM_BRANCH value can be updated to match the appropriate ACM version in use in that release
- Fixed tmp dir not getting cleaned up due to incorrect working directory when rm was called
- Add .gitignore entry for built kustomize files

### DIFF
--- a/ztp/tools/pgt2acmpg/scripts/build-acmpg-plugin.sh
+++ b/ztp/tools/pgt2acmpg/scripts/build-acmpg-plugin.sh
@@ -11,8 +11,8 @@ echo "Created ${TMP_DIR}"
 pushd "${TMP_DIR}"
 
 # The details we will use to query ACM
-ACM_FORK="stolostron"
-ACM_BRANCH="main"
+ACM_FORK="${ACM_FORK:-stolostron}"
+ACM_BRANCH="${ACM_BRANCH:-main}"
 
 # We need to check what version of the policy generator tag is in use
 POLICY_GENERATOR_TAG=$(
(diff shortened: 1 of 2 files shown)


ztp/tools/pgt2acmpg/scripts/build-acmpg-plugin.sh
```diff
@@ -4,16 +4,38 @@ set -o errexit -o nounset -o pipefail
...
 TMP_DIR=$(mktemp -d -p .)
-echo "Created $TMP_DIR"
-cd "$TMP_DIR"
-git clone --depth 1 --branch main --single-branch https://github.com/open-cluster-management-io/policy-generator-plugin.git
-cd policy-generator-plugin
+echo "Created ${TMP_DIR}"
+pushd "${TMP_DIR}"
+
+# The details we will use to query ACM
+ACM_FORK="stolostron"
```
fontivan, line 14: @edcdavid I did see that originally you had used open-cluster-management instead of stolostron, I am curious if you had a specific reason for that or not

ztp/tools/pgt2acmpg/scripts/build-acmpg-plugin.sh
```diff
@@ -4,16 +4,38 @@ set -o errexit -o nounset -o pipefail
...
+ACM_BRANCH="main"
```
edcdavid, line 15: Should these env variables have default values instead 
: "${ACM_FORK:=solostron}"
: "${ACM_BRANCH:=main}"

edcdavid (approved):
Thanks for fixing this issue.

ztp/tools/pgt2acmpg/scripts/build-acmpg-plugin.sh
edcdavid: I guess it was my mistake, sorry about it

PR #2092, opened 2024-10-25
ztp: CNF-14990: Fix golang ocp builder image in resource generator

fontivan:
- Should match the golang and ocp versions used in current release

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ module github.com/openshift-kni/cnf-features-deploy
 // This should also be matched within this project at:
 //   - cnf-tests/Dockerfile*
 //   - openshift-ci/Dockerfile*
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -5,6 +5,7 @@ module github.com/openshift-kni/cnf-features-deploy
 // This should also be matched within this project at:
 //   - cnf-tests/Dockerfile*
 //   - openshift-ci/Dockerfile*
+//   - ztp/resource-generator/Containerfile
```
fontivan, line 8: I think adding this comment means I need a /approve for `infra` instead of `ztp` which is unfortunate

/cc @imiller0 @ffromani

lack (approved):

ffromani (commented):
/approve
/lgtm

PR #2094, opened 2024-10-25
CNF-15059: ztp: reference: Use binary realease of helm-convert instead of go install [master]

lack:


fontivan:
s/realease/release

PR #2095, opened 2024-10-25
CNF-15048: ztp: reference: Use binary release of helm-convert instead of go install [release-4.17]

lack:


fontivan:
s/realease/release

PR #2096, opened 2024-10-25
CNF-14178: ztp: reference: Use binary realease of helm-convert instead of go install [release-4.16]

lack:


fontivan:
s/realease/release

PR #2129, opened 2024-12-06
fix yaml indentation in ReduceMonitoringFootprint.yaml

borball:
All older versions have the same issue as well, need a cherry pick if this PR is accepted and merged.


alosadagrande:
PTAL since it is causing some policies to become non-compliant when the hub cluster is not applying then. They are applied at installation time outside ACM/MCE.

PR #2114, opened 2024-11-22
cnf-tests: Reduce size of cnf-tests image

zeeke:
Container image size of `quay.io/openshift-kni/cnf-tests:4.18` is 1.85GB:
```
$  docker images quay.io/openshift-kni/cnf-tests:4.18
REPOSITORY                        TAG       IMAGE ID       CREATED        SIZE
quay.io/openshift-kni/cnf-tests   4.18      a58152e3611a   22 hours ago   1.85GB
```

This cause test flakiness, as pulling the image might exceed timeouts. [1]

Most of the problem is in the dnf cache:
```
$ docker run quay.io/openshift-kni/cnf-tests:4.18 du -sh /var/cache/dnf
1.2G    /var/cache/dnf
```

Add `yum clean all` command after installing dependencies in the final image

With these changes, the image size is reduced to `665MB`:
```
$ docker images local-cnf-tests
REPOSITORY        TAG       IMAGE ID       CREATED          SIZE
local-cnf-tests   latest    681d224cbd65   44 seconds ago   665MB
```

[1] https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-kni_cnf-features-deploy/1990/pull-ci-openshift-kni-cnf-features-deploy-master-e2e-aws-ci-tests/1843231458743816192

PR #2139, opened 2024-12-30
CNF-15235: cnf-tests: move to rhel 9

Tal-or:
* move all builder to rhel9
* move cnf-tests image to rhel9
* replace `libhugetlbfs` ,which is not present in rhel9, with a local tool for allocating hugepages (more details in the commits)

### DIFF
--- a/cnf-tests/Dockerfile
+++ b/cnf-tests/Dockerfile
@@ -22,6 +22,20 @@ WORKDIR $TESTER_PATH
 
 RUN go build -mod=vendor -o /sctptest
 
+# build hugepages-allocator's binary
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.19 AS builder-hugepages-allocator
+
+ENV PKG_NAME=github.com/openshift-kni/cnf-features-deploy
+ENV PKG_PATH=/go/src/$PKG_NAME
+ENV TOOL_PATH=$PKG_PATH/cnf-tests/pod-utils/hugepages-allocator
+
+RUN mkdir -p $PKG_PATH
+
+COPY . $PKG_PATH/
+WORKDIR $TOOL_PATH
+
+RUN go build -mod=vendor -o /hugepages-allocator
+
 # build latency-test's runner binaries
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.18 AS builder-latency-test-runners
 
@@ -83,6 +97,8 @@ COPY --from=builder-latency-test-runners /hwlatdetect-runner /usr/bin/hwlatdetec
 COPY --from=builder-latency-test-tools /hwlatdetect /usr/bin/hwlatdetect
 
 COPY --from=oc /go/src/github.com/openshift/oc/oc /usr/bin/oc
+
+COPY --from=builder-hugepages-allocator /hugepages-allocator /usr/bin//hugepages-allocator
 
 COPY --from=go-builder /app/cnf-tests/bin/mirror /usr/bin/mirror
 COPY --from=go-builder /app/cnf-tests/mirror/images.json /usr/local/etc/cnf
--- a/cnf-tests/Dockerfile
+++ b/cnf-tests/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.18 AS builder-stresser
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19 AS builder-stresser
 ENV PKG_NAME=github.com/openshift-kni/cnf-features-deploy
 ENV PKG_PATH=/go/src/$PKG_NAME
 ENV TESTER_PATH=$PKG_PATH/cnf-tests/pod-utils/stresser
@@ -10,7 +10,7 @@ WORKDIR $TESTER_PATH
 
 RUN go build -mod=vendor -o /stresser
 
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.18 AS builder-sctptester
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19 AS builder-sctptester
 ENV PKG_NAME=github.com/openshift-kni/cnf-features-deploy
 ENV PKG_PATH=/go/src/$PKG_NAME
 ENV TESTER_PATH=$PKG_PATH/cnf-tests/pod-utils/sctptester
@@ -23,7 +23,7 @@ WORKDIR $TESTER_PATH
 RUN go build -mod=vendor -o /sctptest
 
 # build hugepages-allocator's binary
-FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.19 AS builder-hugepages-allocator
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19 AS builder-hugepages-allocator
 
 ENV PKG_NAME=github.com/openshift-kni/cnf-features-deploy
 ENV PKG_PATH=/go/src/$PKG_NAME
@@ -37,7 +37,7 @@ WORKDIR $TOOL_PATH
 RUN go build -mod=vendor -o /hugepages-allocator
 
 # build latency-test's runner binaries
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.18 AS builder-latency-test-runners
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19 AS builder-latency-test-runners
 
 ENV PKG_NAME=github.com/openshift-kni/cnf-features-deploy
 ENV PKG_PATH=/go/src/$PKG_NAME
@@ -53,7 +53,7 @@ RUN go build -mod=vendor -o /oslat-runner oslat-runner/main.go && \
     go build -mod=vendor -o /hwlatdetect-runner hwlatdetect-runner/main.go
 
 # build latency testing suite
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.18 AS go-builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19 AS go-builder
 WORKDIR /app
 COPY . .
 RUN make test-bin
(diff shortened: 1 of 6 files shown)


Tal-or:
More work is needed in https://github.com/openshift/release/blob/master/ci-operator/config/openshift-kni/cnf-features-deploy/openshift-kni-cnf-features-deploy-master.yaml 
and on d/s side

cnf-tests/Dockerfile
```diff
@@ -22,6 +22,20 @@ WORKDIR $TESTER_PATH
...
+
+RUN mkdir -p $PKG_PATH
+
+COPY . $PKG_PATH/
+WORKDIR $TOOL_PATH
+
+RUN go build -mod=vendor -o /hugepages-allocator
+
 # build latency-test's runner binaries
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.18 AS builder-latency-test-runners
```
yanirq, line 40: duplicate ?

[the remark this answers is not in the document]
Tal-or: Nope. my bad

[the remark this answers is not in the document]
Tal-or: Actually yes, this is the standard `Dockerfile` not the `Dockerfile.openshift` I wasn't sure whether we want to change images here as well, because the cnf-tests image is centos7 based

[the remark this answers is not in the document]
SchSeba: I must say I am not sure who use this file but if you can please update everything to rhel 9 base.

I don't think this dockerfile really work as we build stuff on rhel 8 and 9 but put everything inside a centos7 image...

cnf-tests/Dockerfile.openshift
```diff
@@ -1,5 +1,5 @@
 # This dockerfile is specific to building the OpenShift CNF stresser image
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.18 as builder-stresser
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.19 as builder-stresser
```
SchSeba, line 2: base on the ART automation 4.19 is base on 1.23 https://github.com/openshift/sriov-network-operator/blob/master/Dockerfile.sriov-network-config-daemon.ocp#L1

can you please update the images?

cnf-tests/pod-utils/hugepages-allocator/main.go
```diff
@@ -0,0 +1,72 @@
...
+		0,                          // Let the kernel choose the address
+		uintptr(args.HugePageSize), // Size of the memory
+		uintptr(syscall.PROT_READ|syscall.PROT_WRITE), // Read/Write permissions
+		uintptr(mmapFlags), // mmap flags
+		0,                  // File descriptor (not used for anonymous memory)
+		0,                  // Offset
+	)
+	if errno != 0 {
+		// there's little we can do here
+		_, _ = fmt.Fprintf(os.Stderr, "Failed to allocate hugepage. errno=%v\n", errno)
```
SchSeba, line 52: should we print using the log package?

cnf-tests/pod-utils/hugepages-allocator/main.go
```diff
@@ -0,0 +1,72 @@
...
+	klog.InfoS("Successfully allocated 1GB HugePage memory", "address", fmt.Sprintf("%p", unsafe.Pointer(addr)))
+
+	klog.InfoS("Sleeping", "period", args.Sleep.String())
+	time.Sleep(args.Sleep)
+
+	// Cleanup: Unmap the memory
+	_, _, errno = syscall.Syscall(syscall.SYS_MUNMAP, addr, uintptr(args.HugePageSize), 0)
+	if errno != 0 {
+		// there's little we can do here
+		_, _ = fmt.Fprintf(os.Stderr, "Failed to unmap hugepage. errno=%v\n", errno)
```
SchSeba, line 68: same as above

cnf-tests/pod-utils/hugepages-allocator/main.go
```diff
@@ -0,0 +1,72 @@
...
+	// Write a byte to the allocated memory
+	*(*byte)(memory) = 42
+
+	klog.InfoS("Successfully allocated 1GB HugePage memory", "address", fmt.Sprintf("%p", unsafe.Pointer(addr)))
+
+	klog.InfoS("Sleeping", "period", args.Sleep.String())
+	time.Sleep(args.Sleep)
+
+	// Cleanup: Unmap the memory
+	_, _, errno = syscall.Syscall(syscall.SYS_MUNMAP, addr, uintptr(args.HugePageSize), 0)
```
SchSeba, line 65: lets put this in a defer function so we are should this gets called

cnf-tests/pod-utils/hugepages-allocator/main.go
```diff
@@ -0,0 +1,72 @@
...
+		os.Exit(1)
+	}
+	memory := unsafe.Pointer(addr)
+	// Write a byte to the allocated memory
+	*(*byte)(memory) = 42
+
+	klog.InfoS("Successfully allocated 1GB HugePage memory", "address", fmt.Sprintf("%p", unsafe.Pointer(addr)))
+
+	klog.InfoS("Sleeping", "period", args.Sleep.String())
+	time.Sleep(args.Sleep)
```
SchSeba, line 62: to make this one right we need to have a sig term channel and when we remove the pod it gets called and we do the cleanup before we exist this app

SchSeba (commented):
nice work!!!

I left a few comments :)

cnf-tests/pod-utils/hugepages-allocator/main.go
Tal-or: From where we should get the signal?

cnf-tests/pod-utils/hugepages-allocator/main.go
Tal-or: I thought it would be a good idea to print to stderr because when we exec the command in the pod we usually expect some data in stderr/sdtout.

How do you suggest to handle this though?

cnf-tests/pod-utils/hugepages-allocator/main.go
SchSeba: https://github.com/GoogleCloudPlatform/golang-samples/blob/main/run/sigterm-handler/main.go#L48

some thing like that so after the test we delete that pod it will have the opportunity to clean the RAM

cnf-tests/pod-utils/hugepages-allocator/main.go
SchSeba: The command is in the pod definition we don't do exec into it.
we can just do `logs` and we will see all the logs from the pod. you already used the klog lib in the code so you can just use it here I think :)

cnf-tests/pod-utils/hugepages-allocator/main.go
Tal-or: When I deleted the pod the memory got cleared, but we can add it anyway, just in case

cnf-tests/Dockerfile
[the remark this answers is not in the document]
Tal-or: Done.

cnf-tests/Dockerfile.openshift
Tal-or: Updated to 1.23, Thanks!

cnf-tests/pod-utils/hugepages-allocator/main.go
Tal-or: Done.

Tal-or:
related to #https://github.com/openshift/release/pull/60244

cnf-tests/Dockerfile
```diff
@@ -59,7 +73,7 @@ RUN yum install -y numactl-devel make gcc && \
     cp hwlatdetect /hwlatdetect && \
     cp cyclictest /cyclictest
 
-FROM quay.io/openshift/origin-oc-rpms:4.16 AS oc
+FROM quay.io/openshift/origin-oc-rpms:4.19 AS oc
 
 # Final image
 FROM centos:7
```
SchSeba, line 79: I think you need to update this one also

SchSeba (commented):
lgtm just a small change in the regular dockerfile :)

cnf-tests/Dockerfile
Tal-or: It was on purpose actually,  if we're changing to rhel9 then both `Dockerfile` and `Dockerfile.openshift` are identical. So one can be deleted

zeeke:
Can you please update also:
- https://github.com/openshift-kni/cnf-features-deploy/blob/089d5f920b42b89047113b90aa583a8bc8b4a1af/cnf-tests/mirror/images.json#L4
- https://github.com/openshift-kni/cnf-features-deploy/blob/089d5f920b42b89047113b90aa583a8bc8b4a1af/cnf-tests/testsuites/pkg/images/images.go#L32

These defaults are needed to use this PR's changes during tests.

Also, it should fix the test:
```
CNF Features e2e integration tests: [It] [tuningcni] tuningcni over macvlan should be able to create pod with sysctls over macvlan 
```

as if flakes because of the cnf-tests image size. See
- https://github.com/openshift-kni/cnf-features-deploy/pull/2114

Tal-or:
```
  > Enter [It] should be able to create pod with bond interface over macvlan interfaces - /go/src/github.com/openshift-kni/cnf-features-deploy/cnf-tests/testsuites/e2esuite/bond/bond.go:36 @ 01/05/25 10:58:16.77
  [FAILED] Unexpected error:
      <wait.errInterrupted>: 
      timed out waiting for the condition
      {
          cause: <*errors.errorString | 0xc000110730>{
              s: "timed out waiting for the condition",
          },
      }
  occurred
```

Tal-or:
depends on: https://github.com/openshift/release/pull/60333

cnf-tests/mirror/images.json
```diff
@@ -1,7 +1,7 @@
 [
     {
         "registry": "quay.io/openshift-kni/",
-        "image": "cnf-tests:4.16"
+        "image": "cnf-tests:4.19"
```
zeeke, line 4:
```suggestion
        "image": "cnf-tests:4.18"
```

cnf-tests has not been branched yet:

```
$ docker pull quay.io/openshift-kni/cnf-tests:4.19
Error response from daemon: manifest for quay.io/openshift-kni/cnf-tests:4.19 not found: manifest unknown: manifest unknown
```

And openshift/release doesn't have any configuration for 4.19 yet.
We'll bump to 4.19 here when we'll branch out.

Tal-or: Done. Thanks for catching that

zeeke (approved):
I can't say much about `cnf-tests/pod-utils/hugepages-allocator/main.go`, the rest LGTM

PR #2146, opened 2025-01-06
[release-4.15] OCPBUGS-47798: Ensure BMH deletion before InfrEnv and NS

imiller0:
This is a manual cherry-pick of #2136 

OCPBUGS-42945, during node or cluster cleanup the BMH must complete deletion before the InfraEnv, NMStateConfig or Namespace are removed. This means the BMH must be in a higher wave (later for creation, earlier for deletion) than those other CRs.

The requirement comes from the cleanup phase of assisted installer/baremetal operator. If the BMH automatedCleanupMode is not disabled the node is rebooted into a new ISO image which cleans the disk. The creation of this ISO requires the Namespace to be not in the deleted state so that a new PreProvisioningImage CR can be created. The ISO creation also requires that the correct NMStateConfig CR be present (it is included in the ISO). The booting of the ISO requires the InfraEnv CR to be present.

### DIFF
--- a/ztp/siteconfig-generator/siteConfig/clusterCRsV1.go
+++ b/ztp/siteconfig-generator/siteConfig/clusterCRsV1.go
@@ -108,7 +108,7 @@ metadata:
   name: "{{ .Node.HostName }}"
   namespace: "{{ .Cluster.ClusterName }}"
   annotations:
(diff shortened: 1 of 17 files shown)


lack:
Oops, overrode the wrong tests :facepalm: but they had already passed :)

PR #2172, opened 2025-01-24
ztp: OCPBUGS-48244: extra-manifests filenames > 30 cause IBI install to fail

abraham2512:
Included in this PR

- Filename validation function in ReadExtraManifestResourceFile() to error if extra-manifest filename not **atmost 30**
- Renamed folders in extra-manifests-builder to reflect ISO standard for filenames. Rendered new renamed machine configs into source-crs/extra-manifests
- Renamed non-rendered files, such as openshift-marketplace-ns.yaml 

/assign @sabbir-47 @imiller0


ztp/siteconfig-generator/siteConfig/filesHandler.go
```diff
@@ -120,3 +126,14 @@ func GetExtraManifestResourceFiles(manifestsPath string) ([]os.FileInfo, error)
 	}
 	return files, err
 }
+
+func CheckFileName(filePath string) error {
```
imiller0, line 130: SiteConfig is deprecated starting in 4.18. I'm fine with the check here but we need to correlate this with behavior in ClusterInstance CR to provide the same protections.

ztp/kube-compare-reference/metadata.yaml
```diff
@@ -38,7 +38,7 @@ parts:
               ignore-unspecified-fields: true
           - path: required/cluster-tuning/monitoring-configuration/ReduceMonitoringFootprint.yaml
           - path: required/cluster-tuning/operator-hub/DefaultCatsrc.yaml
-          - path: required/cluster-tuning/09-openshift-marketplace-ns.yaml
+          - path: required/cluster-tuning/09-openshift-marketplace.yaml
```
lack, line 41: Hmmmm..... I'm a bit surprised there wasn't more fallout in here than just this one rename...

abraham2512:
Closing this as obsolete, since go-diskfs was updated in assisted-service
https://github.com/openshift/assisted-image-service/pull/333

PR #2185, opened 2025-01-28
ztp: ci check to ensure source-cr filenames are within 255 limit

abraham2512:
This PR adds a CI check in ztp/ to ensure files in the source-crs directory are within 255 character limit.

cc: @irinamihai @sabbir-47

### DIFF
--- a/ztp/Makefile
+++ b/ztp/Makefile
@@ -1,6 +1,6 @@
-.PHONY: ci-job test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize test-acm-ztp-generated-policies check-reference
+.PHONY: ci-job test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize test-acm-ztp-generated-policies check-reference checkSourceCRsPath
 
-ci-job: test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize test-acm-ztp-generated-policies check-reference
+ci-job: test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize test-acm-ztp-generated-policies check-reference checkSourceCRsPath
 
 test-policygen:
 	@echo "ZTP: Build policy generator and run test"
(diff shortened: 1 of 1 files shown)


ztp/README.md
```diff
@@ -25,3 +25,7 @@ We suggest breaking down the site plan into components that are common, relevant
 * Site: IP addresses, SRIOV configuration
 
 We look forward to user feedback and will gladly accept pull requests and issues for consideration.
+
+## ZTP Guidelines
+
+To avoid issues due to filename length limitations (ISO standard: not greater than 255), please ensure files in the source-crs folder are within this limit. This includes the length of the full path as per ZTP_HOME set in the ztp-site-generate container.
```
irinamihai, line 31: Maybe also add that it's ISO 9660 with the Rock Ridge extension.

ztp/Makefile
```diff
@@ -19,18 +19,15 @@ checkSourceCRsAnnotation:
 		fi; \
 	done; \
 
-
-
 # ZTP_HOME is a temporary hard-coded path to site-generate container's ztp folder
 # Should remain consistent across upstream, midstream and CI script
 # https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/resource-generator/Containerfile#L28C1-L28C23
```
irinamihai, line 24: This link seems a bit off to me.
I think we can skip it completely, it case that file changes in the future.

ztp/Makefile
```diff
@@ -19,18 +19,15 @@ checkSourceCRsAnnotation:
...
 ZTP_HOME=/home/ztp/
-SOURCE_CRS := source-crs
-# checkSourceCRsPath CI Job is to ensure source crs path length do not exceed 255 characters
-# Related Issue: https://issues.redhat.com/browse/OCPBUGS-48244
-checkSourceCRsPath:
+
+# checkFilePath function ensures that filenames in this repo are not greater than 255.
```
irinamihai, line 27: I would add an extra line explaining where the 255 limit comes from.

ztp/Makefile
```diff
@@ -41,6 +38,15 @@ checkSourceCRsPath:
 	done; \
 	exit $$failures
 
+# CI jobs defined below ensure source crs path length do not exceed 255 characters
```
irinamihai, line 41: I think it would make sense to group these under one target, something like `check-ztp-paths` and use this one at the top of the ztp/Makefile.

ztp/Makefile
```diff
@@ -1,6 +1,6 @@
-.PHONY: ci-job test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize test-acm-ztp-generated-policies check-reference checkSourceCRsPath
+.PHONY: ci-job test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize test-acm-ztp-generated-policies check-reference checkSourceCRsPath checkArgoCDPath checkKubeCompareRefPath
 
-ci-job: test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize test-acm-ztp-generated-policies check-reference checkSourceCRsPath
+ci-job: test-policygen checkExtraManifests checkSourceCRsAnnotation test-policygen-kustomize test-siteconfig test-siteconfig-kustomize test-acm-ztp-generated-policies check-reference checkSourceCRsPath checkArgoCDPath checkKubeCompareRefPath
```
irinamihai, line 3: Same comment as above.

ztp/Makefile
abraham2512: Removed it

ztp/Makefile
abraham2512: Done, created single checkZtpPaths target in makefile

PR #2186, opened 2025-01-29
[release-4.18]: Align test images version 4.18

rbaturov:
This PR consists of the following changes:
* Correcting container test images to `release-4.18`
* Adding the test images as ENV vars in the dockerfile
* Removal of a redundant `dockerfile`


rbaturov:
@Tal-or

hack/run-functests.sh
```diff
@@ -10,8 +10,8 @@ export GINKGO_PARAMS=${GINKGO_PARAMS:-'-vv --show-node-events -timeout 6h'}
 
 #env variables needed for the containerized version
 export TEST_POD_IMAGES_REGISTRY="${TEST_POD_IMAGES_REGISTRY:-quay.io/openshift-kni/}"
-export TEST_POD_CNF_TEST_IMAGE="${TEST_POD_CNF_TEST_IMAGE:-cnf-tests:4.12}"
-export TEST_POD_DPDK_TEST_IMAGE="${TEST_POD_DPDK_TEST_IMAGE:-dpdk:4.12}"
+export TEST_POD_CNF_TEST_IMAGE="${TEST_POD_CNF_TEST_IMAGE:-cnf-tests:4.18}"
+export TEST_POD_DPDK_TEST_IMAGE="${TEST_POD_DPDK_TEST_IMAGE:-dpdk:4.18}"
```
swatisehgal, line 14: We can use the `OCP_VERSION` variable here as well (like we are doing in https://github.com/openshift-kni/cnf-features-deploy/pull/2187/files) as the OCP_VERSION defined in common.sh is 4.18 anyway: https://github.com/openshift-kni/cnf-features-deploy/blob/release-4.18/hack/common.sh#L18.

cnf-tests/Dockerfile.openshift
```diff
@@ -70,6 +70,10 @@ FROM quay.io/openshift/origin-oc-rpms:4.16 AS oc
 # Final image
 FROM openshift/origin-base
 
+ENV IMAGE_REGISTRY=quay.io/openshift-kni/
+ENV CNF_TESTS_IMAGE=cnf-tests:4.18
+ENV DPDK_TESTS_IMAGE=dpdk:4.18
+
```
swatisehgal, line 76: I know we are locking to 4.18 here but might be better to define a single environment variable `OCP_VERSION`  and then populate the versions here based on that here.

```suggestion

ENV OCP_VERSION 4.18
ENV IMAGE_REGISTRY=quay.io/openshift-kni/
ENV CNF_TESTS_IMAGE=cnf-tests:${OCP_VERSION}}
ENV DPDK_TESTS_IMAGE=dpdk:${OCP_VERSION}}

```

Tal-or (commented):
I'm not sure about the deletion of the redundant dockerfile.
it's true that we deleted it on master, but I'm not sure we're suppose to backport this change

hack/run-functests.sh
rbaturov: Yes, we can do that. I didn't invest in having this change since this is not the main branch, and we are not expecting to bump this variable, so I was thinking this might be redundant and it can stay hardcoded.

rbaturov:
> I'm not sure about the deletion of the redundant dockerfile. it's true that we deleted it on master, but I'm not sure we're suppose to backport this change

@SchSeba WDYT?

SchSeba:
> > I'm not sure about the deletion of the redundant dockerfile. it's true that we deleted it on master, but I'm not sure we're suppose to backport this change
> 
> @SchSeba WDYT?

Yes I am fine with that this docker file is ancient and base on centos7 that is more or less dead.
I have no problem removing this file

cnf-tests/Dockerfile.openshift
rbaturov: I added these recommendation [here](https://github.com/openshift-kni/cnf-features-deploy/pull/2187/files) for the master branch

swatisehgal: Sure, makes sense. Thanks!

hack/run-functests.sh
swatisehgal: Sure, it was just a suggestion, not a blocker.

swatisehgal (commented):
/lgtm

cnf-tests/Dockerfile.openshift
rbaturov: @swatisehgal Thanks for your review.

swatisehgal:
/assign @ffromani
for approval

ffromani (commented):
/approve

PR #2187, opened 2025-01-29
Dynamically select test image versions

rbaturov:
Since the OCP_VERSION environment variable is defined in common.sh, we can use it to dynamically determine the versions of the cnf-tests and dpdk images instead of hardcoding them.


swatisehgal (approved):
/lgtm

swatisehgal:
/assign @ffromani
for approval

swatisehgal (commented):
/lgtm

ffromani:
/hold

does the variable expansion work as we hope/expect (did we verify?)
Is this the missing piece to auto-inject the correct version or is this a piece of the puzzle?

ffromani:
/approve

all that said, we do want to move this direction

bartwensley:
@rbaturov - just curious which `common.sh` is setting the ENV variable?

bartwensley:
> > @rbaturov - just curious which `common.sh` is setting the ENV variable?
> 
> https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/common.sh#L19

It looks like that common.sh is used when building the test binaries - how is it getting pulled in by the Dockerfile.openshift? My apologies if this is a dumb question.

rbaturov:
> > > @rbaturov - just curious which `common.sh` is setting the ENV variable?
> > 
> > 
> > https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/common.sh#L19
> 
> It looks like that common.sh is used when building the test binaries - how is it getting pulled in by the Dockerfile.openshift? My apologies if this is a dumb question.

Do you refer to [this](https://github.com/rbaturov/cnf-features-deploy/blob/master/cnf-tests/hack/build-test-bin.sh#L6)?
If yes, then this is getting pulled by the dockerfile [here](https://github.com/rbaturov/cnf-features-deploy/blob/master/cnf-tests/Dockerfile.openshift#L65)

bartwensley:
> > > > @rbaturov - just curious which `common.sh` is setting the ENV variable?
> > > 
> > > 
> > > https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/common.sh#L19
> > 
> > 
> > It looks like that common.sh is used when building the test binaries - how is it getting pulled in by the Dockerfile.openshift? My apologies if this is a dumb question.
> 
> Do you refer to [this](https://github.com/rbaturov/cnf-features-deploy/blob/master/cnf-tests/hack/build-test-bin.sh#L6)? If yes, then this is getting pulled by the dockerfile [here](https://github.com/rbaturov/cnf-features-deploy/blob/master/cnf-tests/Dockerfile.openshift#L65)

OK - so in the end, your comment about "dynamically determine the versions of the cnf-tests and dpdk images" only applies to run-functests.sh - the OCP_VERSION is still hardcoded in the Dockerfile.openshift but I guess that is OK since it needs to be updated each release to use the new builder images.

rbaturov:
> > > > > @rbaturov - just curious which `common.sh` is setting the ENV variable?
> > > > 
> > > > 
> > > > https://github.com/openshift-kni/cnf-features-deploy/blob/master/hack/common.sh#L19
> > > 
> > > 
> > > It looks like that common.sh is used when building the test binaries - how is it getting pulled in by the Dockerfile.openshift? My apologies if this is a dumb question.
> > 
> > 
> > Do you refer to [this](https://github.com/rbaturov/cnf-features-deploy/blob/master/cnf-tests/hack/build-test-bin.sh#L6)? If yes, then this is getting pulled by the dockerfile [here](https://github.com/rbaturov/cnf-features-deploy/blob/master/cnf-tests/Dockerfile.openshift#L65)
> 
> OK - so in the end, your comment about "dynamically determine the versions of the cnf-tests and dpdk images" only applies to run-functests.sh - the OCP_VERSION is still hardcoded in the Dockerfile.openshift but I guess that is OK since it needs to be updated each release to use the new builder images.

Correct

rbaturov:
@SchSeba Can you please override the failed lane?

zeeke:
```
E2E Suite: [It] metallb MetalLB deploy should work when deployed with BGP type FRR-K8s Mode
```
is a known failure

PR #2193, opened 2025-01-30
ztp: OCPBUGS-48088: Add monitoring label for openshift-logging ns

irinamihai:
/cc @imiller0 @lack @sabbir-47


lack:
@irinamihai I think the bug metadata needs a bit of a tweak, then this should merge!

PR #2198, opened 2025-01-31
OCPBUGS-47784: ztp: reference: Enable user customization of PTP priority1, priority2, and domainNumber fields

lack:
Part 2: Include the phc2sys '-n' option for domainNumber, as well as all
configs missed in the 1st commit.

### DIFF
--- a/ztp/kube-compare-reference/optional/ptp-config/PtpConfigBoundary.yaml
+++ b/ztp/kube-compare-reference/optional/ptp-config/PtpConfigBoundary.yaml
@@ -9,7 +9,7 @@ spec:
   profile:
   - name: "boundary"
     ptp4lOpts: "-2"
-    phc2sysOpts: "-a -r -n 24"
+    phc2sysOpts: "-a -r -n (?<domainNumber>[0-9]+)"
     ptpSchedulingPolicy: SCHED_FIFO
     ptpSchedulingPriority: 10
     ptpSettings:
--- a/ztp/kube-compare-reference/optional/ptp-config/PtpConfigMasterForEvent.yaml
+++ b/ztp/kube-compare-reference/optional/ptp-config/PtpConfigMasterForEvent.yaml
@@ -14,7 +14,7 @@ spec:
     # The interface name is hardware-specific
     interface: {{ .interface }}
     ptp4lOpts: "-2 --summary_interval -4"
-    phc2sysOpts: "-a -r -m -n 24 -N 8 -R 16"
+    phc2sysOpts: "-a -r -m -n (?<domainNumber>[0-9]+) -N 8 -R 16"
     ptpSchedulingPolicy: SCHED_FIFO
     ptpSchedulingPriority: 10
     ptpSettings:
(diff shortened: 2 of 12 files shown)


ztp/kube-compare-reference/optional/ptp-config/PtpConfigMasterForEvent.yaml
```diff
@@ -26,9 +26,9 @@ spec:
       #
       twoStepFlag 1
       slaveOnly 0
-      priority1 128
-      priority2 128
-      domainNumber 24
+      priority1 (?<priority1>[0-9]+)
+      priority2 (?<priority2>[0-9]+)
+      domainNumber (?<domainNumber>[0-9]+)
```
aneeshkp, line 31: since line 17 overrides line 31 do we still want to do it for readability ?

ztp/kube-compare-reference/optional/ptp-config/PtpConfigBoundary.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   profile:
   - name: "boundary"
     ptp4lOpts: "-2"
-    phc2sysOpts: "-a -r -n 24"
+    phc2sysOpts: "-a -r -n (?<domainNumber>[0-9]+)"
```
aneeshkp, line 12: I have few more suggestions, we can have single ptpconfig for events and non events if we compare it only opts changes . The reason we have different files  because customer never update ptpconfig after they enable events . I can make changes to code to dynamically update options when event is enabled or do you have better method to validate?. 

*** Event **  
    ptp4lOpts: "-2 --summary_interval -4"
    phc2sysOpts: "-a -r -m -n (?<domainNumber>[0-9]+) -N 8 -R 16"
***  Non event  **
     ptp4lOpts: "-2"
    phc2sysOpts: "-a -r -n 24"

aneeshkp: Also since you are here updating config Can i ask to make these two changes 
Update  MaxInSpecOffset to 1500 for T-GM and we are missing  ptpClockThreshold which is configurable for all types of ptpconfig 

 ptpClockThreshold:
      holdOverTimeout: 1500  
      maxOffsetThreshold: 100
      minOffsetThreshold: -100

ztp/kube-compare-reference/optional/ptp-config/PtpConfigMasterForEvent.yaml
lack: Line 17 is for phc2sys, and line 31 is for ptp4l - Does line 17 affect line 31?

ztp/kube-compare-reference/optional/ptp-config/PtpConfigBoundary.yaml
lack: I really do love the idea of simplifying our config, and if we can automatically add the proper event options when events are enabled, that's probably a great enhancement.

In fact, I think these are all good ideas, but maybe worth their own separate bugs and separate PRs...

PR #2199, opened 2025-02-03
Add script to check outdated image references

rbaturov:
This script ensures that developers update image references when creating a new branch.

How it works:
* A developer creates a new branch (e.g., release-4.19) and pushes it **without** verifying all image references are updated (using 4.19 images).
* He then submits a PR to the release repo to add a new CI configuration for the branch.
* If an outdated image references are detected (e.g 4.18), the script will fail the lane until they are updated and merged.

Note: I have prepared a POC to simulate this working in #2212.
This is the [failure results](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-kni_cnf-features-deploy/2212/pull-ci-openshift-kni-cnf-features-deploy-release-4.18-ci/1886798140703510528)

depends on:
https://github.com/openshift-kni/cnf-features-deploy/pull/2168

### DIFF
--- a/hack/verify-images-updated.sh
+++ b/hack/verify-images-updated.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+# Purpose: This script ensures that specific files containing hardcoded release versions are up to date with the current release and do not reference any previous versions.
+
+# Functionality:
+# 1. Determines if the current branch is a release branch (e.g., release-4.x).
+# 2. Extracts the previous version (e.g. 4.16 for release-4.17).
+# 3. Checks specific files for any occurrences of the previous version.
+# 4. If matches are found, it prompts the user to update references to the current release branch and exits with an error code.
+# 5. Skips checks for non-release branches or missing files.
+
+set -x
+
+branch_name=$(git rev-parse --abbrev-ref HEAD)
+
+release_regex="release-([0-9]+)\.([0-9]+)"
+
+if [[ ! $branch_name =~ $release_regex ]]; then
+    echo "Branch '$branch_name' is not a release branch. Skipping checks."
+    exit 0
+fi
+
+x="${BASH_REMATCH[1]}"
+y="${BASH_REMATCH[2]}"
+previous_version="$x.$((y - 1))" # Example: 4.16 for branch release-4.17
+
+files_to_check=(
+    "cnf-tests/Dockerfile.openshift"
+    "cnf-tests/Dockerfile.konflux"
+    "cnf-tests/mirror/images.json"
+    "cnf-tests/testsuites/pkg/images/images.go"
+    "hack/common.sh"
+    "hack/run-functests.sh"
+)
+
+match_found=false
+
+for file in "${files_to_check[@]}"; do
+    if [[ ! -f "$file"  ]]; then
+        echo "Warning: File '$file' does not exist. Skipping."
+        continue
+    fi
+    if grep -nw "$file" -e "$previous_version" 2>/dev/null ; then
+        echo "Reference to $previous_version found in $file."
+        match_found=true
+    fi
+done
+
+if $match_found; then
+    echo "----------------------------------------------------------------------------------"
+    echo "The files above contain references to the previous release version ($previous_version)."
+    echo "Please update them to the current release branch ($branch_name) to ensure consistency."
+    echo "This can be done by creating a PR against ($branch_name)."
+    echo "For more details, refer to the Branching section in the project's readme.md."
+    exit 1
+else
+    echo "All files are up-to-date. No references to $previous_version were found."
+fi
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,10 @@ govet:
 verify-commits:
 	hack/verify-commits.sh
 
-ci-job: verify-commits gofmt golint govet cnftests-unit
+verify-images-updated:
+	hack/verify-images-updated.sh
+
+ci-job: verify-commits verify-images-updated gofmt golint govet cnftests-unit
 	
 ztp-ci-job:
 	$(MAKE) -C ztp ci-job


hack/verify-images-updated.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# Purpose: This script verifies that no files in a release branch contain references to images from the previous release version.
```
bartwensley, line 3: Just curious why you are limiting this to the previous release version? Why not check for any release that doesn't match the current release. For example, if the current release is release-4.17 fail on anything matching "release-4.*" where * isn't "17".

rbaturov: I was thinking of implementing it by using a regex pattern like 4.* to match versions that are not equal to the current release. However, I’m concerned that this approach might lead to false positives since the 4.* pattern is too broad. There might be a smarter way to handle this—I’m still thinking it through, and this is just a draft. 😊

However, since I am aligning the master branch in all these references to the new release branch (4.19), there should theoretically be no scenario in the future where we skip two or more versions. This is ensured by the script, which would fail the CI and prevent such a situation.

hack/verify-images-updated.sh
```diff
@@ -0,0 +1,56 @@
...
+release_regex="release-([0-9]+)\.([0-9]+)"
+
+if [[ ! $branch_name =~ $release_regex ]]; then
+    echo "Branch '$branch_name' is not a release branch. Skipping checks."
+    exit 0
+fi
+
+x="${BASH_REMATCH[1]}"
+y="${BASH_REMATCH[2]}"
+previous_version="$x.$((y - 1))" # Example: 4.16 for branch release-4.17
```
swatisehgal, line 23: What if the branch was significantly outdated and contained references older than the previous release?

hack/verify-images-updated.sh
```diff
@@ -0,0 +1,56 @@
...
+for file in "${files_to_check[@]}"; do
+    if [[ -f "$file" ]]; then
+        if grep -nw "$file" -e "$previous_version"; then
+            echo "Reference to $previous_version found in $file."
+            match_found=true
+        fi
+    else
+        echo "Warning: File '$file' does not exist. Skipping."
+    fi
+done
```
swatisehgal, line 45: Let's skip early if the file does not exist:
```suggestion
for file in "${files_to_check[@]}"; do
    if [[ ! -f "$file"  ]]; then
        echo "Warning: File '$file' does not exist. Skipping."
        continue
    fi
    if grep -nw "$file" -e "$previous_version" 2>/dev/null ; then
        echo "Reference to $previous_version found in $file."
        match_found=true
    fi
done
```

hack/verify-images-updated.sh
rbaturov: This won't be possible, look at my comment to Bart. Since we aligned the master branch to reference 4.19 images and assuming having this script this would be impossible

hack/verify-images-updated.sh
bartwensley: In what cases would the "release-4.*" pattern be too broad? I still think checking for any release that doesn't match the current one is a good idea.

hack/verify-images-updated.sh
zeeke: +1

hack/verify-images-updated.sh
rbaturov: @bartwensley we are `grep`-ing "4.* " and not "release-4.* ", as most of the image references (if not all) don't have the `release` prefix. Look [here](https://github.com/openshift-kni/cnf-features-deploy/pull/2187/files) for example.
"4.*" is a pattern which will potentially fail us on false positives.

hack/verify-images-updated.sh
rbaturov: Thanks. Updated the code.

hack/verify-images-updated.sh
rbaturov: @bartwensley Did I provide enough clarity here?

bartwensley: OK - I guess the other option would be to match on "=4.*" or ":4.*" but maybe that doesn't catch all the cases.

hack/verify-images-updated.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Purpose: This script ensures that specific files containing hardcoded release versions are up to date with the current release and do not reference any previous versions.
+
+# Functionality:
+# 1. Determines if the current branch is a release branch (e.g., release-4.x).
+# 2. Extracts the previous version (e.g. 4.16 for release-4.17).
+# 3. Checks specific files for any occurrences of the previous version.
+# 4. If matches are found, it prompts the user to update references to the current release branch and exits with an error code.
+# 5. Skips checks for non-release branches or missing files.
+
```
Tal-or, line 11: Maybe it worth add
`set -x` to make it easier for debugging in case we'll find an issue with this script

hack/verify-images-updated.sh
```diff
@@ -0,0 +1,57 @@
...
+    fi
+    if grep -nw "$file" -e "$previous_version" 2>/dev/null ; then
+        echo "Reference to $previous_version found in $file."
+        match_found=true
+    fi
+done
+
+if $match_found; then
+    echo "----------------------------------------------------------------------------------"
+    echo "The following files contain references to the previous release version ($previous_version):"
```
yanirq, line 50: I am missing the variable (list) of the "The following files" here

rbaturov: They are printed before we reach this part. Please take a look of the POC PR failure, in this PR description

yanirq: So you might want to change `The following files` to `The files above`

rbaturov: Yeah that is more correct. Updated.

hack/verify-images-updated.sh
```diff
@@ -0,0 +1,57 @@
...
+    fi
+    if grep -nw "$file" -e "$previous_version" 2>/dev/null ; then
+        echo "Reference to $previous_version found in $file."
+        match_found=true
+    fi
+done
+
+if $match_found; then
+    echo "----------------------------------------------------------------------------------"
+    echo "The files above contain references to the previous release version ($previous_version):"
```
yanirq, line 50: remove` :`

SchSeba:
lgtm nice work!

I will let @bartwensley and @swatisehgal time to answer before I approve it

hack/verify-images-updated.sh
swatisehgal: Thanks for looking into this. I am okay to move ahead.

/lgtm

PR #2206, opened 2025-02-04
OCPBUGS-44271: ztp: reference: Make SriovOperatorConfig optional fields truly optional

lack:

### DIFF
--- a/ztp/kube-compare-reference/required/sriov-operator/SriovOperatorConfig.yaml
+++ b/ztp/kube-compare-reference/required/sriov-operator/SriovOperatorConfig.yaml
@@ -23,6 +23,12 @@ spec:
   #          openshift.io/<resource_name>:  "1"
   #        requests:
   #          openshift.io/<resource_name>:  "1"
+  {{- if hasKey .spec "enableInjector" }}
   enableInjector: false
+  {{- end }}
+  {{- if hasKey .spec "enableOperatorWebhook" }}
   enableOperatorWebhook: false
+  {{- end }}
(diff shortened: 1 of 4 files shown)


ztp/source-crs/SriovOperatorConfigForSNO.yaml
```diff
@@ -21,8 +21,8 @@ spec:
   #          openshift.io/<resource_name>:  "1"
   #        requests:
   #          openshift.io/<resource_name>:  "1"
-  enableInjector: false
-  enableOperatorWebhook: false
   # Disable drain is needed for Single Node Openshift
   disableDrain: true
+  enableInjector: false
+  enableOperatorWebhook: false
```
lack, line 27: Reordering for consistency; no functional change.

ztp/kube-compare-reference/required/sriov-operator/SriovOperatorConfig.yaml
```diff
@@ -23,6 +23,12 @@ spec:
   #          openshift.io/<resource_name>:  "1"
   #        requests:
   #          openshift.io/<resource_name>:  "1"
+  {{- if hasKey .spec "enableInjector" }}
```
lack, line 26: Using `hasKey` will ensure these values are enforced if they're present either in the defaults that create source-crs, or in the CRs we're comparing, but fully ignored if they're not present (as they are optional in the json, and these zero-values will be counted by json marshallers as "unset" and therefore be omitted)

PR #2213, opened 2025-02-04
ztp over Konflux

rauhersu:


rauhersu:
this won´t  fix the current prefix issue (won't re-write history)

rauhersu:
extra commits can't re-write branch history to fix the prefixes.

PR #2229, opened 2025-02-07
cnf-tests: Dump `openshift-multus` namespace on failure

zeeke:
Multus pod logs contains log entries from CNI instances, which  might be critical when debugging test failures.

Dump multus logs either on the SR-IOV conformance test suite and the integration test suite.


SchSeba (approved):
/lgtm
/approve

PR #2274, opened 2025-02-27
ztp: CNF-15527: update resource-generator to pull source-crs from telco-reference

abraham2512:
This PR updates the ZTP site generate container to get telco-ran from telco-reference and create source-crs, extra-manifest, gitops-subscriptions and kube-compare-reference directories in /home/ztp


rauhersu:
/hold

Will discuss some limitations and alternatives about this current approach in regards to hermetic builds.

PR #2280, opened 2025-02-28
tools: add new helper: machineinfo

ffromani:
machineinfo reports hw information in JSON, which can be used by the testsuite to inspect the machine HW capabilities (e.g. NUMA layout, LLC layout).

currently only the k8s format is supported to minimize the dependencies.

The intended workflow is:
- consumer code fetches the output of the tool (JSON format)
  to learn about the machine HW properties
- the tool can be run just once to produce the output, to be stored on
  the node if possible, regenerated each time if needed
- the consumer code, e.g. the e2e suite, can now filter nodes
  depending on the HW properties (e.g. split L3...)
- without this tool the test code or any other consumer code would
  need to bundle this HW detection code, or similar code to get
  the same functionality.
- while bundling the HW detection code in the consumer side (e.g.
  testsuite) is feasible, the HW detection code is very self contained,
  so it seems worthy to bundle it in the cnf-tests image, on which
  e2e tests depend anyway.


ffromani:
> Cool addition. What are the current use cases that we have that require this tool?

done, amended the commit message and the PR description. PTAL.

ffromani:
we prefer #2282

PR #2281, opened 2025-03-03
ztp: make ztp a hermetic build

rauhersu:
We are enabling rpm pre-fetch based on `rpms.lock.yaml` for hermetic builds.

`rpms.lock.yaml `has been generated from the input provided by rpms.in.yaml:

- 'packages': rpms we depend on
- 'repofiles': currently, ubi8 is the runtime base image for ztp

using `rpm-lockfile-prototype` as a generator:

https://github.com/konflux-ci/rpm-lockfile-prototype?tab=readme-ov-file#installation

No additional subscription-manager/activation-key  config has been carried out,as the rpms we currently depend on don't require a Red Hat subscription.

Note: do not update these configuration files manually. The process about how to generate hermetic builds + rpms has been described as part of the konflux doc:

https://konflux.pages.redhat.com/docs/users/how-tos/configuring/activation-keys-subscription.html#_configure_rpm_lockfile_prototype

Additionally, the push/pull yaml files have been configured for hermetic builds + rpm pre-fetch.

### DIFF
--- a/.tekton/ztp-site-generate-4-19-pull-request.yaml
+++ b/.tekton/ztp-site-generate-4-19-pull-request.yaml
@@ -31,6 +31,10 @@ spec:
     value: ztp/resource-generator/Containerfile
   - name: build-args-file
     value: build-args-konflux.conf
+  - name: hermetic
+    value: "true"
+  - name: prefetch-input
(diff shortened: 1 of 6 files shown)


rauhersu:
Kindly review

/assign @fontivan

lack:
Can we add a note about how to update this somewhere in a README in the source tree as well?

PR #2285, opened 2025-03-05
cnf-tests: konflux: EUS RPM-repo support

shajmakh:
TBD

### DIFF
--- a/cnf-tests/.konflux/redhat.repo
+++ b/cnf-tests/.konflux/redhat.repo
@@ -1,0 +1,27 @@
+[rhel-9-for-$basearch-appstream-rpms]
+name = Red Hat Enterprise Linux 9 for $basearch - AppStream (RPMs)
+baseurl = https://cdn.redhat.com/content/dist/rhel9/$releasever/$basearch/appstream/os
+enabled = 1
+gpgcheck = 1
+gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
+sslverify = 1
+sslcacert = /etc/rhsm/ca/redhat-uep.pem
+sslclientkey = /etc/pki/entitlement/3797378807134559379-key.pem
+sslclientcert = /etc/pki/entitlement/3797378807134559379.pem
+sslverifystatus = 1
+metadata_expire = 86400
+enabled_metadata = 1
+
+[rhel-9-for-$basearch-baseos-rpms]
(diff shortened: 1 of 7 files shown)


cnf-tests/.konflux/redhat.repo
```diff
@@ -0,0 +1,56 @@
+[rhel-9-for-$basearch-baseos-eus-rpms]
+name = Red Hat Enterprise Linux 9 for $basearch - BaseOS - Extended Update Support (RPMs)
+baseurl = https://cdn.redhat.com/content/eus/rhel9/9.4/$basearch/baseos/os
```
cgoncalves, line 3: Why are you pinning to 9.4? Why not using the `version` label from the container image? That way you would not need to update here to 9.5, 9.6, etc later on.

shajmakh: Thanks for the comments, I am only missing where is {version} defined? when I ran that it wasn't replaced with its value:
```
WARNING:dnf:Errors during downloading metadata for repository 'rhel-9-for-x86_64-baseos-rpms':
  - Status code: 404 for https://cdn.redhat.com/content/dist/rhel9/${version}/x86_64/baseos/os/repodata/repomd.xml (IP: 2.16.30.83)
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/dnf/repo.py", line 574, in load
    ret = self._repo.load()
  File "/usr/lib64/python3.9/site-packages/libdnf/repo.py", line 331, in load
    return _repo.Repo_load(self)
libdnf._error.Error: Failed to download metadata for repo 'rhel-9-for-x86_64-baseos-rpms': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```
it also failed to detect it with only `{version}`:

`  - Status code: 404 for https://cdn.redhat.com/content/dist/rhel9/{version}/x86_64/baseos/os/repodata/repomd.xml (IP: 2.16.30.83)`

shajmakh: Thanks! that worked indeed but I had to remove some of the parameters originally provided in redhat.repo. I wonder if that is possible without the need to add the repos in the input file but in .repo. I'm discovering this option.
do you have any reference to docs that help know the accepted structure of rpms.in.yaml?

cgoncalves: See the README file in https://github.com/konflux-ci/rpm-lockfile-prototype

PR #2297, opened 2025-03-10
CNF-16811: ztp: Add an example PtpConfig for the T-GM 3-card WPC

lack:

### DIFF
--- a/ztp/kube-compare-reference/default_value.yaml
+++ b/ztp/kube-compare-reference/default_value.yaml
@@ -75,6 +75,53 @@ optional_ptp_config_PtpConfigDualCardGmWpc:
   captureGroup_defaults:
     priority2: 128
     domainNumber: 24
+optional_ptp_config_PtpConfigThreeCardGmWpc:
+- spec:
+    profile:
+    - plugins:
+        e810:
+          settings:
+            LocalMaxHoldoverOffSet: 1500
+            LocalHoldoverTimeout: 14400
+            MaxInSpecOffset: 1500
+          pins:
+            $iface_timeTx1:
+              SMA1: 2 1
+              SMA2: 2 2
+              U.FL1: 0 1
+              U.FL2: 0 2
+            $iface_timeTx2:
+              SMA1: 1 1
+              SMA2: 0 2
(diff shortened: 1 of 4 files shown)


ztp/kube-compare-reference/default_value.yaml
```diff
@@ -75,6 +75,53 @@ optional_ptp_config_PtpConfigDualCardGmWpc:
...
+              U.FL2: 0 2
+            $iface_timeTx3:
+              SMA1: 1 1
+              SMA2: 0 2
+              U.FL1: 0 1
+              U.FL2: 0 2
+      ptpClockThreshold:
+        holdOverTimeout: 5
+        maxOffsetThreshold: 100
+        minOffsetThreshold: -100
```
vitus133, line 106: This threshold is not aligned with the MaxInSpecOffset at line 86, which is 1500ns. It will create a problem when recovering from holdover.
Let's assume you are in holdover and reached the offset of 1000ns, still below MaxInSpecOffset, and then the GNSS reception recovers. When it happens, we move to Locked where ptpClockThreshold guards the offset. The time error we have accumulated in holdover (1000ns) might not be immediately fixed, and we will cross ptpClockThreshold and move to Free-run
Described here:
https://issues.redhat.com/browse/OCPBUGS-51311

ztp/source-crs/PtpConfigThreeCardGmWpc.yaml
```diff
@@ -0,0 +1,267 @@
...
+      #
+      productDescription ;;
+      revisionData ;;
+      manufacturerIdentity 00:00:00
+      userDescription ;
+      timeSource 0x20
+    ptpClockThreshold:
+      holdOverTimeout: 5
+      maxOffsetThreshold: 100
+      minOffsetThreshold: -100
```
vitus133, line 262: Same as above

vitus133 (commented):
Added a comment on thresholds alignment.
Otherwise looks good!

lack:
Github seems to have broken itself on this PR.  Re-opening a new one shortly.

PR #2312, opened 2025-03-17
[release-4.17] cnf-tests: switch to ocp/builder base for multi-arch support

shajmakh:
To be able to support aarch64 cnf-tests build u/s, the base image need
to be arm compatible and using `ocp/builder:rhel-8-enterprise-base-multi-openshift-4.18`
will provide a multi-arch build.


SchSeba:
do we need the same for main?

https://github.com/openshift-kni/cnf-features-deploy/blob/master/cnf-tests/Dockerfile.openshift#L99

if that is the case we should start there

PR #2315, opened 2025-03-17
ztp: Remove deprecated Event API version configuration

rdavid:


ztp/source-crs/PtpOperatorConfigForEvent.yaml
```diff
@@ -9,7 +9,6 @@ spec:
   daemonNodeSelector:
     node-role.kubernetes.io/$mcp: ""
   ptpEventConfig:
-    # example "2.0", default to "1.0" if not set
     apiVersion: $event_api_version
```
jzding, line 12: I think this should be removed in 4.19+

rdavid: You mean line `apiVersion: $event_api_version` should be removed, right?

aneeshkp: master will be 4.19 right @lack ?

rdavid: @aneeshkp What do you think?

PR #2319, opened 2025-03-18
ztp: Adjust 3-card PTP T-GM reference to not rely on fallback behavior

lack:
Until we fix OCPBUGS-53199, there's really no good way to mix different
match classes, since matches with an exact api/kind/namespace/name will
always trump api/kind/namespace/* even if the latter is a closer match.

This change works around that behavior but forces the PtpConfig to be
named "grandmaster".  We can either relax the reference all at once
(allowing all PtpConfig to be named anything), or wait until the bug is
fixed and released to just relax some of the PtpConfigs.


aneeshkp (approved):
/lgtm

PR #2328, opened 2025-03-21
ztp: Support the patching of stringData in Secret source files

janosi:
The purpose of this contribution is to implement the support of patching the stringData field in Secret source files via the PGT. 

It can be beneficial in the cases when the value of a secret is used to deliver a configuration file to the target application, and some parameters in the file should be collected from other secrets with the "fromSecret" template function.

Example:
The Secret definition in the policy should look like:

```yaml
kind: Secret
apiVersion: v1
metadata:
  name: name
  namespace: namespace
stringData:
  mySecretFile.yaml: |
    username: {{hub fromSecret "ns" "secretname" "username" | base64dec hub}}
    password: {{hub fromSecret "ns" "secretname" "password" | base64dec hub}}
    host: https://github.com
    port: 3445
```

### DIFF
--- a/ztp/policygenerator/policyGen/policyBuilder.go
+++ b/ztp/policygenerator/policyGen/policyBuilder.go
@@ -315,6 +315,15 @@ func (pbuilder *PolicyBuilder) getCustomResource(sourceFile utils.SourceFile, so
 		resourceMap["binaryData"] = pbuilder.setValues(resourceMap["binaryData"].(map[string]interface{}), sourceFile.BinaryData)
 	}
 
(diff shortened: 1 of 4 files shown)


irinamihai:
Thank you for addressing this.
/lgtm

@imiller0, would you kindly take a quick look?

imiller0 (commented):
One minor comment, otherwise LGTM. Thanks!

janosi:
Hello @imiller0 Thank you for the review! I think, I fixed the finding.

PR #2329, opened 2025-03-21
Allow source files without metadata

janosi:
Sometimes we would like to fill the metadata of a resource from a PGT patch. For example, when the PGT is constructed with an "overlay"  template and the metadata of a resource should come from the input values of the overlay template. In such cases the creation of the metadata in the source file with some placeholder values should not be necessary.

Moreover, if someone does not define metadata for the resource in the source file and then in the PGT they define metadata fields the current implementation panics, as the code accesses resourceMap["metadata"] without any check for the existence of the metadata field. A panic should not occur.

This is a proposal here to implement a solution that avoids the panic and that enables the creation of the metadata from the PGT.

### DIFF
--- a/ztp/policygenerator/policyGen/policyBuilder.go
+++ b/ztp/policygenerator/policyGen/policyBuilder.go
@@ -241,6 +241,14 @@ func (pbuilder *PolicyBuilder) getCustomResource(sourceFile utils.SourceFile, so
 	if err != nil {
 		return resourceMap, err
 	}
+	if sourceFile.Metadata.Name != "" ||
+		sourceFile.Metadata.Namespace != "" ||
+		len(sourceFile.Metadata.Labels) != 0 ||
+		len(sourceFile.Metadata.Annotations) != 0 {
+		if _, exists := resourceMap["metadata"]; !exists {
+			resourceMap["metadata"] = make(map[string]interface{})
+		}
+	}
 	if sourceFile.Metadata.Name != "" {
 		resourceMap["metadata"].(map[string]interface{})["name"] = sourceFile.Metadata.Name
 	}
--- a/ztp/policygenerator/policyGen/policyBuilder_test.go
+++ b/ztp/policygenerator/policyGen/policyBuilder_test.go
@@ -1517,3 +1517,60 @@ spec:
 	assert.Nil(t, err)
 	assert.NotNil(t, policies)
 }
+
+func TestMetadataCreation(t *testing.T) {
+	input := `
+apiVersion: ran.openshift.io/v1
+kind: PolicyGenTemplate
+metadata:
+  name: "test"
+  namespace: "test"
+spec:
+  sourceFiles:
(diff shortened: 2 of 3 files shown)


irinamihai:
Hello! Many thanks for the PR and for addressing the panic. That should not happen, for sure.

My only comment here is that we probably don't want to accept source-crs without `metadata.name`. Having the `metadata.namespace` missing would be acceptable. This behaviour would be consistent with the [ACM PolicyGenerator](https://github.com/open-cluster-management-io/policy-generator-plugin/tree/main) which currently is the recommended method for applying the DU profile.

As reference, when `metadata.name` is missing in the source-cr, this is the message that the ACM PolicyGenerator returns:
`error generating policies from the PolicyGenerator file 'pgt.yaml': failed to process the manifest at "example-source-cr.yaml": all manifests must have the "metadata.name" field set to a non-empty string`

Another note is that PGT can override the metadata name and namespace, so they could have any name in the source-crs.

Please let me know what you think of this. I'm happy to discuss further.

janosi:
On the other hand: if we introduce now a check for the existence of `metadata.name` then I am afraid we may break some environments: for example, a user may have now a source file like

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  namespace: test
```
Such a source file was OK with the existing version of the generator, because it had `metadata`. 

But if we introduce the check for `metadata.name` then that will break such use cases.

janosi:
So, now I introduced the following logic:
- if `metadata` is missing in the source file: error log, no panic
- if `metadata.name` is missing in the source file but the PGT contains `metadata.name` patch: continue the processing
- if `metadata.name` is missing in the source file and the PGT does not contain `metadata.name` patch: error log, no panic

janosi:
UPDATE: during the tests it turned out that some resources are generated without a `metadata.name.` E.g. `SriovNetworkNodeState` is generated with "inform" policy to validate the existing policies on the cluster. As the metadata.name of `SriovNetworkNodeState` is the name of the corresponding node, the validator generator cannot fill the `metadata.name`, and it seems to be a valid use case.
For that reason I removed the check for the existence of `metadata.name`.
It means, that from now on the PR is just about fixing the panic when `metadata` is missing.

irinamihai:
I think this is a good approach.
@imiller0, would you have time to take a quick look?

/lgtm

imiller0:
Thanks @janosi for the PR!
I made a couple notes on very minor fixes but the updated logic lgtm.

janosi:
Thank you @imiller0 for the review!
Excuse me, but I cannot see those notes you mention. Could you please help me in finding those? Thank you!

imiller0:
@janosi apologies. I missed on submitting the review! They should be visible now. 
Thanks!

janosi:
@imiller0  No problem :) Thank you for the comments. I think I fixed those.

imiller0 (approved):
/lgtm

PR #2333, opened 2025-03-26
ztp: add dual-follower ptpconfig example

edcdavid:
Adding PTP Dual Follower configuration templates

### DIFF
--- a/ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
+++ b/ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
@@ -1,0 +1,138 @@
+# In this example, 2 ports of a single NIC are used in a active/standby configuration to synchronize to a 
+# remote grandmaster via interfaces $iface1 and $iface1
+apiVersion: ptp.openshift.io/v1
+kind: PtpConfig
+metadata:
+  name: dual-ordinary-clock
+  namespace: openshift-ptp
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "10"
+spec:
+  profile:
+  {{- range .spec.profile }}
+  - name: oc1
+    phc2sysOpts: -a -r -n (?<domainNumber>[0-9]+) -N 8 -R 16 -u 0
+    ptp4lConf: |-
+      [(?<iface1>[[:alnum:]]+)]
+      masterOnly 0
+      [(?<iface2>[[:alnum:]]+)]
+      masterOnly 0
+
+      [global]
+      #
+      # Default Data Set
+      #
+      slaveOnly 1
+      twoStepFlag 1
+      priority1 128
+      priority2 (?<priority2>[0-9]+)
+      domainNumber (?<domainNumber>[0-9]+)
+      #utc_offset 37
+      clockClass 255
+      clockAccuracy 0xFE
+      offsetScaledLogVariance 0xFFFF
+      free_running 0
+      freq_est_interval 1
+      dscp_event 0
+      dscp_general 0
+      dataset_comparison G.8275.x
+      G.8275.defaultDS.localPriority 128
+      #
+      # Port Data Set
+      #
+      logAnnounceInterval -3
+      logSyncInterval -4
+      logMinDelayReqInterval -4
+      logMinPdelayReqInterval -4
+      announceReceiptTimeout 3
+      syncReceiptTimeout 0
+      delayAsymmetry 0
+      fault_reset_interval -4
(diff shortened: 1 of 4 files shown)


ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
```diff
@@ -0,0 +1,132 @@
...
+      productDescription ;;
+      revisionData ;;
+      manufacturerIdentity 00:00:00
+      userDescription ;
+      timeSource 0xA0
+    ptp4lOpts: -2  --summary_interval -4
+    ptpClockThreshold:
+      holdOverTimeout: 65
+      maxOffsetThreshold: 100
+      minOffsetThreshold: -100
```
lack, line 125: We usually template this out as well in the other PTP templates, as we expect users may want to edit this section based on their own requirements for offset quality and holdover.

ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+      [global]
+      #
+      # Default Data Set
+      #
+      slaveOnly 1
+      twoStepFlag 1
+      priority1 128
+      priority2 128
+      domainNumber 24
```
lack, line 28: Ditto here as above to add a capturegroup for the domainNumber.

lack (changes requested):
Needs a little more templating/capturegroup work to allow user flexibility, but it's close!

ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
edcdavid: Fixed

ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
edcdavid: Fixed

ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
```diff
@@ -0,0 +1,138 @@
...
+metadata:
+  name: ordinary-clock-1
+  namespace: openshift-ptp
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "10"
+spec:
+  profile:
+  {{- range .spec.profile }}
+  - name: oc1
+    phc2sysOpts: -a -r -n 24 -N 8 -R 16 -u 0
```
lack, line 14:
```suggestion
    phc2sysOpts: -a -r -n (?<domainNumber>[0-9]+) -N 8 -R 16 -u 0
```
This `24` should be customizable, and the same as `domainNumber`, below.  By using the same named capturegroup for both we will enforce (in 4.19) that they are the same value.

ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
```diff
@@ -0,0 +1,138 @@
+# In this example, 2 ports of a single NIC are used in a active/standby configuration to synchronize to a 
+# remote grandmaster via interfaces $iface1 and $iface1
+apiVersion: ptp.openshift.io/v1
+kind: PtpConfig
+metadata:
+  name: ordinary-clock-1
```
lack, line 6: Maybe a name like `dual-ordinary-clock` would be more descriptive?

lack (changes requested):

lack:
The ztp-ci failure looks like a build system issue.

Try rebasing to the latest master branch in case it's fixed there... If not we can override the test.

ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
edcdavid: Fixed

ztp/kube-compare-reference/optional/ptp-config/PtpConfigDualFollower.yaml
edcdavid: Fixed

edcdavid:
> The ztp-ci failure looks like a build system issue.
> 
> Try rebasing to the latest master branch in case it's fixed there... If not we can override the test.

I rebased but looks like the CI issue is still present.

PR #2336, opened 2025-03-27
dummy pr

sudomakeinstall2:


openshift-ci/Dockerfile.tools
fontivan, line 1: does fedora 40 have golang 1.23 or does that also need to be bumped?

go.mod
```diff
@@ -7,8 +7,7 @@ module github.com/openshift-kni/cnf-features-deploy
 //   - openshift-ci/Dockerfile*
 //   - ztp/resource-generator/Containerfile
 //   - ztp/tools/pgt2acmpg/go.mod
-go 1.22
-toolchain go1.22.5
+go 1.23
```
fontivan, line 10: I believe one of the dependency bots will make this 1.23.0 if you don't

openshift-ci/Dockerfile.tools
sudomakeinstall2: It was passing locally, but updated it anyway

ztp/resource-generator/Containerfile
```diff
@@ -1,4 +1,4 @@
-ARG ZTP_BUILD_IMAGE=registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.18
+ARG ZTP_BUILD_IMAGE=registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19
```
lack, line 1: I don't think we can do this yet - The build image has to be rhel-8 based as long as we produce binaries (site-generator and pgt-generator) that need to run inside a rhel8 container (ArgoCD)

PR #2338, opened 2025-03-28
infra: Bump fedora to 41 in Dockerfile.tools

sudomakeinstall2:


lack:
The Dockerfile change needs to get in to unblock other CI lanes.  Override the e2e test that should not be affected by the Dockerfile changes.

PR #2339, opened 2025-03-28
ztp: reference: Ensure PtpOperatorConfig apiVersion is a string, and transportHost is optional

lack:


aneeshkp (approved):
/lgtm

PR #2346, opened 2025-04-01
cnf-tests: remove `special-resource-operator` dependency

zeeke:
`github.com/openshift-psap/special-resource-operator` API dependency is not used by the project


go.mod
```diff
@@ -7,8 +7,9 @@ module github.com/openshift-kni/cnf-features-deploy
 //   - openshift-ci/Dockerfile*
 //   - ztp/resource-generator/Containerfile
 //   - ztp/tools/pgt2acmpg/go.mod
-go 1.22
-toolchain go1.22.5
+go 1.22.4
```
SchSeba, line 10: this is weird no?

the tool version newer than the go version?

zeeke: I don't know. It is changed by `go mod tidy`.

But there is no reason to commit that bump in this PR, so I reverted those two lines.

PR #2349, opened 2025-04-01
CNF-13139: ztp: Add hardwareTuning to PerformanceProfile

bartwensley:
Add example hardwareTuning section to PerformanceProfile CR to show how isolated and reserved CPU frequencies can be configured.


bartwensley:
Had to update the commit title as I didn't realize the "ztp:" had to be the first thing in the title.

PR #2357, opened 2025-04-04
ztp: CNF-15527: sub-module for telco-reference

abraham2512:
This PR adds telco-reference repository as a sub-module in resource generator and a ci-check to restrict updates to ztp folders.

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -2,6 +2,8 @@
 
 . $(dirname "$0")/components.sh
 . $(dirname "$0")/check-skipped-files.sh
+. $(dirname "$0")/skipped-files.sh
+. $(dirname "$0")/restricted-dirs.sh
 
 # check if we must skip any check for files in staging
 filenames=$(git diff --name-only --cached | tr '\n' ' ')
@@ -52,4 +54,15 @@ for filename in $filenames; do
 			exit 12
 		fi
 	fi
+done
+
+# check if commits contain files in restricted directories
+for filename in $filenames; do
+    for dir in "${restricted_dirs[@]}"; do
+        if [[ "$filename" == "$dir"* ]]; then
+            echo "ERROR: $filename is in restricted directory $dir"
+            echo "The following ztp directories have been moved: source-crs, extra-manifests, gitops-subscriptions and kube-compare-reference. Please create a pull request in https://github.com/openshift-kni/telco-reference"
+            exit 16
+        fi
+    done
 done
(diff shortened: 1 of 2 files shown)


.githooks/commit-msg
```diff
@@ -2,6 +2,8 @@
 
 . $(dirname "$0")/components.sh
 . $(dirname "$0")/check-skipped-files.sh
+. $(dirname "$0")/skipped-files.sh
```
rauhersu, line 5: Do we need to add this logic here, as I did too: https://github.com/openshift-kni/cnf-features-deploy/commit/37cef2fb6cbf12aff23e1ea66ef305472c29c268#diff-4511378f6fef48566c3597f7d9263c2f5469182c8e7216912f6a7d8d2dc4b777R65 ?

renovate.json
```diff
@@ -0,0 +1,29 @@
...
+        "automergeStrategy": "rebase",
+        "automergeType": "pr",
+        "enabled": true,
+        "ignoreTests": false,
+        "platformAutomerge": true,
+        "schedule": ["0 1 * * *"]
+    },
+    "git-submodules": {
+        "_comment":"We want the Tekton Mintmaker to update git-submodules daily",
+        "schedule": ["0 1 * * *"]
```
rauhersu, line 27: Thanks for considering this !

.githooks/commit-msg
abraham2512: There is a check for skipped files [here](https://github.com/openshift-kni/cnf-features-deploy/blob/2e2c95fd54e15810ff67293a3c7e8b8b48001523/.githooks/commit-msg#L10)
If I'm not wrong, it will only pass through if not all files are skipped

renovate.json
abraham2512: Yes I agree the daily cadence for refresh would align with our current process

.githooks/commit-msg
rauhersu: yeah, I believe that if a developer has the commit hook enabled, it will use your (.githooks/commit-msg) logic, performing that check.

But that might not be 100% sure. This is the reason why 'verify-commits.sh' is sometimes used as a double check: it can provide the same check if the developer runs 'make ci-job' locally or performs the push directly (Prow will run that command as part of the tests too).

In my case I performed this double check, for the code I linked above, including logic in both places. If we are happy with that '.githooks/commit-msg' here we can leave this as it is.

PR #2372, opened 2025-04-10
Update ghw library to latest

mrniranjan:
This is required for newly added machineinfo
feature where hardware information can be fetched

latest update fixes a bug related to
https://github.com/jaypipes/ghw/issues/283

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/golang/glog v1.2.4
 	github.com/google/go-cmp v0.6.0
 	github.com/ishidawataru/sctp v0.0.0-20210707070123-9a39160e9062
-	github.com/jaypipes/ghw v0.14.0
+	github.com/jaypipes/ghw v0.16.0
 	github.com/k8snetworkplumbingwg/multi-networkpolicy v1.0.1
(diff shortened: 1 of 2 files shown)


ffromani:
> This is required for newly added machineinfo feature where hardware information can be fetched
> 
> latest update fixes a bug related to [jaypipes/ghw#283](https://github.com/jaypipes/ghw/issues/283)

it won't hurt, but the issue was on NTO side

go.mod
```diff
@@ -7,8 +7,9 @@ module github.com/openshift-kni/cnf-features-deploy
 //   - openshift-ci/Dockerfile*
 //   - ztp/resource-generator/Containerfile
 //   - ztp/tools/pgt2acmpg/go.mod
-go 1.22
-toolchain go1.22.5
+go 1.22.4
+
+toolchain go1.23.7
```
SchSeba, line 12: I don't think this is right lets bump the all project to 1.23 not only the toolchain

ffromani: `ghw` 0.16.0 doesn't require golang 1.23, so we should probably avoid this change entirely

PR #2406, opened 2025-05-01
CNF-23433: ztp: simplify test-generator.sh script

sebrandon1:
Refactor and cleanup of the test-generator.sh script because I saw there was some repeated code.

### Script improvements:
* Updated the script shebang to use `/usr/bin/env bash` and added stricter shell settings (`set -euo pipefail` and `IFS`) for improved error handling and script safety.
* Added a check to ensure the `generator` script exists before sourcing it, with a clear error message if not found.

### Test structure refactoring:
* Introduced a `check_output` helper function to streamline test output validation and reduce repetitive code. This function runs a command, captures its output, and checks it against an expected pattern.
* Replaced repetitive test case logic with calls to the `check_output` function, simplifying test case definitions for both `run_siteconfigGen` and `run_policyGen` commands.

### DIFF
--- a/ztp/resource-generator/tests/test-generator.sh
+++ b/ztp/resource-generator/tests/test-generator.sh
@@ -1,17 +1,31 @@
-#!/bin/bash
+#!/usr/bin/env bash
+IFS=$'\n\t'
 
 test_script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-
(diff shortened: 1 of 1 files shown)


sebrandon1:
Jira status update: [CNF-23433](https://redhat.atlassian.net/browse/CNF-23433) is currently **To Do**.

PR #2316, opened 2025-03-18
cnf-test: konflux: support arm64 build

shajmakh:
Adapt build config to be arm64 compatible. See: https://konflux.pages.redhat.com/docs/users/getting-started/multi-platform-builds.html

PR #2416, opened 2025-05-05
cnf-tests: add `findutils` binary

shajmakh:
The binary is useful for testing so add it.

### DIFF
--- a/cnf-tests/Dockerfile.openshift
+++ b/cnf-tests/Dockerfile.openshift
@@ -104,7 +104,7 @@ ENV CNF_TESTS_IMAGE=cnf-tests:${OCP_VERSION}
 ENV DPDK_TESTS_IMAGE=dpdk:${OCP_VERSION}
 
 # python3 is needed for hwlatdetect
-RUN yum install -y lksctp-tools iproute tmux ethtool iputils numactl-libs iptables kmod linuxptp iperf3 python3 nc iptables && \
+RUN yum install -y lksctp-tools iproute tmux ethtool iputils numactl-libs iptables kmod linuxptp iperf3 python3 nc iptables findutils && \
     yum clean all
 
 RUN mkdir -p /usr/local/etc/cnf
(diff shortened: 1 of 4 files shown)


ffromani:
/lgtm

the typo in the commit message summary is not worth a resubmit

shajmakh:
> /lgtm
> 
> the typo in the commit message summary is not worth a resubmit

Thank you. I'll fix that in the backport

shajmakh:
@ffromani I pushed again to update the OpenShift Dockerfile as well as I missed it before (and fixed commit msg)

cnf-tests/Dockerfile.openshift
```diff
@@ -104,7 +104,7 @@ ENV CNF_TESTS_IMAGE=cnf-tests:${OCP_VERSION}
 ENV DPDK_TESTS_IMAGE=dpdk:${OCP_VERSION}
 
 # python3 is needed for hwlatdetect
-RUN yum install -y lksctp-tools iproute tmux ethtool iputils numactl-libs iptables kmod linuxptp iperf3 python3 nc iptables && \
+RUN yum install -y lksctp-tools iproute tmux ethtool iputils numactl-libs iptables kmod linuxptp iperf3 python3 nc iptables findutils && \
```
ffromani, line 107: ok, I missed this but why again do we need this? aren't we moving to konflux as the only pipeline?

shajmakh: yes we are but openshift ci uses this dockerfile to generate the u/s images and QE needs ARM supported cnf-tests image.The Arm support via konflux is filed (https://github.com/openshift-kni/cnf-features-deploy/pull/2316) yet need to enhance it, but since there are no plans to support konflux-builds with ARM soon, we want to preserve the openshift ci option and keep relevant files updated.

ffromani: ok, makes sense. Thanks.

ffromani (commented):
/lgtm

PR #2422, opened 2025-05-06
cnf-tests:k5x: adapt official conforma

shajmakh:
Adapt to the rules of the official conforma:
* replace untrusted image with production image using the lastest
  closest available build.

### DIFF
--- a/cnf-tests/.konflux/Dockerfile
+++ b/cnf-tests/.konflux/Dockerfile
@@ -1,6 +1,6 @@
 ARG RHEL_VERSION=9.4
 
-FROM brew.registry.redhat.io/rh-osbs/openshift-ose-cli-rhel9:v4.19 AS oc
+FROM registry.redhat.io/openshift4/ose-cli-rhel9:v4.18 AS oc
 
 FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:v1.23 AS builder-stresser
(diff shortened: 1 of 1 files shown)


shajmakh:
official conformas passed. `cnf-tests-conforma-on-pr-4-20 / cnf-tests-4-20` is stale.

cnf-tests/.konflux/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG RHEL_VERSION=9.4
 
-FROM brew.registry.redhat.io/rh-osbs/openshift-ose-cli-rhel9:v4.19 AS oc
+FROM registry.redhat.io/openshift4/ose-cli-rhel9:v4.18 AS oc
```
yanirq, line 3: can we use the latest here ? (for 4.20 at this point)

shajmakh: there is no 4.20 tag in production, latest one is 4.18. we need to update that once 4.19 is GA'd. This was discussed offline as well.

shajmakh: we need anyhow to align with same OC version eventually so we need to update anyway. Not specifying the tag will give us the latest which is anyhow 4.18

ffromani (commented):
/approve

PR #2433, opened 2025-05-12
cnf-tests: k5x: tag go binaries with `strictfipsruntime`

shajmakh:
For FIPs compliancy, go binaries in images must be built while using go tag `strictfipsruntime`.


shajmakh:
/hold
let's wait on https://github.com/openshift/cluster-node-tuning-operator/pull/1328 to get in first; update the submodules; and then test the generated build via the integration test

ffromani:
/approve

Worst case this is overkill (I don't think it is), unlikely to cause any harm

PR #2470, opened 2025-05-16
tekton: add owners file in ztp konflux

abraham2512:
cc : @rauhersu @fontivan 
Adds k5x owners file in ztp/.konflux


abraham2512:
@sabbir-47 could you please add approve

PR #2435, opened 2025-05-13
[release-4.19] cnf-tests konflux backports

shajmakh:
Updates backports to 4.19 and target_branch fix. Please see commits for detailed info.

PR #2504, opened 2025-05-19
chore(deps): update konflux references

shajmakh:
clone from https://github.com/openshift-kni/cnf-features-deploy/pull/2399/commits without the submodule update to unblock 4.19 conformas.


shajmakh:
not a good idea, we need TA tasks first which are in https://github.com/openshift-kni/cnf-features-deploy/pull/2435

PR #2564, opened 2025-05-26
tekton: update service account in ztp 4.19

abraham2512:
This PR adds service account to ztp site generate


abraham2512:
@fontivan CI should be enabled here once this PR merges https://github.com/openshift/release/pull/63998

PR #2597, opened 2025-06-05
renovate cleanup 20250605

shajmakh:
Please review the commits' details.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -47,8 +47,6 @@
         ]
     },
     "git-submodules": {
-        "schedule": [
-            "at any time"
-        ]
+        "enabled": true
     }
 }


shajmakh:
meant Steven Skeard but reviews are welcome if relevant. 
/cc @fontivan

renovate.json
```diff
@@ -35,11 +32,17 @@
...
+            {
+                "automerge": true,
+                "matchUpdateTypes": [
+                    "digest"
+                ]
+            }
         ]
     },
     "git-submodules": {
-        "schedule": [
```
abraham2512, line 41: Hi @shajmakh, why are we removing this `at any time` config? 
Removing this will make Renovate bot update submodules only once a week (Tuesday) and as often as allowed (4 hrs) like we need for telco-reference submodule. Please take a look at https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#scheduling

renovate.json
```diff
@@ -35,11 +32,17 @@
...
-            "at any time"
```
abraham2512, line 42: Hi @shajmakh , why are we removing this at any time config?
Removing this will make Renovate bot update submodules only once a week (Tuesday) and as often as allowed (4 hrs) like we need for telco-reference submodule. Please take a look at https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#scheduling

renovate.json
abraham2512, line 1: We might want to add this extends in the file, in case some future renovate fixes are applied from extensions
```
  "extends": [
    "config:recommended"
  ]
```
Found in konflux-ci build service Renovate [config](https://github.com/konflux-ci/build-service/blob/da411cbd6a3b2c3122e5644e8ede2adc16c2ae61/renovate.json#L3)

fontivan: nope, we do not
Mintmaker automatically extends their configuration behind the scenes so if we add another one upstream it can cause problems

renovate.json
```diff
@@ -35,11 +32,17 @@
...
         "schedule": [
             "at any time"
+        ],
+        "packageRules": [
+            {
+                "automerge": true,
+                "matchUpdateTypes": [
+                    "digest"
+                ]
+            }
```
fontivan, line 42: Missing the `addLabels` field to allow auto merging of tekton digest 
prs
https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/main/renovate.json#L59

renovate.json
shajmakh: Thanks for the comment. The intention was exactly so (see https://github.com/openshift-kni/cnf-features-deploy/blob/master/.gitmodules).  I wasn't aware RDS has is getting updated frequently and that requires immediate reflect for ZTP runs. Apparently I accounted only for NTO,Metallb and sriov submodules updates which updating them once a week seemed enough. I'll check for a way to split between the affected paths and adjust that. If there is no such option I'll drop this change.

renovate.json
shajmakh: I though automerge is enough based on: https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#automerge:~:text=%22packageRules%22%3A%20%5B%0A%20%20%20%20%7B%0A%20%20%20%20%20%20%22description%22%3A%20%22Automerge%20non%2Dmajor%20updates%22%2C%0A%20%20%20%20%20%20%22matchUpdateTypes%22%3A%20%5B%22minor%22%2C%20%22patch%22%5D%2C%0A%20%20%20%20%20%20%22automerge%22%3A%20true%0A%20%20%20%20%7D
it makes sense to add the labels though, thanks for highlighting this.

fontivan: The documentation is still pending some changes in https://issues.redhat.com/browse/KONFLUX-8240

PR #2752, opened 2025-07-28
ztp: sort bindingRules in PGT plugin for consistent ordering of generated placement rules

abraham2512:
This PR sorts the map of bindingRules and bindingExcludedRules to ensure the generated placement rules are ordered consistently.

### DIFF
--- a/ztp/policygenerator/policyGen/policyHelper.go
+++ b/ztp/policygenerator/policyGen/policyHelper.go
@@ -2,6 +2,7 @@ package policyGen
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


abraham2512:
cc: @irinamihai ptal when possible

ztp/policygenerator/policyGen/policyHelper.go
```diff
@@ -85,7 +86,15 @@ func CreatePlacementRule(name string, namespace string,
 	placementRule.Metadata.Namespace = namespace
 	expressions := make([]map[string]interface{}, 0)
 
-	for key, value := range bindingRules {
+	// Sort bindingRules to ensure consistent ordering of generated PlacementRules
```
irinamihai, line 89: Could we have a separate sorting function and use it in both places?

PR #2770, opened 2025-08-04
AA: cnf-tests: tekton: use PR number as quay builds tag

shajmakh:
Instead of using the commit revision as the quay build tag, use the PR number to easily track and identify the build in quay.

Assisted-by: Cursor v1.2.2
Assited-by-model: claude-4-sonnet


shajmakh:
workes as expected:
quay.io/redhat-user-workloads/telco-5g-tenant/cnf-tests-4-20:on-pr-2770
/unhold

PR #2801, opened 2025-08-11
ztp: update renovate config with matchFileNames

abraham2512:
This updates renovate config for telco-reference submodule "at any time" refresh

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -50,7 +50,7 @@
         "enabled": true,
         "packageRules": [
             {
-                "includePaths": [
+                "matchFileNames": [
                     "ztp/resource-generator/**"
                 ],
                 "schedule": [
@@ -59,7 +59,7 @@
                 "additionalBranchPrefix": "ztp/resource-generator-"
             },
             {
-                "includePaths": [
+                "matchFileNames": [
                     "cnf-tests/submodules/**"
                 ],
                 "additionalBranchPrefix": "cnf-tests/submodules-"


abraham2512:
@sabbir-47 From these docs here https://docs.renovatebot.com/configuration-options/#packagerules , we need atleast one `match...` key in each packageRules. Without it, it appears that it wont be matched to anything.

abraham2512:
cc: @shajmakh

abraham2512:
cc: @imiller0

renovate.json
```diff
@@ -50,7 +50,7 @@
         "enabled": true,
         "packageRules": [
             {
-                "includePaths": [
```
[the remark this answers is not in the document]
abraham2512: I think the packageRules are only required when we need to have different sync frequency for each submodule. It can be avoided if the default Renovate config

lack (approved):

renovate.json
```diff
@@ -59,7 +59,7 @@
                 "additionalBranchPrefix": "ztp/resource-generator-"
             },
             {
-                "includePaths": [
+                "matchFileNames": [
                     "cnf-tests/submodules/**"
```
fontivan, line 63: does matchFileNames not require the leading and trailing `/` for regex like I had to do here: https://github.com/openshift-kni/lifecycle-agent/pull/2046

abraham2512: According to the examples in this [doc](https://docs.renovatebot.com/configuration-options/#matchfilenames), it seems a match like "cnf-tests/submodules/**" would match all files that start with cnf-tests/submodules/

PR #2804, opened 2025-08-12
Ocpbugs 60163 siteconfig converter

sudomakeinstall2:


ztp/tools/siteconfig-converter/README.md
```diff
@@ -77,15 +79,73 @@ make build
...
+the request: Attempted to change AgentClusterInstall.Spec which is immutable
+after install started, except for ClusterMetadata,IgnitionEndpoint fields.
+Unsupported change: \n\tManifestsConfigMapRefs: ([] => [{Name:sno1}])"
+```
+
+**Before performing the migration**, verify whether the existing `AgentClusterInstall` resource on the hub cluster contains any of these deprecated fields. If it does, you have two resolution options:
+
+**Option 1: Create a Custom Cluster Template**
+
+Create a custom cluster template that uses the deprecated singular field forms instead of the current plural forms. The following diff shows the differences between the original template and the custom template:
```
irinamihai, line 100:
```suggestion
Create a ConfigMap that contains the custom cluster template that uses the deprecated singular field forms instead of the current plural forms. The following diff shows the differences between the original template and the custom template:
```?
It's initially unclear where the custom template should be put.

File: ztp/tools/siteconfig-converter/README.md [hunk 2]
```diff
@@ -138,22 +198,51 @@ Generate ClusterInstance by referencing the configmap
...
-siteconfig-generator -outPath </path/out> -extraManifestOnly </path/siteconfig.yaml>
+# Generate ConfigMap kustomization files automatically
+./siteconfig-converter -d ./output siteconfig.yaml
 ```
-Using `./hack/kustomize-generator.bash` you can genereate a `kustomization.yaml` that include all yaml files in a directory
-```bash
-./hack/kustomize-generator.bash <configmap-name> <configmap-namespace> </path/out/siteconfig-name/>
+
+This process:
+1. Generates extra manifests using the `siteconfig-generator` binary
```
irinamihai, line 211:
```suggestion
1. Extracts all the extra manifests using the `siteconfig-generator` binary
```?

File: ztp/tools/siteconfig-converter/README.md [hunk 3]
```diff
@@ -138,22 +198,51 @@ Generate ClusterInstance by referencing the configmap
...
+$ siteconfig-converter -d output cnfdf28.yaml
+...
+
+$ tree output
+output
+├── cnfdf28.yaml
+├── extra-manifests
+│   ├── cnfdf28_machineconfig_98-var-lib-containers-partitioned.yaml
+│   └── cnfdf28_machineconfig_99-set-core-user-password.yaml
+└── kustomization-configMapGenerator-snippet.yaml
```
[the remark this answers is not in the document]
sudomakeinstall2: in the parent directory.

ztp/tools/siteconfig-converter/convert.go
```diff
@@ -965,3 +1002,85 @@ func writeClusterInstanceToFile(clusterInstance *ClusterInstance, filename strin
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal kustomization to YAML: %w", err)
+	}
+
+	kustomizationContent := string(kustomizationData)
+
+	outputFile := filepath.Join(outputDir, KustomizationConfigMapGeneratorSnippetFile)
+	if err := os.WriteFile(outputFile, []byte(kustomizationContent), 0644); err != nil {
+		return fmt.Errorf("failed to write %s: %w", KustomizationConfigMapGeneratorSnippetFile, err)
+	}
```
imiller0, line 1077: What happens if the tool is used to convert multiple SiteConfig files into the same output directory. Will this overwrite them all?  Or is this enforcing one converted site per output directory? 
The kbase article we have shows all the ClusterInstance CRs in the same directory (eg [here](https://access.redhat.com/articles/7124143#transitioning-sno2))

ztp/tools/siteconfig-converter/main.go
```diff
@@ -234,20 +236,22 @@ type Partition struct {
...
+		clusterTemplate            = flag.String("t", "open-cluster-management/ai-cluster-templates-v1", "Comma-separated list of template references for Cluster (format: namespace/name,namespace/name,...)")
+		nodeTemplate               = flag.String("n", "open-cluster-management/ai-node-templates-v1", "Comma-separated list of template references for Nodes (format: namespace/name,namespace/name,...)")
+		extraManifestsRefs         = flag.String("m", "", "Comma-separated list of ConfigMap names for extra manifests references")
+		suppressedManifests        = flag.String("s", "", "Comma-separated list of manifest names to suppress at cluster level")
+		writeWarnings              = flag.Bool("w", false, "Write conversion warnings as comments to the head of converted YAML files")
+		copyComments               = flag.Bool("c", false, "Copy comments from SiteConfig to ClusterInstance YAML files")
+		extraManifestConfigMapName = flag.String("extraManifestConfigMapName", "extra-manifests-cm", "Name for the extra manifest ConfigMap")
+		manifestsDir               = flag.String("manifestsDir", "extra-manifests", "Directory containing extra manifest files")
```
imiller0, line 247: If we default to <clusterName> is this option needed?

PR #2815, opened 2025-08-13
cnf-tests: switch to newer vhost_user virtual driver

SchSeba:
we should stop using the tap virtual driver as it's not well maintain in the dpdk project and only use the vhost_user one.


bartwensley:
@SchSeba - just going through old PRs - is this change still valid?

PR #2890, opened 2025-08-20
[release-4.17] tekton: task digest updates

abraham2512:
Tekton task digest updates


abraham2512:
cc: @rauhersu @fontivan

PR #2948, opened 2025-09-10
ztp: OCPBUGS-61357: update policyTemplate namespaceSelector

abraham2512:
Updating namespaceSelector variable in Policy Template to camel case.

Previously namespaceselector was ignored because of namespace field being present. This change is now required because of [ACM-21257](https://issues.redhat.com/browse/ACM-21257), which has been done to raise field validation errors instead of ignoring.


abraham2512:
cc: @irinamihai @sabbir-47

PR #2993, opened 2025-09-25
ztp: remove the disableDrain for SNO

SchSeba:
On newer versions of the sriov-network-operator there is no need to request the disableDrain: true for SNO.

PR: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/850


cgoncalves:
/lgtm
/hold
Hold for other reviewers.

SchSeba:
Hi @lack, so I should open a PR to https://github.com/openshift-kni/telco-reference/tree/main/telco-ran ?
Also I am not sure if I want this to be optional. We are trying to hide this field from user. it was a hack in the operator to support SNO. But that is not the right case because if the operator needs to drain for any configuration change running workloads will still be running but the VF will be removed from them.

PR #3023, opened 2025-10-03
ztp: Add cgroup and runtime checks

abraham2512:
All workloads should be moving to cgroup v2. Add a required check for cgroup. If users remain on cgroup v1 the check serves as a reminder that workloads must be moved to v2.

Manual backport of https://github.com/openshift-kni/telco-reference/pull/269


imiller0 (approved):
/lgtm

PR #3029, opened 2025-10-07
activate auto-generating of rpm lockfile by default for ztp and cnf-tests

shajmakh:
Now that konflux pipelines support updating lockfiles on mono-repos,
ensure this is the default update method for this file on cnf-tests and ztp
while preserving the manual option.


shajmakh:
/cc @ffromani
 can you take a look please?

PR #3141, opened 2025-11-06
cnf-tests: rpms: use custom authentication certificate

shajmakh:
Following
https://konflux.pages.redhat.com/docs/users/mintmaker/rpm-lockfile.html#rpm-lockfile-with-rpms-that-require-subscription, use env variables to pass on the auth certificate of the telco tenant activation key.


cnf-tests/.konflux/Makefile
```diff
@@ -41,6 +41,9 @@ konflux-update-rpm-lock-runtime: sync-telco5g-konflux-submodule ## Update the rp
 	cp $(PROJECT_DIR)/Dockerfile $(PROJECT_DIR)/lock-runtime/Dockerfile
 	@echo "Copying rpms.in.yaml to lock-runtime directory..."
 	cp $(PROJECT_DIR)/rpms.in.yaml $(PROJECT_DIR)/lock-runtime/rpms.in.yaml
+	sed -i 's|sslclientkey: $$SSL_CLIENT_KEY|sslclientkey: /etc/pki/entitlement/placeholder-key.pem|g' $(PROJECT_DIR)/lock-runtime/rpms.in.yaml
```
fontivan: I don't think I follow this combined with the change to the yaml

Why change the placeholder in the yaml file just to change it back with sed? How does this get populated by the env?

PR #3219, opened 2025-11-28
infra: upversion go to 1.24.8 - CVE-2025-58183

abraham2512:
Upversion go with fix version 1.24.8 in response to CVE: https://www.cve.org/CVERecord?id=CVE-2025-58183

Note: Clair-scan is still flagging the CVE because the go1.24 builder image in konflux is on golang1.24.6 and we are awaiting the bump to golang1.24.8 (with the CVE fix) by the go-toolset team. Once this is available, this should be automatically resolved without any changes needed from our end. That target date currently for this bump is Jan 1, 2026.

### DIFF
--- a/cnf-tests/testsuites/pkg/junit-merger/junit-merger_test.go
+++ b/cnf-tests/testsuites/pkg/junit-merger/junit-merger_test.go
@@ -11,7 +11,7 @@ import (
 func TestJUnitMerger(t *testing.T) {
 	suites, err := loadJUnitFiles([]string{"testdata/junit1.xml", "testdata/junit2.xml"})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


abraham2512:
> Thanks, I approve this change. I see go.mod is updated, can you please update vendor for the bumped deps?

There are no vendor changes generated by this change. This is because,
1. This only includes a go version bump (1.23 -> 1.24). 
2. The changes in diff are dependency re-classifications, moved from direct -> indirect, but they are still the same versions so there is no change in the vendor directory.

abraham2512:
@lack e2e-aws-ci-tests is failing due to an unrelated error. Could you please override?

PR #3475, opened 2026-01-14
tekton: switch to 4.22 for master

shajmakh:
update 4.21->4.22 for cnf-tests and ztp tekton files.


shajmakh:
appearently there is an infra/outage issue in the pipelines

PR #3760, opened 2026-02-23
CNF-22019: infra: tools: use latest fedora

shajmakh:
The mintmaker (auto updater) bumps to the greatest tag available in quay. This commit fixes this to use the latest tag which will be last released tag.


shajmakh:
/unhold
It is by design that this change will be checked after it gets merged: 
https://docs.ci.openshift.org/architecture/ci-operator/#:~:text=In%20this%20case%2C%20the%20Dockerfile%20will%20always%20be%20obtained%20from%20current%20HEAD%20of%20the%20given%20branch%2C%20even%20if%20ci%2Doperator%20runs%20in%20the%20context%20of%20a%20PR%20that%20updates%20that%20Dockerfile.
see the comment on the Jira card for more details.

PR #3848, opened 2026-03-27
[release-4.21] CNF-22668: Replace centos references with ubi

fontivan:
We should not be using centos at this point, so instead use ubi


cnf-tests/GATEKEEPER.md
```diff
@@ -80,7 +80,7 @@ metadata:
 spec:
   containers:
   - name: main
-    image: centos
+    image: ubi8
```
yanirq: @fontivan  why ubi8 and not ubi9?

PR #3915, opened 2026-04-01
Add by2waysprojects to konflux OWNERS_ALIASES

by2waysprojects:
## Summary

Add `by2waysprojects` to `konflux-approvers` and `konflux-reviewers` in `OWNERS_ALIASES`.

Made with [Cursor](https://cursor.com)

PR #3979, opened 2026-04-15
infra: add `by2waysprojects` to konflux owners

shajmakh:


shajmakh:
already proposed in #3915

PR #4028, opened 2026-05-01
ztp: CNF-23439: Replace panic with error return in policy-object-template-diff

sebrandon1:
## Summary

- Replace `panic("No caller information")` with proper error return in `getObjectDefinitionFromConfigurationPolicy`
- Check and propagate error from `DefaultUnstructuredConverter.ToUnstructured` instead of silently discarding it
- Update call site in `extractCR` to log and skip on error rather than using bad data
- Hoist `path.Dir(filename)` out of the per-template loop to avoid redundant computation
- Replace if-chain with switch for CR kind matching

## Jira

https://issues.redhat.com/browse/CNF-23439

## Test plan

- [x] `go build ./ztp/tools/policy-object-template-diff/` compiles clean
- [ ] CI passes

### DIFF
--- a/ztp/tools/policy-object-template-diff/main.go
+++ b/ztp/tools/policy-object-template-diff/main.go
@@ -257,7 +257,11 @@ func extractCR(filePath string) []unstructured.Unstructured {
 
 	var allCRs []unstructured.Unstructured
 	for _, cP := range configurationPolicy {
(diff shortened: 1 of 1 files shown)


sebrandon1:
Jira status update: [CNF-23439](https://redhat.atlassian.net/browse/CNF-23439) is currently **To Do**.

PR #4070, opened 2026-05-19
Restrict MintMaker submodule updates to NTO

yanirq:
Disable Renovate updates for all submodules under cnf-tests/submodules/ and re-enable only cluster-node-tuning-operator. This stops unwanted PRs for sriov-network-operator and metallb-operator.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -77,7 +77,7 @@
                     "allow-automatic-merge"
                 ],
                 "matchFileNames": [
-                    "cnf-tests/submodules/**"
+                    "cnf-tests/submodules/cluster-node-tuning-operator"
                 ],
                 "additionalBranchPrefix": "cnf-tests/submodules-"
             },


SchSeba:
why we don't want to update also sriov for example?
if we have new tests or we fix flaky tests I would like our CI to catch them no?

yanirq:
> why we don't want to update also sriov for example? if we have new tests or we fix flaky tests I would like our CI to catch them no?

@SchSeba are you running the tests from u/s ?  this is a relatively  new behavior the automatic update of the submodule given by mintmaker.  as I understand from @sshnaidm  the submodule alignment is not happening here but in the test lanes 

on the other hand we do build cnf-tests image from the submodule of NTO

SchSeba:
oh ok so if the test run the git submodule command and use a branch for example 4.22 before it runs that test then I don't care here we can disable it :)

PR #4195, opened 2026-06-08
konflux: bump ose cli 4.22

yanirq:


cnf-tests/.konflux/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.redhat.io/openshift4/ose-cli-rhel9:v4.21 AS oc
+FROM registry.redhat.io/openshift4/ose-cli-rhel9:v4.22 AS oc
```
shajmakh, line 1: is this already released to production?

yanirq: we should probably need to wait post GA

yanirq:
/hold cancel

@by2waysprojects  PTAL

PR #4863, opened 2026-09-11
CNF-22668: Replace CentOS references with UBI

fontivan:
We should have stopped using CentOS a long time ago, but late is better than never


tools/s2i-dpdk/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/centos/centos:stream9
+FROM registry.access.redhat.com/ubi9/ubi:latest
```
shajmakh: who/what uses this Dockerfile? 
if it is prow CI then we need to take into consideration that it likely will not pick the updated Dockerfile in the PR run with the new ubi base, that's a known limitation in prow config. IOW need to make sure that who/whatever is using this Dockerfile will not face unexpected complications due to the base change.

fontivan: To my understanding this dockerfile is used by the e2e-telco5g-cnftests so I don't believe it has the same retest limitation as the openshift-ci/ dockerfile

zeeke: I think this image is not used anymore. At least I can't find any published image in the [catalog](https://catalog.redhat.com/en/search?q=dpdk&searchType=Containers), in the tekton configuration or in the [ART build history](https://art-build-history-art-build-history.apps.artc2023.pc3z.p1.openshiftapps.com/?assembly=stream&outcome=Success&engine=konflux&hermetic=both&buildtype=image&buildtype=bundle&buildtype=fbc&dateRange=2026-09-22+to+2026-09-24&nvr=dpdk)

This new one is what should be used, I guess.
https://github.com/openshift-kni/dpdk-base

I'm not 100% sure about my statement above, as it's hard to say if there is no automation around pointing to this.

If you think it's too risky to drop the folder entirely, the changes proposed by this PR are good enough to be merged.

